PR #17, opened 2021-05-27
adding a comment explaining the type def for context

jhjaggars:


omaciel:
I'd approve it and merge it but it doesn't look like I have the right access rights to do it :/

PR #18, opened 2021-05-27
blocking out some ideas for repo hosting

jhjaggars:
This PR introduces a new endpoint `/repos` that proxies repo content.  The idea is to separate commit data from actual repo data, and allow space for the repo to be constructed and served in distinct steps.

### DIFF
--- a/pkg/commits/commits.go
+++ b/pkg/commits/commits.go
@@ -1,7 +1,6 @@
 package commits
 
 import (
-	"archive/tar"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -16,6 +15,7 @@ import (
 	"github.com/go-chi/chi"
 
 	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/edge-api/pkg/common"
 	"github.com/redhatinsights/edge-api/pkg/db"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 	"gorm.io/gorm"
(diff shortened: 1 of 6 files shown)


pkg/commits/commits.go
```diff
@@ -57,6 +57,11 @@ func MakeRouter(sub chi.Router) {
 	})
 }
 
+// This provides type safety in the context object for our "commit" key.  We
```
omaciel, line 60: Looks like you may have to rebase? This was part of your previous PR

pkg/repo/repo.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+
+	if cr.TarURL == "" {
+		http.Error(w, "tarUrl must be set", http.StatusBadRequest)
+		return
+	}
+
+	if cr.Name == "" {
+		cr.Name = "default" // should be randomized?
```
omaciel, line 43: As part of the common module, someone could implement something like https://golangdocs.com/generate-random-string-in-golang:

```go
import (
    "fmt"
    "math/rand"
)
 
func RandomString(n int) string {
    var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
 
    s := make([]rune, n)
    for i := range s {
        s[i] = letters[rand.Intn(len(letters))]
    }
    return string(s)
}
 
func main() {
    fmt.Println(RandomString(10))        // BpLnfgDsc2
}
```

omaciel (changes requested):
I think you need to rebase but otherwise, let 'er rip :)

maxamillion:
I'm merging because I'm going to base the next thing I do on the work in this PR and Jesse is going to be afk for a bit

PR #26, opened 2021-06-02
Run migrations via an initContainer.

chambridge:
* update Clowdapp definition


chambridge:
Here you can see the init container was created and run:
image

Here you can see the details of the initConatiner:
image
and the inherited environment:
image

maxamillion (approved):

PR #28, opened 2021-06-03
Add commit update staging

maxamillion:


cmd/migrate/migrate.go
```diff
@@ -13,18 +13,22 @@ func main() {
...
+		"LogLevel":    cfg.LogLevel,
+		"Debug":       cfg.Debug,
+		"BucketName":  cfg.BucketName,
+	}).Info("Configuration Values:")
 	db.InitDB()
 	err := db.DB.AutoMigrate(&commits.Commit{})
 	if err != nil {
 		panic(err)
 	}
+	err = db.DB.AutoMigrate(&commits.UpdateRecord{})
```
jhjaggars, line 29: This can be combined with the previous AutoMigrate call fyi

pkg/commits/updates.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/edge-api/pkg/common"
+	"github.com/redhatinsights/edge-api/pkg/db"
+	"gorm.io/gorm"
+
+	"github.com/cavaliercoder/grab"
+	log "github.com/sirupsen/logrus"
+)
+
+var UPDATE_TMPPATH = "/tmp/updates/"
```
jhjaggars, line 29: Stick this in config somehow, you'll thank me later :)

pkg/commits/uploader.go
```diff
@@ -0,0 +1,117 @@
...
+		)
+		if err != nil {
+			log.Warnf("error: %w", err)
+			return err
+		}
+		return nil
+	})
+
+	region := *u.Client.Config.Region
+	s3URL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s/%s", u.Bucket, region, account, strings.TrimPrefix(src, UPDATE_TMPPATH))
```
jhjaggars, line 91: I assume this will be the proxy url later, is that the intention?

jhjaggars (changes requested):

pkg/commits/uploader.go
maxamillion: Yes, definitely .... I wanted to talk to you about how we were going to glue those bits together but for the time being I wanted to have an entry and some information about what was uploaded and where it went.

config/config.go
```diff
@@ -8,17 +8,18 @@ import (
...
+	Hostname       string
+	Auth           bool
+	WebPort        int
+	MetricsPort    int
+	Logging        *loggingConfig
+	LogLevel       string
+	Debug          bool
+	Database       *dbConfig
+	BucketName     string
+	AccessKey      string
```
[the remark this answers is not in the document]
chambridge: or maybe `objectstorage` since it should be able to be s3-compliant.

chambridge:
Certainly not required and not new to this PR, but would suggest the S3 interaction using the `minio` [golang client](https://docs.min.io/docs/golang-client-quickstart-guide.html) instead of the AWS SDK. If we ever want IVT test to run as part of the pr_check we would need this.

pkg/commits/updates.go
```diff
@@ -0,0 +1,403 @@
...
+
+func updateFromReadCloser(rc io.ReadCloser) (*UpdateRecord, error) {
+	defer rc.Close()
+	var update UpdateRecord
+	err := json.NewDecoder(rc).Decode(&update)
+
+	log.Debugf("updateFromReadCloser::update: %#v", update)
+	log.Debugf("updateFromReadCloser::update.UpdateCommitID: %d", update.UpdateCommitID)
+
+	if !(update.UpdateCommitID > 0) {
```
jhjaggars, line 64: `if update.UpdateCommitID <= 0`?

test_vpn.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
maxamillion: it was not

config/config.go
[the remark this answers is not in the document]
maxamillion: I didn't add those, they're included in my changelog because gofmt automatically shifted the spacing when I added `UpdateTempPath`.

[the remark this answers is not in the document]
jhjaggars: Let's clean this up in a later PR.

maxamillion:
> Certainly not required and not new to this PR, but would suggest the S3 interaction using the `minio` [golang client](https://docs.min.io/docs/golang-client-quickstart-guide.html) instead of the AWS SDK. If we ever want IVT test to run as part of the pr_check we would need this.

That sounds perfectly fine to me, I'd like to do that in a follow-up PR though.

PR #31, opened 2021-06-04
Add playbook and readme to setup the project using ansible

acosferreira:


chambridge:
Update seems fine. Wondering whether it makes sense to place the playbook in a folder instead of at the top level (like a `dev` folder; the default_config.* files could potentially be relocated there)? Just thinking of longer term project structure.

biancarosa:
I'm not particularly a fan on ansible playbooks to run apps inside of the apps repo because it looks odd - I think I mentioned that to @acosferreira.

But maybe we could have a collection of playbooks that helps us to set up our local machines and maybe they could live in a repo like... `fleet-management-development-setup` or something? We could add *several* playbooks to help us there, not just this one!

acosferreira:
@chambridge I've moved the file to a dev-setup folder and we can open a discussion on the team about @biancarosa suggestion.

maxamillion (changes requested):
For the playbook I'd prefer we use builtin Ansible modules as much as possible, so something like this:

```yaml
- name: "edge-api-locally"
  hosts: localhost
  vars:
    src_dir: "{{ansible_user_dir}}"
  tasks:
    - name: "ensure {{src_dir}} exists"
      file:
        path: "{{src_dir}}"
        state: directory
    - name: clone edge-api
      git:
        repo: "git@github.com:RedHatInsights/edge-api.git"
        dest: "{{src_dir}}/edge-api"
    - name: run the migrations
      shell: "go run cmd/migrate/migrate.go"
      args:
        chdir: "{{src_dir}}/edge-api"
```

Also for the final task, I'm not sure we want to run the service because that's going to cause Ansible to block waiting for it to complete which will eventually timeout and then the edge-api local service will be killed. We could potentially background it but something tells me that's not the desired functionality because then users would have to hunt around to find the pid to kill in order to stop the service.

Thoughts?

PR #33, opened 2021-06-07
setup api documentation with go-swagger

biancarosa:
This PR sets up API documentation capabilities with go-swagger. It has a README to help whoever wants to generate these docs. 

The build was failing and it made me enter a rabbit hole and I found a issue relative to a bug in Go 1.16: https://github.com/golang/go/issues/44129. We were using Go 1.14 for development (according to docs), Go ^1.14 on GH Actions (it was pulling Go 1.16) and Go 1.15 in stage/production because of the version that [go-toolset](https://catalog.redhat.com/software/containers/rhel8/go-toolset/5b9c810add19c70b45cbd666?container-tabs=packages) pulls. It feels like a better practice to pin all of our version and leverage them to the version that is gonna be used in production so we can prevent errors accordingly and not waste time and effort in errors that only happens locally and/or in our CI platform.


Makefile
```diff
@@ -82,4 +82,7 @@ restart-app:
 	sleep 5
 	$(MAKE) scale-up NAMESPACE=$(NAMESPACE)
 
+generate-docs:
+	swagger generate spec -o ./swagger.yml
+
 .PHONY: help build
```
omaciel, line 88: Do you need to explicitly add `generate-docs` here?

biancarosa: In order to the `make generate-docs` command to work, yes, but the person can always run these manually without using the Makefile.

maxamillion:
@biancarosa I just ran this and it didn't produce the expected openapi 2.0 spec, it's missing all of the `paths` ... is there something else we need to do in order for those things to be populated?

biancarosa:
oh, yeah, we need to add a couple more annotations on each endpoint, but I'd like to setup this first before starting to write documentation on all endpoints. there are 2 user stories to write docs around each group of endpoints.

maxamillion (approved):

PR #35, opened 2021-06-09
Added new make target for validating swagger file.

omaciel:
Added a new `validate-swagger` make target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,4 +85,7 @@ restart-app:
 generate-docs:
 	swagger generate spec -o ./swagger.yml
 
+validate-swagger:
+	swagger validate ./swagger.yml
+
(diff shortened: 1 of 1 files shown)


omaciel:
```zsh
» make validate-swagger 
swagger validate ./swagger.yml
2021/06/09 15:54:20 
The swagger spec at "./swagger.yml" is valid against swagger specification 2.0
2021/06/09 15:54:20 
The swagger spec at "./swagger.yml" showed up some valid but possibly unwanted constructs.
2021/06/09 15:54:20 See warnings below:
2021/06/09 15:54:20 - WARNING: spec has no valid path defined
```

biancarosa (approved):
LGTM!

PR #37, opened 2021-06-09
WIP - Add tests to commit class

acosferreira:


pkg/commits/commits_test.go
```diff
@@ -16,3 +25,187 @@ func TestPatch(t *testing.T) {
...
 }
+
+func TestGetAllEmpty(t *testing.T) {
+	config.Init()
+	config.Get().Debug = true
+	db.InitDB()
+	err := db.DB.AutoMigrate(Commit{})
+	if err != nil {
+		panic(err)
+	}
```
chambridge, line 36: I would pull this out into a common function that the test could use instead of it being copied an pasted into each test case.

pkg/commits/commits_test.go
```diff
@@ -16,3 +25,187 @@ func TestPatch(t *testing.T) {
...
+	t.Run("returns empty commits", func(t *testing.T) {
+		request, _ := http.NewRequest(http.MethodGet, "/", nil)
+		response := httptest.NewRecorder()
+
+		GetAll(response, request)
+
+		got := response.Body.String()
+		want := "[]\n"
```
chambridge, line 44: Instead of string comparison Its probably worth processing this as [JSON](https://stackoverflow.com/a/31129967).

I would also make the response --> JSON processing a reusable function for each of the tests.

pkg/commits/commits_test.go
```diff
@@ -16,3 +25,187 @@ func TestPatch(t *testing.T) {
...
+		response := httptest.NewRecorder()
+
+		GetAll(response, request)
+		got := response.Body.String()
+		if !strings.Contains(got, "0000000") {
+			db.DB.Delete(&cmt)
+			t.Errorf("got %q", got)
+		}
+	})
+	db.DB.Delete(&cmt)
```
chambridge, line 79: The initialize and delete db steps are a good candidate for a [setup/teardown flow](http://www.inanzzz.com/index.php/post/2t08/using-setup-and-teardown-in-golang-unit-tests).

chambridge (changes requested):
Some potential opportunities for code reuse and test structuring.
As well as suggestions for output testing to be certain we are returning valid JSON.

pkg/commits/commits_test.go
biancarosa: We could also start taking a look at test suites and make use of SetUp / TearDown methods

biancarosa (approved):
This is the way.

chambridge:
Looks like just a minor **go vet** issue to resolve then good 👍 

```
Run make vet
go vet $(go list ./... | grep -v /vendor/)
# github.com/redhatinsights/edge-api/pkg/commits
pkg/commits/uploader.go:85:4: Warnf call has error-wrapping directive %w
make: *** [Makefile:50: vet] Error 2
Error: Process completed with exit code 2.
```

chambridge (approved):
LGTM

PR #38, opened 2021-06-10
Feature: Create image by calling compose endpoint on Image Builder

biancarosa:
bad bad request
`curl -v -H "Content-Type: application/json"  http://localhost:3000/api/edge/v1/images -d '{wrong json}'`

minimum valid request
`curl -v -H "Content-Type: application/json" http://localhost:3000/api/edge/v1/images -d '{"Distribution": "rhel-8", "OutputType": "tar", "Commit": {"Arch": "x86_64"}}'`

with packages:
`curl -v -H "Content-Type: application/json" http://localhost:3000/api/edge/v1/images -d '{"Distribution": "rhel-8", "OutputType": "tar", "Commit": {"Arch": "x86_64"}, "Packages" : [ { "name" : "vim"  }]}'`

Still have to:
- Write better tests

Low-priority, probably not worthy to do right now:
- Perfect API Docs - We're gonna need to spend some time figuring out how to migrate to a tool that uses open api v3


pkg/models/images.go
```diff
@@ -0,0 +1,31 @@
...
+import "errors"
+
+// An Image is what generates a OSTree Commit.
+//
+// swagger:model image
+type Image struct {
+	Distribution string // rhel-8
+	Architecture string // x86_64
+	OSTreeRef    string // "rhel/8/x86_64/edge"
+	OSTreeURL    string
```
maxamillion, line 12: Something we could do here is have `Architecture`, `OSTreeRef`, and `OSTreeURL`  be only stored in the [`Commit`](https://github.com/RedHatInsights/edge-api/blob/8bd239ed5ba1fbd0e82a6eabbe97cf7683785b18/pkg/commits/commits.go#L22) and the `Image` type then have a field that correlated to a `Commit` either as a `Commit.ID`, a `*Commit` or a foreign key in the table.

Thoughts?

pkg/images/images.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/redhatinsights/edge-api/pkg/errors"
+	"github.com/redhatinsights/edge-api/pkg/models"
+)
+
```
maxamillion, line 11: @jhjaggars does need anything from the middleware here to deal with accounts?

maxamillion (changes requested):

pkg/images/images.go
jhjaggars: Yes, tenancy has to be built into the model itself (i.e. add an account/org field), and requests need to be filtered by the Identity object from middleware as in the other handlers.

pkg/models/images.go
biancarosa: I like that! Gonna update the code to reference a commit here

pkg/images/images.go
biancarosa: I did that, don't know exactly if I did it on the best possible way since Image has a Commit and both have an Account

biancarosa:
@maxamillion apart from tests, this is the best i could do right now 😓 

take a look, especially at the models, because I'm still not sure what goes into the image and what goes into the commit (maybe the packages have to move to the commit)

pkg/models/images.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// swagger:model image
+type Image struct {
+	gorm.Model
+	Name         string
+	Account      string
+	Distribution string // rhel-8
+	Description  string
+	OutputType   string
+	Packages     []Package `gorm:"many2many:image_packages;"`
```
maxamillion, line 19: The `Package`s manifest belongs to the `Commit` and the `Commit` to the `Image`.

Effectively "An `Image` contains a `Commit` and a `Commit` contains many `Packages`. Currently in the `Commit` struct there's a field `NEVRAManifest` which was a meant for that to specify that it was a NEVRA representation of the package manifest for the `Commit` (but I have zero preference on what we actually call the thing, "`Packages`" is perfectly fine).

biancarosa: Will move that! I think we should be able to get the _NEVRAManifest_ field as a response from Image Builder at some point, so we could keep both fields for now and see if we really need that info duplicated or not. We were a bit a concerned about the size of the package list.

pkg/models/commits.go
```diff
@@ -0,0 +1,35 @@
...
+	ImageBuildTarURL     string
+	OSTreeCommit         string
+	OSTreeParentCommit   string
+	OSTreeRef            string
+	BuildDate            string
+	BuildNumber          uint
+	BlueprintToml        string
+	NEVRAManifest        string
+	Arch                 string
+	Packages             []Package `gorm:"many2many:image_packages;"`
```
maxamillion, line 21: If we're going to use `Packages` here, we can remove the `NEVRAManifest` field since it's not longer needed (it's a duplicate with the introduction of `Packages`).

maxamillion:
If you can make the GitHub Actions checks happy, I'm good with the PR. LGTM :+1:

biancarosa:
@maxamillion i think we're good now! i still wanna make those tests look a little bit better but this is a blocker for the other tasks so we can do it in another PR

biancarosa:
Merging this since Adam's on PTO and already commented that he was good with the changes :)

PR #39, opened 2021-06-15
Feat: Relay status from img builder

biancarosa:
Depends on #38


biancarosa:
This is getting a weird diff from main, I'm opening a new PR

PR #40, opened 2021-06-15
Fix S3Proxy

maxamillion:
* Re-Add NewS3Proxy
* Fix prefix parsing to include the update number
* Add debug output

### DIFF
--- a/main.go
+++ b/main.go
@@ -84,6 +84,9 @@ func main() {
 	server = &repo.FileServer{
 		BasePath: "/tmp",
 	}
+	if cfg.BucketName != "" {
+		server = repo.NewS3Proxy()
+	}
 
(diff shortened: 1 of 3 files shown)


biancarosa (approved):
LGTM. Only check not passing is relative to something that's already on the master branch, I  fixed on my PR already, so I don't think it's worthy to hold this on

maxamillion:
Yeah, don't merge this yet. I just realized this breaks something. I'll deal with that later.

maxamillion:
Closing for now, I'll re-open when I fix it but I'm going on PTO so it'll be a little bit.

maxamillion:
Never mind, the failure I ran into turns out to be disjoint from this.

PR #41, opened 2021-06-16
don't 500 when we get a 404

maxamillion:

### DIFF
--- a/pkg/repo/server.go
+++ b/pkg/repo/server.go
@@ -88,7 +88,7 @@ func (p *S3Proxy) ServeRepo(w http.ResponseWriter, r *http.Request) {
 		Key:    aws.String(realPath),
 	})
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		http.Error(w, err.Error(), http.StatusNotFound)
 		return
 	}
 


maxamillion:
@jhjaggars if we 500 then `rpm-ostree` bombs out, apparently `rpm-ostree` tries to check for some optional index files and if they 404 then it's fine and the transaction keeps going ... if we 500 then it bombs. I'm not sure how to properly handle this case because I don't know how to extract the status code from the S3 error returned, but this worked for swinging the hammer.

jhjaggars:
the docs have this example that you might consider starting from:
```
if err != nil {
    if aerr, ok := err.(awserr.Error); ok {
        switch aerr.Code() {
        case s3.ErrCodeNoSuchKey:
            fmt.Println(s3.ErrCodeNoSuchKey, aerr.Error())
        case s3.ErrCodeInvalidObjectState:
            fmt.Println(s3.ErrCodeInvalidObjectState, aerr.Error())
        default:
            fmt.Println(aerr.Error())
        }
    } else {
        // Print the error, cast err to awserr.Error to get the Code and
        // Message from an error.
        fmt.Println(err.Error())
    }
    return
}
```

biancarosa (commented):
@maxamillion d'you think we can fix those comments that the linter is complaining about?

jhjaggars:
Looks like this needs a rebase to pull in your fixes @biancarosa

PR #42, opened 2021-06-16
Feat: Get status from img builder

biancarosa:
`curl -v -X GET -H "Content-Type: application/json" http://localhost:3000/api/edge/v1/images/<image-id>/status`

response:

`{ "Status" : "SUCCESS"}`
`{ "Status" : "BUILDING"}`
`{ "Status" : "ERROR"}`

This PR adds an endpoint that returns the status of an image.
If still building, it makes a request to image builder's API, gets the updated image status and saves the URL returned from them in our database.

### DIFF
--- a/pkg/imagebuilder/client.go
+++ b/pkg/imagebuilder/client.go
@@ -72,15 +72,15 @@ const (
 
 type UploadStatus struct {
 	Options S3UploadStatus `json:"options"`
-	Status  string            `json:"status"`
-	Type    UploadTypes       `json:"type"`
+	Status  string         `json:"status"`
+	Type    UploadTypes    `json:"type"`
 }
 type ComposeResult struct {
 	Id string `json:"id"`
 }
 
 type S3UploadStatus struct {
-	Url string `json:"url"`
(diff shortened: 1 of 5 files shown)


pkg/imagebuilder/client.go
```diff
@@ -106,6 +135,40 @@ func (c *ImageBuilderClient) Compose(image *models.Image) (*models.Image, error)
...
+	url := fmt.Sprintf("%s/composes/%s", cfg.ImageBuilderConfig.URL, image.ComposeJobID)
+	req, _ := http.NewRequest("GET", url, nil)
+
+	client := &http.Client{}
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	respBody, err := ioutil.ReadAll(res.Body)
```
chambridge, line 155: You may want to process the response status and our content-type before processing the body.

chambridge (changes requested):
Minor comment. Overall good progress.

chambridge (approved):

PR #52, opened 2021-06-23
more gracefully handle s3 object 404 vs 500 error

maxamillion:

### DIFF
--- a/pkg/repo/server.go
+++ b/pkg/repo/server.go
@@ -1,6 +1,8 @@
 package repo
 
 import (
+	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pkg/repo/server.go
```diff
@@ -88,7 +89,22 @@ func (p *S3Proxy) ServeRepo(w http.ResponseWriter, r *http.Request) {
 		Key:    aws.String(realPath),
 	})
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case s3.ErrCodeNoSuchKey:
+				log.Debugf("S3Proxy::ServeRepo::s3.ErrCodeNoSuchKey: %#v", realPath)
+				http.Error(w, err.Error(), http.StatusNotFound)
```
biancarosa, line 96: At some point Chris pointed out that there's a pattern for errors, so I wrote a couple of helpers in the error package to help us with that. D'you think we can start using them here? `pkg/images/images.go` uses it.

biancarosa: (okay, there's not really helpers, but just structs that helps us with the standard)

acosferreira (approved):

biancarosa (approved):

PR #54, opened 2021-06-24
Generate openapi docs v3 & add docs to all image endpoints

biancarosa:


biancarosa:
@maxamillion couldn't go much further, I could generate the Schema for the models but had to write the path by hand. Thoughts on it?

c/c @acosferreira

maxamillion:
@biancarosa I'm good with it either way, my question to you would be: do you feel that this is better or worse than maintaining the openapi spec by hand long term? :)

biancarosa:
@maxamillion better! we need to maintain the path.yaml file by hand, but it's just the routes. the models itself it's definitely better to generate like that,

biancarosa:
Okay, so I added docs to all endpoints being used to the front-end right now :)

Figured we could split the work on the other ones, or only add as they're necessary, but I'm happy to work on all of them in this PR, too, if y'all think it's a good idea.

maxamillion (approved):
LGTM

PR #57, opened 2021-06-28
Fix/fix get repo when name contains part of prefix

biancarosa:
Trying to do the whole flow, came across this bug where a curl to `http://localhost:3000/api/edge/v1/repos/1/repo/config` 

gave me:
`{
  "Code": "NOT_FOUND",
  "Status": 404,
  "Title": "S3 Object 0000000/1/repos/1/repo/config not found."
}`

after that fix, I got:
`[core]
repo_version=1
mode=archive-z2`


omaciel:
Glad to see tests being included :)

pkg/repo/common.go
```diff
@@ -15,9 +15,14 @@ func getNameAndPrefix(r *http.Request) (string, string, error) {
...
-	_r := strings.Index(r.URL.Path, name)
+	pathPrefix := getPathPrefix(r.URL.Path, name)
+	return name, pathPrefix, nil
+}
+
+func getPathPrefix(path string, name string) string {
+	_r := strings.Index(path, "/"+name+"/")
 	log.Debugf("getNameAndPrefix::_r: %#v", _r)
-	pathPrefix := string(r.URL.Path[:_r])
+	pathPrefix := string(path[:_r+1])
```
acosferreira, line 25: is it possible to move this "magic number" to const that makes sense of this value?

acosferreira (approved):

PR #59, opened 2021-06-29
update manager data model

maxamillion:
If a device is not actively reachable or "online" via Cloud Connector, then
Playbook Dispatcher will 404 error when trying to schedule work for the device.
This behavior is the same for "offline" devices as well as invalid
device UUIDs, this means that we are going to have to manage the update
transaction. This change provides the basic data model for the update
transaction.

### DIFF
--- a/pkg/models/updates.go
+++ b/pkg/models/updates.go
@@ -1,0 +1,67 @@
+package models
+
+import (
+	"errors"
+
+	"gorm.io/gorm"
+)
+
+/*
+Device
+
+	A record of Edge Devices referenced by their UUID as per the
+	cloud.redhat.com Inventory.
+
+	ConnectionState refers to the devices Cloud Connector state, 0 is unavailable
+	and 1 is reachable.
+*/
+type Device struct {
+	gorm.Model
+	UUID            string
+	ConnectionState int `gorm:"default:1"`
+}
+
+/*
+UpdateRecord
+
+	Reporesents the combination of an OSTree commit and a set of Inventory
+	hosts that need to have the commit deployed to them
+
+	This will ultimately kick off a transaction where the old version(s) of
+	OSTree commit that are currently deployed onto those devices are combined
+	with the new commit into a new OSTree repo, static deltas are computed, and
+	then the result is stored in a way that can be served(proxied) by a
+	Server (pkg/repo/server.go).
+*/
+type UpdateRecord struct {
+	gorm.Model
+	UpdateCommitID uint
+	Account        string
+	OldCommitIDs   string
+	InventoryHosts []Device `gorm:"many2many:updaterecord_devices;"`
+	Status         string
+	UpdateRepoURL  string
+}
+
+const (
+	// Errors
+	UpdatCommitIDCantBeNilMessage    = "update commit id can't be empty"
+	AccountCantBeEmptyMessage        = "account can't be empty"
+	InventoryHostsCantBeEmptyMessage = "inventory hosts can not be empty"
+
+	// Status
+	UpdateStatusCreated  = "CREATED"
+	UpdateStatusBuilding = "BUILDING"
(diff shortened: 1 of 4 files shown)


acosferreira (changes requested):
Should fix the call on image_test
https://github.com/maxamillion/edge-api/blob/update_manager/pkg/images/images_test.go#L27

pkg/models/updates.go
```diff
@@ -0,0 +1,67 @@
...
+type Device struct {
+	gorm.Model
+	UUID            string
+	ConnectionState int `gorm:"default:1"`
+}
+
+/*
+UpdateRecord
+
+	Reporesents the combination of an OSTree commit and a set of Inventory
```
biancarosa, line 27:
```suggestion
	Represents the combination of an OSTree commit and a set of Inventory
```

pkg/models/updates.go
```diff
@@ -0,0 +1,67 @@
...
+	Account        string
+	OldCommitIDs   string
+	InventoryHosts []Device `gorm:"many2many:updaterecord_devices;"`
+	Status         string
+	UpdateRepoURL  string
+}
+
+const (
+	// Errors
+	UpdatCommitIDCantBeNilMessage    = "update commit id can't be empty"
```
biancarosa, line 48:
```suggestion
	UpdateCommitIDCantBeNilMessage    = "update commit id can't be empty"
```

maxamillion:
> Should fix the call on image_test
> https://github.com/maxamillion/edge-api/blob/update_manager/pkg/images/images_test.go#L27

okay ... this is super weird, my local branch has that change 

```
$ git diff main cmd/migrate/migrate.go
diff --git a/cmd/migrate/migrate.go b/cmd/migrate/migrate.go
index d2ebb8d..aabfb79 100644
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -3,7 +3,6 @@ package main
 import (
        "github.com/redhatinsights/edge-api/config"
        l "github.com/redhatinsights/edge-api/logger"
-       "github.com/redhatinsights/edge-api/pkg/commits"
        "github.com/redhatinsights/edge-api/pkg/db"
        "github.com/redhatinsights/edge-api/pkg/models"
        log "github.com/sirupsen/logrus"
@@ -23,7 +22,7 @@ func main() {
                "BucketName":  cfg.BucketName,
        }).Info("Configuration Values:")
        db.InitDB()
-       err := db.DB.AutoMigrate(&models.Commit{}, &commits.UpdateRecord{}, &models.Package{}, &models.Image{})
+       err := db.DB.AutoMigrate(&models.Commit{}, &models.UpdateRecord{}, &models.Package{}, &models.Image{})
        if err != nil {
                panic(err)
        }

$ git push origin update_manager
Everything up-to-date
```

maxamillion:
@acosferreira oh ha! sorry ... I totally blanked, fixing.

pkg/models/updates.go
```diff
@@ -0,0 +1,67 @@
...
+	InventoryHosts []Device `gorm:"many2many:updaterecord_devices;"`
+	Status         string
+	UpdateRepoURL  string
+}
+
+const (
+	// Errors
+	UpdateCommitIDCantBeNilMessage   = "update commit id can't be empty"
+	AccountCantBeEmptyMessage        = "account can't be empty"
+	InventoryHostsCantBeEmptyMessage = "inventory hosts can not be empty"
```
acosferreira, line 50: Should we keep a standard in the messages? (can not/can't)
other suggestion , Can we let as generic message and pass the context as parameter?

acosferreira (approved):

pkg/models/updates.go
maxamillion: Yes! I think that's a great idea.

PR #60, opened 2021-06-30
Drop auth for ostree repo endpoint

biancarosa:
Depends on #57


maxamillion (approved):

maxamillion:
For posterity, I wanted to note that this is required at least in the near term because of a limitation of Hosted Image Builder.

PR #61, opened 2021-06-30
Fix extraction of tar files

biancarosa:
This pull request fixes the extraction of tar files. I was getting too many open files and I think we ran across the problem described here on #2 
https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -80,13 +80,13 @@ func Untar(rc io.ReadCloser, dst string) error {
 			}
 			continue
 		}
-
 		file, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, info.Mode())
 		if err != nil {
 			return err
 		}
 		_, err = io.Copy(file, tarReader)
 		if err != nil {
+			file.Close()
 			return err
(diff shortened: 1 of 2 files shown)


pkg/common/common.go
```diff
@@ -85,11 +85,11 @@ func Untar(rc io.ReadCloser, dst string) error {
 		if err != nil {
 			return err
 		}
-		defer file.Close()
 		_, err = io.Copy(file, tarReader)
 		if err != nil {
```
boaz0, line 89: Nice catch. Just I was wondering :thinking: shouldn't we have `file.Close` under this `if` statement?
I will explain myself. Let's say I opened the file and in the middle of copying the data from `tarReader` to `file`. we don't need the file anymore and we can close it (before returning `err`). Or I am missing something here.

biancarosa: Great catch @boaz0 !

boaz0:
CI failure is not related to @biancarosa changes.

boaz0 (approved):
LGTM :+1: 
Amazing job :100: @biancarosa

PR #62, opened 2021-07-02
Endpoint upgrade

acosferreira:
Created endpoint to get them to accept either a single device UUID or an Inventory tag and query Inventory for device metadata


config/config.go
```diff
@@ -55,6 +60,7 @@ func Init() {
 	options.SetDefault("Debug", false)
 	options.SetDefault("EdgeTarballsBucket", "rh-edge-tarballs")
 	options.SetDefault("ImageBuilderUrl", "http://image-builder:8080")
+	options.SetDefault("InventoryUrl", "https://cloud.stage.redhat.com/")
```
biancarosa, line 63: What's the underlying API we're accessing? Do we have to go through cloud.stage or we can access within the cluster? @chambridge

pkg/devices/devices.go
```diff
@@ -0,0 +1,120 @@
...
+		return "Error to parse inventory url"
+	}
+	inventoryUrl.Path = path.Join(inventoryUrl.Path, inventoryAPI)
+	fullUrl := inventoryUrl.String()
+	return fullUrl
+}
+
+func ReturnDevices(w http.ResponseWriter, r *http.Request) (Inventory, error) {
+	transport := setProxy()
+	client := &http.Client{Transport: transport}
```
biancarosa, line 59: I'd avoid the proxy + basic auth and just go with the URL. The combo of 

IMAGEBUILDERURL=https://<user>:<pass>@cloud.stage.redhat.com/api/image-builder
HTTP_PROXY=http://<user>:<pass>@squid.corp.redhat.com:3128/
HTTPS_PROXY=http://<user>:<pass>t@squid.corp.redhat.com:3128/

worked for me pretty well. Can we try that, since the proxy it's not really needed for production?

biancarosa (commented):
Left a couple of comments + I think we should pay attention to this noisy linter, so there are a couple of things to fix on the devices.go file :)

config/config.go
chambridge: Should be able to access the service within the cluster.

omaciel:
I think we should avoid using internal URLs here :)

acosferreira:
@omaciel I'm working to move the URLs to env and to work in the cluster as suggested by @biancarosa

PR #68, opened 2021-07-06
Add test to untar method in common

adeliaferreira:
### What is inside?
Add a new unit test to untar method in common package

### How is it done?

- Create a new test method
- create a mock tar file based in a files map structure
- call untar method from common package
- iterate over files map and check if all files are available after untar
- delete file to avoid generate files unecessary to project

### DIFF
--- a/pkg/common/common_test.go
+++ b/pkg/common/common_test.go
@@ -1,9 +1,12 @@
 package common
 
 import (
+	"archive/tar"
 	"context"
+	"log"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"testing"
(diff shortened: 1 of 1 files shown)


acosferreira (approved):

biancarosa (approved):

omaciel:
One thing to look into is using the `ioutil` package to generate temporary files, as shown below:

```
tmpFile, err := ioutil.TempFile(tempPath, "tempfile_*.txt")
	if err != nil {
		panic(err)
	}
defer os.Remove(tmpFile.Name())
```

PR #70, opened 2021-07-06
Update openapi.json mount point

chambridge:
* update with api prefix for doc serving

image


biancarosa (approved):
LGTM! Tks @chambridge, it was a small change I couldn't do it before :)

PR #73, opened 2021-07-07
Endpoint upgrade

acosferreira:
Created endpoint to get the devices by tag or id
Returns the device metadata
Returns the current rpm-ostree commits (checksum)


pkg/devices/devices.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ReturnDevicesByTag will return the list of devices by tag
+func ReturnDevicesByTag(w http.ResponseWriter, r *http.Request) (Inventory, error) {
+
+	tags := chi.URLParam(r, "devices")
+	tagsParam := "?tags=" + tags
+
+	url := fmt.Sprintf("%s/api/inventory/v1/hosts", config.Get().InventoryConfig.URL)
+	fullURL := url + filterParams + tagsParam
+	fmt.Printf("Requesting url: %s\n", fullURL)
```
biancarosa, line 111: Can we use the logger instead of prints? :)

biancarosa: `log.Infof....`

pkg/devices/devices.go
```diff
@@ -0,0 +1,133 @@
...
+	fullURL := url + filterParams + tagsParam
+	log.Infof("Requesting url: %s\n", fullURL)
+	req, _ := http.NewRequest("GET", fullURL, nil)
+	req.Header.Add("Content-Type", "application/json")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return Inventory{}, err
+	}
```
chambridge, line 120: After reviewing the file, it feels like this section could be made into a common method for interacting with the host inventory.  Perhaps just taking a map of parameters to append to the path?

pkg/updates/updates.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+}
+
+func getDevices(w http.ResponseWriter, r *http.Request) {
+	fmt.Printf("getDevices")
+	devices, err := devices.ReturnDevices(w, r)
+	fmt.Printf("devices: %v\n", devices)
+	if err != nil {
+		return
+	}
```
chambridge, line 29: Need to handle the error condition and still reply with a response, otherwise you are just going to end up with a timeout 503 from the 3scale gateway.

pkg/updates/updates.go
```diff
@@ -0,0 +1,79 @@
...
+
+func getDevicesByID(w http.ResponseWriter, r *http.Request) {
+	uuid, tags := parseParams(r)
+	fmt.Printf("tags: %v\n", len(tags))
+	fmt.Printf("getDevicesById: %v\n", len(uuid))
+	if len(uuid) > 0 {
+		validUUID := isUUID(uuid)
+		if validUUID {
+			devices, err := devices.ReturnDevicesByID(w, r)
+			//FIXME: Load results into DB
```
chambridge, line 40: Are we really trying to store the results? Seems like we would just be duplicating the data in the host inventory, no?

pkg/updates/updates.go
```diff
@@ -0,0 +1,79 @@
...
+			fmt.Printf("validUuid devices: %v\n", devices)
+			if err != nil {
+				return
+			}
+		} else {
+			fmt.Printf("Invalid UUID")
+		}
```
chambridge, line 47: This should have an error response back, right? otherwise it will just time out rather than giving a 400 for bad request.

pkg/updates/updates.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/go-chi/chi"
+	"github.com/google/uuid"
+	"github.com/redhatinsights/edge-api/pkg/devices"
+)
+
+// MakeRouter adds support for operations on update
+func MakeRouter(sub chi.Router) {
+	sub.Get("/", getDevices)
+	sub.Route("/device_uuid={device_uuid}", func(r chi.Router) {
```
chambridge, line 15: I could be wrong but I would think this would just be:

```suggestion
	sub.Route("/{device_uuid}", func(r chi.Router) {
```

pkg/updates/updates.go
```diff
@@ -0,0 +1,79 @@
...
+		r.Get("/", getDevicesByID)
+	})
+	sub.Route("/tags={tags}", func(r chi.Router) {
```
chambridge, line 18: Is this meant to be a route or are you really just trying to handle query parameters?

chambridge (changes requested):
Need some updates to work when deployed and handle error cases.

main.go
```diff
@@ -90,6 +91,8 @@ func main() {
 		s.Route("/commits", commits.MakeRouter)
 		s.Route("/repos", repo.MakeRouter(server))
 		s.Route("/images", images.MakeRouter)
+		// s.Route("/devices", devices.MakeRouter)
```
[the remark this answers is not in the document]
acosferreira: No, I'll remove that, is not used route

pkg/updates/updates.go
maxamillion: @chambridge we need a record of which update correlates to which devices and we also need to store the device uuids so we can send Playbook Dispatcher payloads. If there's a better way to accomplish this then that would be awesome, but I don't know of any way.

pkg/devices/devices.go
maxamillion: Agreed, let's move this to `pkg/common/`

pkg/updates/updates.go
maxamillion, line 16: @chambridge I'd suggest it be namespaced like this:
```suggestion
	sub.Route("/device/{device_uuid}", func(r chi.Router) {
```

maxamillion: actually on second thought it probably should just be
```suggestion
	sub.Route("/", func(r chi.Router) {
```
and handle params in the context manager

pkg/updates/updates.go
chambridge: I'm good with storing device ids --> update relationship. The FIXME didn't give a lot of clarity around what is being stored. I didn't want to be storing duplicate data.

Depending on whether the update information can/will be contained in the inventory down the line you could look at 
[Host Data Syndication](https://clouddot.pages.redhat.com/docs/dev/services/inventory.html#cyndi) to give you this relationship.

pkg/updates/updates.go
```diff
@@ -0,0 +1,84 @@
...
+			fmt.Printf("validUuid devices: %v\n", devices)
+			if err != nil {
+				err := errors.NewInternalServerError()
+				err.Title = fmt.Sprintf("Failed to get device %s", uuid)
+				w.WriteHeader(err.Status)
+				return
+			}
+			json.NewEncoder(w).Encode(&devices)
+		} else {
+			err := errors.NewInternalServerError()
```
biancarosa, line 53: I'd do a errors.NewBadRequest here because it seems like an error related to the user input

pkg/devices/devices.go
```diff
@@ -0,0 +1,152 @@
...
+	inventoryAPI = "api/inventory/v1/hosts"
+	orderBy      = "updated"
+	orderHow     = "DESC"
+	filterParams = "?filter[system_profile][host_type]=edge&fields[system_profile]=host_type,operating_system,greenboot_status,greenboot_fallback_detected,rpm_ostree_deployments"
+)
+
+// ReturnDevices will return the list of devices without filter by tag or uuid
+func ReturnDevices(w http.ResponseWriter, r *http.Request) (Inventory, error) {
+	url := fmt.Sprintf("%s/api/inventory/v1/hosts", config.Get().InventoryConfig.URL)
+	fullURL := url + filterParams
```
boaz0, line 45: small nitpick => replacing this two lines to
```go
url := fmt.Sprintf("%s/api/inventory/v1/hosts%s", config.Get().InventoryConfig.URL, filterParams)
```

biancarosa (approved):
It looks good to me, but I'd wait for @chambridge since he requested more changes :)

boaz0 (commented):
All in all LGTM

maxamillion:
LGTM :+1:

pkg/updates/updates.go
maxamillion: @chambridge ah, sounds good ... yeah, all we're doing is storing the UUID from inventory and then a couple bits of info that are edge-api centric -> https://github.com/RedHatInsights/edge-api/blob/655256c8230cb8efd0bd91f6563c923d83cb1203/pkg/models/updates.go#L18

chambridge (approved):

chambridge:
It appears to be crashlooping in the migrate.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x85053e]

goroutine 1 [running]:
github.com/redhatinsights/edge-api/config.Init()
	/src/mypackage/myapp/config/config.go:109 +0xafe
main.main()
	/src/mypackage/myapp/cmd/migrate/migrate.go:12 +0x3
```

maxamillion:
> It appears to be crashlooping in the migrate.
> 
> ```
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x85053e]
> 
> goroutine 1 [running]:
> github.com/redhatinsights/edge-api/config.Init()
> 	/src/mypackage/myapp/config/config.go:109 +0xafe
> main.main()
> 	/src/mypackage/myapp/cmd/migrate/migrate.go:12 +0x3
> ```

This PR doesn't touch the migration though ... what could cause that? Is it angry at changes to a struct somewhere?

config/config.go
```diff
@@ -58,6 +63,7 @@ func Init() {
 	options.SetDefault("Debug", false)
 	options.SetDefault("EdgeTarballsBucket", "rh-edge-tarballs")
 	options.SetDefault("ImageBuilderUrl", "http://image-builder:8080")
+	options.SetDefault("InventoryUrl", "https://host-inventory-service:8080/")
```
chambridge, line 66:
```suggestion
	options.SetDefault("InventoryUrl", "http://host-inventory-service:8080/")
```

biancarosa:
@acosferreira the ephemeral env is fine now, can you rebase? :)

PR #75, opened 2021-07-08
Fix rhc-worker-playbook package name

biancarosa:
The package name was wrong in the JIRA ticket and I was stupid enough not to check 🤦🏻


AaronH88 (approved):
Looks good to me
+1

PR #80, opened 2021-07-09
Validate create image name

resoluteCoder:
Added name validation for image with regex same as front end

### DIFF
--- a/pkg/models/images.go
+++ b/pkg/models/images.go
@@ -28,7 +28,7 @@ const (
 	// ArchitectureCantBeEmptyMessage is the error message when the architecture is empty
 	ArchitectureCantBeEmptyMessage = "architecture can't be empty"
 	// NameCantBeInvalidMessage is the error message when the name is invalid
-	NameCantBeInvalidMessage = "invalid input for name"
+	NameCantBeInvalidMessage = "name must start with alphanumeric characters and can contain underscore and hyphen characters"
 
 	// ImageTypeInstaller is the installer image type on Image Builder
 	ImageTypeInstaller = "rhel-edge-installer"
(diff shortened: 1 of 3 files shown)


pkg/models/images.go
```diff
@@ -26,6 +27,8 @@ const (
 	DistributionCantBeNilMessage = "distribution can't be empty"
 	// ArchitectureCantBeEmptyMessage is the error message when the architecture is empty
 	ArchitectureCantBeEmptyMessage = "architecture can't be empty"
+	// NameCantBeInvalidMessage is the error message when the name is invalid
+	NameCantBeInvalidMessage = "invalid input for name"
```
boaz0, line 31: How about "name must start with alphanumeric characters and can contain underscore and hyphen characters"

pkg/models/images.go
```diff
@@ -47,6 +50,9 @@ func (i *Image) ValidateRequest() error {
 	if i.Distribution == "" {
 		return errors.New(DistributionCantBeNilMessage)
 	}
+	if re := regexp.MustCompile(`^[A-Za-z0-9]+[A-Za-z0-9\s_-]*$`).FindString(i.Name); re == "" {
```
boaz0, line 53: I would rather do something else than `regexp` but I guess `regexp` is much cleaner.
Anyway, why not using `MatchString(i.Name)`, that way you don't need `re`:
```go
 	if regexp.MustCompile(`^[A-Za-z0-9]+[A-Za-z0-9\s_-]*$`).MatchString(i.Name) {
```

boaz0 (changes requested):
Please add unit tests as well!

pkg/models/images.go
resoluteCoder: With FindString I can get a value back that can be checked for false.

Well, I suppose I can write `MatchString(i.Name) == false` or 
` if !regexp.MustCompile(pattern).MatchString(i.Name) {` and adding the `!`

What do you think?

boaz0: Yes that's what I meant.

biancarosa (approved):

PR #83, opened 2021-07-09
Changed DESC to ASC to fix asc sort issue

matoval:
I changed the default value to ASC from DESC so that asc sorting will work. The initial API call from the front-end is still created_at,
desc when the manage images table loads.


boaz0:
@matoval can you [squash](https://www.internalpointers.com/post/squash-commits-into-one-git) your commits into one?

boaz0 (commented):
LGTM :+1:
@biancarosa can you give a look?
Thanks

biancarosa (approved):

PR #85, opened 2021-07-12
Feat: Add objects on bucket with public read acl

biancarosa:


maxamillion (approved):
Reluctantly +1 - we have to do this temporarily to make hosted image builder happy.

PR #98, opened 2021-07-14
Playbook template

acosferreira:
Init version


maxamillion:
This is great, this is without a doubt the groundwork for what we need. Thank you! :+1:

pkg/updates/template/template.go
```diff
@@ -0,0 +1,55 @@
...
+	GoTemplateGpgVerify  string
+	OstreeRemoteName     string
+	OstreeRemoteURL      string
+	OstreeContentURL     string
+	OstreeGpgVerify      string
+	OstreeGpgKeypath     string
+	OstreeRemoteTemplate string
+}
+
+func main() {
```
maxamillion, line 23: I've been thinking about this lately and I think we might want this to be a library function ... something like `TemplatePlaybook` that will accept in the `remoteName`, `remoteURL`, and `contentURL` as parameters, then template the playbook and upload it into our S3 bucket under some new path ... `/templates/` or something. We'll probably want a database table to keep tabs on these things by ID and their resulting URL in the S3 bucket.

pkg/commits/commits_test.go
```diff
@@ -34,7 +35,7 @@ func setUp() {
 }
 
 func tearDown() {
-	db.DB.Delete(&cmt)
+	db.DB.Exec("DELETE FROM commits")
```
maxamillion, line 38: the `db.DB.Delete(&cmt)` appears to be referencing an existing commit but `db.DB.Exec("DELETE FROM commits")` is not .... is this missing a param passed to it or am I confused on what that `Exec` does?

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+	GoTemplateGpgVerify  string
+	OstreeRemoteName     string
+	OstreeRemoteURL      string
+	OstreeContentURL     string
+	OstreeGpgVerify      string
+	OstreeGpgKeypath     string
+	OstreeRemoteTemplate string
+}
+
+type TemplateRemoteInfo struct {
```
maxamillion, line 28: Comments on exported types to make `go vet` happy please. :)

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func WriteTemplate(tempalteInfo TemplateRemoteInfo) (string, error) {
+	log.Debugf("::WriteTemplate: BEGIN")
+	filePath := "pkg/playbooks/"
+	templateName := "template_playbook_dispatcher_ostree_upgrade_payload.yml"
+	template, err := template.ParseFiles(filePath + templateName)
+	if err != nil {
+		fmt.Println(err)
+		return "Playbook::WriteTemplate::Error::", err
```
maxamillion, line 49: I think this should be:
```go
if err != nil {
    log.Errorf("Playbook::WriteTemplate::Error: %#v", err)
    return "", err
}
```

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+		OstreeGpgVerify:      "true",
+		OstreeGpgKeypath:     "/etc/pki/rpm-gpg/",
+		OstreeRemoteTemplate: "{{ ostree_remote_template }}"}
+
+	fname := fmt.Sprintf("playbook_dispatcher_update_%v", tempalteInfo.UpdateTransaction) + ".yml"
+	path := filePath + fname
+	f, err := os.Create(path)
+	if err != nil {
+		log.Println("create file: ", err)
+		return "Playbook::WriteTemplate::Error::", err
```
maxamillion, line 68: Same here as above to log the error and `return "", err`

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+	UpdateTransaction int
+}
+
+type S3Uploader struct {
+	Client            *s3.S3
+	S3ManagerUploader *s3manager.Uploader
+	Bucket            string
+}
+
+func WriteTemplate(tempalteInfo TemplateRemoteInfo) (string, error) {
```
biancarosa, line 42:
```suggestion
func WriteTemplate(templateInfo TemplateRemoteInfo) (string, error) {
```

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+	uploader = &commits.FileUploader{
+		BaseDir: path,
+	}
+	if cfg.BucketName != "" {
+		uploader = commits.NewS3Uploader()
+	}
+	repoURL, err := uploader.UploadRepo(path, string(tempalteInfo.UpdateTransaction))
+	if err != nil {
+		log.Println("create file: ", err)
+		return "Playbook::WriteTemplate::Error::", err
```
maxamillion, line 83: Same here as above to log the error and `return "", err`

pkg/playbooks/playbooks.go
```diff
@@ -0,0 +1,91 @@
...
+	log.Debugf("::WriteTemplate: BEGIN")
+	filePath := "pkg/playbooks/"
+	templateName := "template_playbook_dispatcher_ostree_upgrade_payload.yml"
+	template, err := template.ParseFiles(filePath + templateName)
+	if err != nil {
+		fmt.Println(err)
+		return "Playbook::WriteTemplate::Error::", err
+	}
+	templateData := playbooks{
+		GoTemplateRemoteName: tempalteInfo.RemoteName,
```
biancarosa, line 52: Instead of copying all of those fields, can you just create a templateInfo field that receives directly the templateInfo object?

maxamillion (changes requested):

pkg/playbooks/playbooks.go
biancarosa, line 68: Is that string used for something in the code that calls it? Best practice is to normally return empty object (nil or "") and err when a method returns an error.

pkg/commits/commits_test.go
acosferreira: for some reason (that I don't know) the data wasn't being deleted and the test was failing. The exec will delete everything from this table.

pkg/playbooks/playbooks.go
acosferreira: This is the way to parse the values inside of the template file.

pkg/commits/commits_test.go
maxamillion: 👍

PR #103, opened 2021-07-15
Left join on commit and installer on images list

biancarosa:
Fixes https://issues.redhat.com/browse/RHELPLAN-89740

### DIFF
--- a/pkg/common/filters.go
+++ b/pkg/common/filters.go
@@ -12,11 +12,16 @@ import (
 // FilterFunc is a function that takes http request and GORM DB adds a query according to the request
 type FilterFunc func(r *http.Request, tx *gorm.DB) *gorm.DB
 
+type Filter struct {
+	QueryParam string
+	DBField    string
+}
+
 // ContainFilterHandler handles sub string values
-func ContainFilterHandler(name string) FilterFunc {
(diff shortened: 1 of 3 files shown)


maxamillion (approved):

maxamillion:
LGTM :+1:

pkg/common/filters.go
```diff
@@ -12,22 +12,27 @@ import (
...
 			tx = tx.Where(sqlQuery, "%"+val+"%")
 		}
 		return tx
 	})
 }
 
 // OneOfFilterHandler handles multiple values filters
-func OneOfFilterHandler(name string) FilterFunc {
-	sqlQuery := fmt.Sprintf("%s IN ?", name)
+func OneOfFilterHandler(filter *Filter) FilterFunc {
```
boaz0, line 32: Is there a reson we pass this value as a ref? Unless Filter is too big or is going to be modified I normally pass it as value.

boaz0 (approved):
LGTM 2 :+1:

pkg/common/filters.go
biancarosa: No particular reason, I just normally pass it as ref because of the slightly better performance, but I'm happy to change to pass as a value. We can take that conversation to a team meeting because we should create some sort of guidelines for that :)

PR #104, opened 2021-07-18
WIP - refactoring updates

maxamillion:

### DIFF
--- a/pkg/commits/repobuilder.go
+++ b/pkg/commits/repobuilder.go
@@ -1,21 +1,14 @@
 package commits
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
-	"errors"
 	"fmt"
-	"io"
-	"net/http"
+
+	"bytes"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
-
-	"github.com/go-chi/chi"
 
 	"github.com/redhatinsights/edge-api/config"
 	"github.com/redhatinsights/edge-api/pkg/common"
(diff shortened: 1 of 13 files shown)


maxamillion:
@biancarosa @acosferreira I didn't get a chance to actually test this so I'm not sure where there be dragons just yet, but this is the changeset I had in mind for the refactor plus/minus bugfixes after I get some testing in.

pkg/models/commits.go
```diff
@@ -20,6 +20,15 @@ type Commit struct {
 	Packages             []Package `json:"Packages" gorm:"many2many:commit_packages;"`
 	ComposeJobID         string    `json:"ComposeJobID"`
 	Status               string    `json:"Status"`
+	RepoID               uint      `json:"RepoID"`
```
biancarosa, line 23:
```suggestion
	RepoID               *uint      `json:"RepoID"`
```

This one needs to be a pointer, running against postgresql it violates a fk contrainst.

pkg/commits/repobuilder.go
```diff
@@ -33,184 +26,24 @@ func InitRepoBuilder() {
...
-func (rb *RepoBuilder) BuildRepo(ur *models.UpdateRecord) (*models.UpdateRecord, error) {
+func (rb *RepoBuilder) BuildRepo(ur *models.UpdateTransaction) (*models.UpdateTransaction, error) {
 	cfg := config.Get()
 
-	var updaterec models.UpdateRecord
-	db.DB.First(&updaterec, ur.ID)
-	updaterec.Status = models.UpdateStatusCreated
-	db.DB.Save(&updaterec)
+	var update models.UpdateTransaction
+	db.DB.First(&update, ur.ID)
```
biancarosa, line 44: We can rename ur to ut here

pkg/commits/repobuilder.go
```diff
@@ -33,184 +26,24 @@ func InitRepoBuilder() {
...
+	update.Status = models.UpdateStatusCreated
+	db.DB.Save(&update)
```
biancarosa, line 46: It might be good to check if there were any errors in the resulting transaction to avoid weird problems further down the process

pkg/commits/repobuilder.go
```diff
@@ -265,19 +98,58 @@ func (rb *RepoBuilder) BuildRepo(ur *models.UpdateRecord) (*models.UpdateRecord,
...
-	var updateRecDone models.UpdateRecord
-	db.DB.First(&updateRecDone, ur.ID)
-	updateRecDone.Status = models.UpdateStatusSuccess
-	updateRecDone.UpdateRepoURL = repoURL
-	db.DB.Save(&updateRecDone)
+	var updateDone models.UpdateTransaction
+	db.DB.First(&updateDone, ur.ID)
+	updateDone.Status = models.UpdateStatusSuccess
+	updateDone.Commit.Repo.URL = repoURL
+	db.DB.Save(&updateDone)
```
biancarosa, line 111: Same here. I'm not too worried about errors on the select because it needs an existing UpdateTransaction, but saves are dangerous

pkg/commits/repobuilder.go
```diff
@@ -265,19 +98,58 @@ func (rb *RepoBuilder) BuildRepo(ur *models.UpdateRecord) (*models.UpdateRecord,
...
+	log.Debugf("RepoBuilder::path: %#v", path)
+	err := os.MkdirAll(path, os.FileMode(int(0755)))
+	if err != nil {
+		return err
+	}
+	err = os.Chdir(path)
+	if err != nil {
+		return err
+	}
+	DownloadExtractVersionRepo(c, path)
```
biancarosa, line 130: I just realized we're not checking errors here and things can go south silently

pkg/updates/updates.go
```diff
@@ -3,150 +3,242 @@ package updates
...
 	}
 
-	// FIXME - need to remove duplicate OldCommit values from UpdateRecord
+	log.Debugf("updateFromHTTP::update: %#v", update)
+	return &update, err
+}
+
+type key int
+
+const UpdateContextKey key = 0
```
biancarosa, line 143: I might be wrong on how context keys work, but I think we need a different key than 0 for UpdateContextKey because we're using 0 for commits (and I just realized that image and pagination are sharing one, so this can be weird if we use then in the same request).

Maybe it'd be best to centralize all context keys in a package to make it easier to read which ones are in use and you don't run the risk of trying to access an object and accessing another.

biancarosa (changes requested):

pkg/updates/updates.go
maxamillion: I honestly don't know, I just assumed since they were each a different context that it would be fine ... is that all a shared instantiation? (my understand of how the context works is very introductory)

biancarosa: yeah, we have one context per request and it's shared, unless you specifically create a new context for a specific transaction

biancarosa (commented):
Now I think that, with my suggestions, the references to Repo.ID are breaking because they aren't expecting a pointer to a uint, but a uint

biancarosa (approved):
LGTM! Awesome refactor, this is great :)

PR #110, opened 2021-07-21
playbook dispatcher book keeping

maxamillion:
This is the data model and initial book keeping mechanism for scheduling
and queuing playbook dispatcher jobs persistently.

### DIFF
--- a/pkg/updates/updates.go
+++ b/pkg/updates/updates.go
@@ -107,6 +107,7 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) (*models.UpdateTrans
 	// Get the models.Commit from the Commit ID passed in via JSON
 	update.Commit, err = common.GetCommitByID(updateJSON.CommitID)
 	log.Debugf("updateFromHTTP::update.Commit: %#v", update.Commit)
+	update.DispatchRecords = []models.DispatchRecord{}
 	if err != nil {
 		err := apierrors.NewInternalServerError()
 		err.Title = fmt.Sprintf("No commit found for CommitID %d", updateJSON.CommitID)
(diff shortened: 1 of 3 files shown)


pkg/updates/updates.go
```diff
@@ -170,9 +172,20 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) (*models.UpdateTrans
...
-		inventoryHosts = append(inventoryHosts, *updateDevice)
-		log.Debugf("updateFromHTTP::inventoryHosts: %#v", inventoryHosts)
-		update.InventoryHosts = inventoryHosts
+		devices = append(devices, *updateDevice)
+		log.Debugf("updateFromHTTP::devices: %#v", devices)
+		update.Devices = devices
+		log.Debugf("updateFromHTTP::update.Devices: %#v", devices)
+
+		dispatchRecord := new(models.DispatchRecord)
+		dispatchRecord.Device = updateDevice
```
acosferreira, line 181: Here you can use the repoURL from the upload

pkg/updates/updates.go
```diff
@@ -170,9 +172,20 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) (*models.UpdateTrans
...
 		log.Debugf("updateFromHTTP::updateDevice: %#v", updateDevice)
-		inventoryHosts = append(inventoryHosts, *updateDevice)
-		log.Debugf("updateFromHTTP::inventoryHosts: %#v", inventoryHosts)
-		update.InventoryHosts = inventoryHosts
+		devices = append(devices, *updateDevice)
+		log.Debugf("updateFromHTTP::devices: %#v", devices)
+		update.Devices = devices
+		log.Debugf("updateFromHTTP::update.Devices: %#v", devices)
+
+		dispatchRecord := new(models.DispatchRecord)
```
biancarosa, line 180: We actually do this like...:
```
&models.DispatchRecord{
      Device: updateDevice,
}
```
Makes it a little bit more easier to read, I guess. Just a suggestion though.

biancarosa: (with all of the fields, I didn't copy them all because I was lazy)

maxamillion: I like that, thanks!

biancarosa (approved):

acosferreira (approved):

PR #116, opened 2021-07-22
Adding username and ssh key into a kickstart file, and removing that file after use.

AaronH88:
Issue RHELPLAN-90198

Adding username and ssh key into the image installer model.
Passing this information into a custom kickstart file by using go templating.
Created a cleanup function that removes the kickstart after use.


pkg/models/installers.go
```diff
@@ -9,4 +9,6 @@ type Installer struct {
 	ImageBuildISOURL string `json:"ImageBuildISOURL"`
 	ComposeJobID     string `json:"ComposeJobID"`
 	Status           string `json:"Status"`
+	Username         string `json:"User name"`
+	Sshkey           string `json:"Ssh Key"`
```
biancarosa, line 13:
```suggestion
	SSHKey           string `json:"SshKey"`
```

biancarosa: Go's best practices normally ask us to keep acronyms uppercase, and maybe we should avoid spaces on JSON keys. This last one is purely feeling, though, I am not sure about best practices around that. Maybe @boaz0 can clear this out.

biancarosa (commented):
Tiny comment, but it looks pretty good.

pkg/images/images.go
```diff
@@ -490,8 +499,84 @@ func CreateInstallerForImage(w http.ResponseWriter, r *http.Request) {
...
+		}
 	}(image.ID)
 
 	w.WriteHeader(http.StatusOK)
 	json.NewEncoder(w).Encode(&image)
 }
+
+// Download the ISO, inject the kickstart with username and ssh key
+// re upload the ISO
+func addUserInfo(image *models.Image, w http.ResponseWriter) error {
```
biancarosa, line 521: Oh, just realized you're returning the error and handling down there, so we don't need ResponseWriter here and to handle json in this function, the route function can do that if it gets an error and it's doing already :)

pkg/models/installers.go
```diff
@@ -9,4 +9,6 @@ type Installer struct {
 	ImageBuildISOURL string `json:"ImageBuildISOURL"`
 	ComposeJobID     string `json:"ComposeJobID"`
 	Status           string `json:"Status"`
+	Username         string `json:"User name"`
```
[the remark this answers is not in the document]
AaronH88: Applied

pkg/models/installers.go
AaronH88: Applied

pkg/images/images.go
AaronH88: Nice catch, removed!

pkg/models/installers.go
boaz0: @biancarosa is correct. You can see for example [here](https://pkg.go.dev/encoding/json@go1.16.6#RawMessage.MarshalJSON) instead of `MarshalJson` it's `MarshalJSON`.
It's kinda unwritten rule in Go but conventionally acronyms like `API`, `HTTP`, `ID`, etc are kept in their "original" form.

biancarosa (approved):
LGTM!

PR #118, opened 2021-07-22
Return correct object on repos endpoint

biancarosa:


pkg/images/images.go
```diff
@@ -508,6 +508,6 @@ func GetRepoForImage(w http.ResponseWriter, r *http.Request) {
 			json.NewEncoder(w).Encode(&err)
 			return
 		}
-		json.NewEncoder(w).Encode(image)
+		json.NewEncoder(w).Encode(repo)
```
boaz0, line 511: is `repo` in the scope? because I am afraid if it's under the scope of the loop above then it's not available here.
But I guess I am wrong because it compiles... SO nevermind (me thinking loud)/

boaz0 (approved):
LGTM :+1:

pkg/images/images.go
biancarosa: It is! It's created inside of the this if block!

boaz0: Oh you're right! Sorry, my eyes missed that. Thanks!

PR #119, opened 2021-07-22
Added function in go to call fleetkick.sh. Created and cleaned a workdir for feetkick.sh

AaronH88:
Feature RHELPLAN-88548 & RHELPLAN-90204

Create a script that can inject a kickstart into an ISO,
Call the script in the correct locations.

Co-author @loadtheaccumulator


pkg/images/images.go
```diff
@@ -490,8 +492,52 @@ func CreateInstallerForImage(w http.ResponseWriter, r *http.Request) {
...
+	absPath, err := filepath.Abs(".")
+	if err != nil {
+		return err
+	}
+
+	kickstart = absPath + kickstart
+	image = absPath + image
+	fleetBashScript := absPath + "fleetkick.sh"
+
+	cmd := exec.Command(fleetBashScript, kickstart, image, image, ".")
```
loadtheaccumulator, line 536: "." will need to be a subdir instead of . to avoid additional content getting rolled into the ISO

pkg/images/images.go
```diff
@@ -646,3 +652,23 @@ func GetRepoForImage(w http.ResponseWriter, r *http.Request) {
...
+// Inject the custom kickstart into the iso via mkksiso.
+func exeInjectionScript(kickstart string, image string) error {
+	absPath, err := filepath.Abs(".")
+	if err != nil {
+		return err
+	}
+
+	kickstart = absPath + kickstart
+	image = absPath + image
+	fleetBashScript := absPath + "fleetkick.sh"
```
loadtheaccumulator, line 665: fleet kick.sh isn't included in the package on compile, right? I'm adding this to be copied to /usr/local/bin in the Dockerfile under another PR.

So this would change to...
   fleetBashScript := "/usr/local/bin/fleetkick.sh"

loadtheaccumulator (changes requested):

loadtheaccumulator:
@AaronH88 How do you feel about abandoning this PR (or just removing the fleetkick.sh file from it? I need to include it with the Dockerfile change so it won't break the build.

pkg/images/images.go
AaronH88: Good idea!

AaronH88:
> @AaronH88 How do you feel about abandoning this PR (or just removing the fleetkick.sh file from it? I need to include it with the Dockerfile change so it won't break the build.

Which ever is easier!

biancarosa:
We have some conflicts here, so if @loadtheaccumulator gets his PR #135 working he might get this piece of code on his PR too

loadtheaccumulator:
I resolved the conflicts, but seeing different errors.

biancarosa (approved):
LGTM!

loadtheaccumulator (approved):

PR #122, opened 2021-07-22
Playbook dispatcher library

acosferreira:


config/config.go
```diff
@@ -8,24 +8,26 @@ import (
...
+	BucketRegion             *string
+	AccessKey                string
+	SecretKey                string
+	RepoTempPath             string
+	OpenAPIFilePath          string
+	ImageBuilderConfig       *imageBuilderConfig
+	InventoryConfig          *inventoryConfig
+	S3ProxyURL               string
+	DefaultOSTreeRef         string
+	PlaybookDispatcherPSK    string
```
biancarosa, line 29: Can the PSK go inside of the playbookDispatcherConfig?

pkg/playbooks/playbook_dispatcher.go
```diff
@@ -0,0 +1,50 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type DispatcherPayload struct {
+	Recipient   string
+	PlaybookURL string
+	Account     string
+}
+
+func ExecuteDispatcher(w http.ResponseWriter, r *http.Request, payload DispatcherPayload) {
```
biancarosa, line 21: Can we avoid receiving the http.ResponseWriter and http.Request here? We can do this essentially receiving the outgoing headers and returning the error so that the route can handle all the status codes.

Let's say, if I wanna call playbook dispatcher in a go routine, I should be able to without further modification here, but this code is tightly coupled to the request and response.

acosferreira: Agreed! should be able to just pass the payload. Doing this change now ;)

biancarosa (approved):

PR #125, opened 2021-07-23
Added absolute path for templateKickstart file.

AaronH88:
Fix for bug, unable to find templateKickstart.ks


biancarosa (approved):
LGTM, lets test on stage and see what happens 🙈

PR #131, opened 2021-07-26
Remove api/edge/v1 path from API docs since these are already the v1 docs

biancarosa:
This was causing duplication on the path on c.rh.c.


thom-at-redhat (approved):
lgtm with notes:
1) In the future, can we change the openapi.json to be jq output type format?  It's hard to review when it's all on one line
2) There are a number of Check annotations in the unchanged files.  Are any of these significant?

biancarosa:
- In the future, can we change the openapi.json to be jq output type format? It's hard to review when it's all on one line

We can, but it's automatically generated and we shouldn't have to read it 🤔 Maybe a validator to see if its format is okay? Or do you think we really need to read the output and check if it's correct? If so we can format the generated output

- There are a number of Check annotations in the unchanged files. Are any of these significant?

They are mostly to enforce comments on exported functions and structs and are highly recommended by Go's best practices

PR #132, opened 2021-07-26
Feat: Add create repo for image to help debug

biancarosa:
Co-author: @AaronH88 

This PR adds the create repo endpoint and an endpoint to add kickstart files to ISOs. They shouldn't be needed in the long run, but until we improve the dev flow I think we need to have those for our sanity.


acosferreira (approved):
lgtm

PR #133, opened 2021-07-26
get the url after uploading the file to S3

biancarosa:
This was breaking during local development, we need to get the BucketRegion not from the config, but from the AWS client because if cfg.Debug is true, we're doing:
```
	if cfg.Debug {
		sess = session.Must(session.NewSessionWithOptions(session.Options{
			// Force enable Shared Config support
			SharedConfigState: session.SharedConfigEnable,
		}))
	} else {
		var err error
		sess, err = session.NewSession(&aws.Config{
			Region:      cfg.BucketRegion,
			Credentials: credentials.NewStaticCredentials(cfg.AccessKey, cfg.SecretKey, ""),
		})
		if err != nil {
			panic(err)
		}
	}
```

which means that we get the region from the shared config - and the AWS_REGION environment variable - that is not used for production.

### DIFF
--- a/pkg/commits/uploader.go
+++ b/pkg/commits/uploader.go
@@ -18,7 +18,7 @@ import (
 //Uploader is an interface for uploading repository
 type Uploader interface {
 	UploadRepo(src string, account string) (string, error)
-	UploadFile(fname string, uploadPath string) error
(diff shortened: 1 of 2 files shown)


AaronH88 (approved):
Looks good to me!

PR #134, opened 2021-07-26
Update list delta diff

acosferreira:
Initial version to return the list of packages added and removed from update images.
This PR should be used to https://issues.redhat.com/browse/RHELPLAN-88855


matoval (approved):
LGTM

pkg/updates/updates.go
```diff
@@ -99,6 +101,11 @@ type UpdatePostJSON struct {
 	DeviceUUID string `json:"DeviceUUID"`
 }
 
+type DeltaDiff struct {
+	Added   []string
```
maxamillion, line 105: I think it would be a good idea to have DeltaDiff be `[]models.Package` instead of `[]string`. Thoughts @acosferreira @biancarosa ?

biancarosa: Yeah, I like that idea. Also, is this struct going to be used outside of this package? We can make it private by naming it `deltaDiff`. If we are gonna use it outside, I'd rename to `PackagesDeltaDiff` to make it clear across the codebase,

acosferreira: @maxamillion the GetPackagesList() is returning  *[]string
Should I change it as well?
cc @biancarosa

biancarosa: In that case, no, because it's just a function that given a list of packages returns a string. If you don't it as a string you can take the packages list as it is, without calling the function

biancarosa (approved):

PR #135, opened 2021-07-27
add fleetkick script and update Dockerfile

loadtheaccumulator:
* Add fleetkick.sh script
* Update Dockerfile to copy fleetkick.sh to /usr/local/bin
* Update Dockerfile to copy templateKickstart.ks to /usr/local/etc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,14 +19,19 @@ RUN CGO_ENABLED=0 go build -o /go/bin/edge-api-migrate cmd/migrate/migrate.go
 ############################
 # STEP 2 build a small image
 ############################
-FROM registry.redhat.io/ubi8-minimal:latest
+#FROM registry.redhat.io/ubi8-minimal:latest
+FROM quay.io/loadtheaccumulator/ubi8-isotools:latest
 
 COPY --from=builder /go/bin/edge-api /usr/bin
 COPY --from=builder /go/bin/edge-api-migrate /usr/bin
 COPY --from=builder /src/mypackage/myapp/cmd/spec/openapi.json /var/tmp
 
 # kickstart inject requirements
-COPY pkg/images/templateKickstart.ks /usr/local/etc
+COPY --from=builder /src/mypackage/myapp/pkg/images/fleetkick.sh /usr/local/bin
+COPY --from=builder /src/mypackage/myapp/pkg/images/templateKickstart.ks /usr/local/etc
+
+#RUN microdnf install -y pykickstart mtools xorriso genisoimage syslinux isomd5sum file
+ENV MTOOLS_SKIP_CHECK=1
 
 USER 1001
 
--- a/pkg/images/fleetkick.sh
+++ b/pkg/images/fleetkick.sh
@@ -1,0 +1,165 @@
+#!/bin/bash
+: <<'_usage_'
+
+KS_FQPATH: Fully qualified path to kickstart file
+ISO_IN: Path to source ISO
+ISO_OUT: Path to target ISO
+WORKDIR: Path to the directory the source ISO will be exploded into.
+
+./fleetkick.sh kickstart-stage.ks \
+/var/lib/libvirt/images/composer-api-d83040bf-b410-4317-9121-1ee7bd726ac1-rhel84-boot.iso \
+/var/lib/libvirt/images/ksinjected/composer-api-d83040bf-b410-4317-9121-1ee7bd726ac1-rhel84-boot_notefi.iso \
+/home/root/edgeiso/download98a
+
+_usage_
+
+KS_FQPATH=$1
+ISO_IN=$2
+ISO_OUT=$3
+WORKDIR=$4
+
+[[ -e $KS_FQPATH ]] || (echo "ERROR: no kickstart file" && exit 1)
+[[ -e $ISO_IN ]] || (echo "ERROR: no source ISO file" && exit 1)
+
+ISOCFG="${WORKDIR}/isolinux/isolinux.cfg"
+EFICFG="${WORKDIR}/EFI/BOOT/grub.cfg"
+EFI_DIR="${WORKDIR}/EFI/BOOT"
+EFI_IMAGEPATH="${WORKDIR}/images/efiboot.img"
+KSFILE=`basename $KS_FQPATH`
+
+
+validate_kickstart() {
+    KS=$1
+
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -19,12 +19,18 @@ RUN CGO_ENABLED=0 go build -o /go/bin/edge-api-migrate cmd/migrate/migrate.go
...
-FROM registry.redhat.io/ubi8-minimal:latest
+FROM registry.redhat.io/ubi8:latest
 
 COPY --from=builder /go/bin/edge-api /usr/bin
 COPY --from=builder /go/bin/edge-api-migrate /usr/bin
 COPY --from=builder /src/mypackage/myapp/cmd/spec/openapi.json /var/tmp
 
+# kickstart inject requirements
+COPY pkg/images/fleetkick.sh /usr/local/bin
+COPY pkg/images/templateKickstart.ks /usr/local/etc
```
loadtheaccumulator, line 30: Changing this to copy from builder instead of external env. Testing and will push to PR

pkg/images/images.go
```diff
@@ -598,12 +598,11 @@ type UnameSsh struct {
 
 // Adds user provided ssh key to the kickstart file.
 func addSSHKeyToKickstart(sshKey string, username string, kickstart string) error {
-	cfg := config.Get()
-
+	kickTemplatePath := "/usr/local/etc/templateKickstart.ks"
```
AaronH88, line 601: Can we drop these changes to images.go please and continue to pull the templateKickstart location from the env vars.

This makes debugging locally easier without having to build via Docker.

pkg/images/images.go
```diff
@@ -598,12 +598,11 @@ type UnameSsh struct {
...
 	td := UnameSsh{sshKey, username}
 
-	log.Infof("Opening file %s", cfg.KickstartPath)
-	t, err := template.ParseFiles(cfg.KickstartPath)
+	log.Infof("Opening file %s", kickTemplatePath)
```
AaronH88, line 604: take from env var please

pkg/images/images.go
```diff
@@ -598,12 +598,11 @@ type UnameSsh struct {
...
+	t, err := template.ParseFiles(kickTemplatePath)
```
AaronH88, line 605: and here also, take from env var please

Dockerfile
```diff
@@ -19,14 +19,19 @@ RUN CGO_ENABLED=0 go build -o /go/bin/edge-api-migrate cmd/migrate/migrate.go
...
+#FROM registry.redhat.io/ubi8-minimal:latest
+FROM quay.io/loadtheaccumulator/ubi8-isotools:latest
 
 COPY --from=builder /go/bin/edge-api /usr/bin
 COPY --from=builder /go/bin/edge-api-migrate /usr/bin
 COPY --from=builder /src/mypackage/myapp/cmd/spec/openapi.json /var/tmp
 
 # kickstart inject requirements
-COPY pkg/images/templateKickstart.ks /usr/local/etc
+COPY --from=builder /src/mypackage/myapp/pkg/images/fleetkick.sh /usr/local/bin
```
AaronH88, line 30: we are sure that `/usr/local/*` exists on this docker image?

AaronH88 (changes requested):

Dockerfile
```diff
@@ -19,14 +19,19 @@ RUN CGO_ENABLED=0 go build -o /go/bin/edge-api-migrate cmd/migrate/migrate.go
 ############################
 # STEP 2 build a small image
 ############################
-FROM registry.redhat.io/ubi8-minimal:latest
+#FROM registry.redhat.io/ubi8-minimal:latest
+FROM quay.io/loadtheaccumulator/ubi8-isotools:latest
```
AaronH88, line 23: pulling from quay now? Should we set up a team account rather than pull from your own account?

pkg/images/fleetkick.sh
```diff
@@ -0,0 +1,165 @@
...
+        || (echo "ERROR: no $ISO file"; exit 1)
+    echo $volid
+}
+
+
+explode_iso() {
+    ISO=$1
+    DIR=$2
+
+    #[[ -e $ISO ]] && 7z x $ISO -o${DIR} || (echo "ERROR: no ISO" && exit 1)
```
AaronH88, line 53: Do you want to keep this commented out code in here?

pkg/images/images.go
loadtheaccumulator: Yeah. I don't remember cfg being in there when I changed that, so looks like some changes were in between now and then.

Dockerfile
loadtheaccumulator: This is temporary. We should be able to skip this step once we are on a Jenkins with subman creds.

omaciel: Could we use https://quay.io/repository/cloudservices/ eventually?

pkg/images/fleetkick.sh
loadtheaccumulator: I have to do another push, so I can remove it with it.

Dockerfile
loadtheaccumulator: Yes. If it sticks. I'm hoping this will go away by EOD today. Keeping this image up-to-date with ubi8 will be an extra step we don't want to mess with. We also haven't seen any of this work e2e yet.

Dockerfile
loadtheaccumulator: Absolutely...
[jholloway@storage-qe14 ~]$ podman run -it --rm quay.io/loadtheaccumulator/ubi8-isotools:latest /bin/bash
[root@c5dd75841e3f /]# ls /usr/local
bin  etc  games  include  lib  lib64  libexec  sbin  share  src

pkg/images/images.go
loadtheaccumulator: File removed from this PR.

Dockerfile
AaronH88: Beautiful!

Dockerfile
AaronH88: Nice, makes sense to me now, ty

AaronH88 (approved):
Looks good to me! OK to merge

biancarosa (approved):
LGTM! Great work here @loadtheaccumulator! 👏🏻

PR #136, opened 2021-07-27
Moved all ISO manipulation to the tmp dir and fixed pathing to that d…

AaronH88:
…ir, edited the image location in uploadISO to point to the modified ISO.

Fix for bug: RHELPLAN-91047

Using absolute paths and no longer assuming the current working dir.

### DIFF
--- a/pkg/images/images.go
+++ b/pkg/images/images.go
@@ -597,7 +597,7 @@ type UnameSsh struct {
 
 // Adds user provided ssh key to the kickstart file.
 func addSSHKeyToKickstart(sshKey string, username string, kickstart string) error {
-	kickTemplatePath := "/var/tmp/templateKickstart.ks"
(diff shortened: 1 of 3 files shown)


pkg/images/images.go
```diff
@@ -595,11 +597,7 @@ type UnameSsh struct {
 
 // Adds user provided ssh key to the kickstart file.
 func addSSHKeyToKickstart(sshKey string, username string, kickstart string) error {
-	absPath, err := filepath.Abs(".")
-	kickTemplatePath := absPath + "/pkg/images/templateKickstart.ks"
-	if err != nil {
-		return err
-	}
+	kickTemplatePath := "/usr/local/etc/templateKickstart.ks"
```
biancarosa, line 600: Can we get the current path if debug=true to avoid having to move this during local dev?

biancarosa (approved):

PR #144, opened 2021-07-28
Adding to get OSTreeCommit from imagebuilder 

Saurabh-Thakre:
Tried explaining issues and steps I've done to test in the comments. Please do let me know if additional info is needed.


biancarosa (commented):
You have:
`var i *models.Image`

but when you call `getImage(w, r)` you're assigning to the `image variable`.

Maybe you drop this `var i *models.Image` declaration and use image on your call on GetMetadata?

Saurabh-Thakre:
> You have:
> `var i *models.Image`
> 
> but when you call `getImage(w, r)` you're assigning to the `image variable`.
> 
> Maybe you drop this `var i *models.Image` declaration and use image on your call on GetMetadata?

Yo done, and its working like charm

pkg/images/images.go
```diff
@@ -182,6 +183,7 @@ func postProcessImage(id uint, headers map[string]string) {
 
 	// TODO: This is also where we need to get the metadata from image builder
 	// in a separate goroutine
+	go imagebuilder.Client.GetMetadata(i, headers)
```
biancarosa, line 186: Oh, we can actually move this to before we create the OSTree repo! After the for {} on line 166

biancarosa (approved):
LGTM!

PR #152, opened 2021-08-02
Feat: Mark images as error during sigint

biancarosa:


pkg/images/images.go
```diff
@@ -161,23 +164,52 @@ func createRepoForImage(i *models.Image) *models.Repo {
...
 	tx := db.DB.Save(i)
 	if tx.Error != nil {
 		panic(tx.Error)
 	}
-	panic(err)
+	if i.Commit != nil {
+		i.Commit.Status = models.ImageStatusError
+		tx := db.DB.Save(i.Commit)
+		if tx.Error != nil {
+			panic(tx.Error)
```
maxamillion, line 176: @biancarosa pardon my golang noob moment, but does this `panic` only affect the goroutine or will this halt the entire backend service?

biancarosa: it will halt the entire backend service! I think I do want it to panic if I can't even mark my image with an error status, something *very wrong* ocurred 🤔

maxamillion: noted :+1:

PR #161, opened 2021-08-04
Add Packages Metadata from imagebuilder

Saurabh-Thakre:
Signed-off-by: Saurabh Thakre sthakre@redhat.com

### DIFF
--- a/pkg/imagebuilder/client.go
+++ b/pkg/imagebuilder/client.go
@@ -341,23 +341,18 @@ func (c *ImageBuilderClient) GetMetadata(image *models.Image, headers map[string
 	}
 
 	var ostree_struct OsTree
-	installed_package := models.Commit{}
 	json.Unmarshal(data, &ostree_struct)
-	json.Unmarshal(data, &installed_package)
 	for n := range ostree_struct.InstalledPackages {
-		installed_pkgs := []models.InstalledPackage{
+		pkg := models.InstalledPackage{
 
-			{
-				Arch: ostree_struct.InstalledPackages[n].Arch, Name: ostree_struct.InstalledPackages[n].Name,
-				Release: ostree_struct.InstalledPackages[n].Release, Sigmd5: ostree_struct.InstalledPackages[n].Sigmd5,
-				Signature: ostree_struct.InstalledPackages[n].Signature, Type: ostree_struct.InstalledPackages[n].Type,
-				Version: ostree_struct.InstalledPackages[n].Version, Epoch: ostree_struct.InstalledPackages[n].Epoch,
-			},
+			Arch: ostree_struct.InstalledPackages[n].Arch, Name: ostree_struct.InstalledPackages[n].Name,
+			Release: ostree_struct.InstalledPackages[n].Release, Sigmd5: ostree_struct.InstalledPackages[n].Sigmd5,
+			Signature: ostree_struct.InstalledPackages[n].Signature, Type: ostree_struct.InstalledPackages[n].Type,
+			Version: ostree_struct.InstalledPackages[n].Version, Epoch: ostree_struct.InstalledPackages[n].Epoch,
 		}
-		installed_package.InstalledPackage = append(installed_package.InstalledPackage, installed_pkgs...)
+		image.Commit.InstalledPackage = append(image.Commit.InstalledPackage, pkg)
 	}
 	image.Commit.OSTreeCommit = ostree_struct.OstreeCommit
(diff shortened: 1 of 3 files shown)


pkg/imagebuilder/client.go
```diff
@@ -103,8 +103,8 @@ type S3UploadStatus struct {
 
 // OsTree struct to get the metadata response : Saurabh
 type OsTree struct {
-	OstreeCommit string `json:"ostree_commit"`
-	Packages     InstalledPackage
+	OstreeCommit     string             `json:"ostree_commit"`
+	InstalledPackage []InstalledPackage `json:"packages"`
```
biancarosa, line 107:
```suggestion
	InstalledPackages []InstalledPackage `json:"packages"`
```

Can we call this field InstalledPackages to make it clear that is it a list?

pkg/imagebuilder/client.go
```diff
@@ -339,9 +339,28 @@ func (c *ImageBuilderClient) GetMetadata(image *models.Image, headers map[string
 	if err != nil {
 		return nil, err
 	}
+
 	var ostree_struct OsTree
+	countPackage := len(ostree_struct.InstalledPackage)
+	installed_package := models.Commit{}
```
biancarosa, line 345: Do we really need this? installed packages are already on the OsTree struct itself, one unmarshall should do the trick.

pkg/imagebuilder/client.go
```diff
@@ -339,9 +339,28 @@ func (c *ImageBuilderClient) GetMetadata(image *models.Image, headers map[string
...
+	n := 0
+	for n < countPackage {
```
biancarosa, line 347: Range (https://tour.golang.org/moretypes/16) is a nice way to loop without creating this variable.

pkg/imagebuilder/client.go
```diff
@@ -339,9 +339,28 @@ func (c *ImageBuilderClient) GetMetadata(image *models.Image, headers map[string
...
+				Arch: ostree_struct.InstalledPackage[n].Arch, Name: ostree_struct.InstalledPackage[n].Name,
+				Release: ostree_struct.InstalledPackage[n].Release, Sigmd5: ostree_struct.InstalledPackage[n].Sigmd5,
+				Signature: ostree_struct.InstalledPackage[n].Signature, Type: ostree_struct.InstalledPackage[n].Type,
+				Version: ostree_struct.InstalledPackage[n].Version, Epoch: ostree_struct.InstalledPackage[n].Epoch,
+			},
+		}
+		installed_package.InstalledPackage = installed_pkgs
+		n++
+	}
 	json.Unmarshal(data, &ostree_struct)
```
biancarosa, line 360: I think this Unmarshall needs to happen before this loop, otherwise ostree_struct won't be filled with the data.

pkg/imagebuilder/client.go
```diff
@@ -339,9 +339,28 @@ func (c *ImageBuilderClient) GetMetadata(image *models.Image, headers map[string
...
+		installed_pkgs := []models.InstalledPackage{
+
+			{
+				Arch: ostree_struct.InstalledPackage[n].Arch, Name: ostree_struct.InstalledPackage[n].Name,
+				Release: ostree_struct.InstalledPackage[n].Release, Sigmd5: ostree_struct.InstalledPackage[n].Sigmd5,
+				Signature: ostree_struct.InstalledPackage[n].Signature, Type: ostree_struct.InstalledPackage[n].Type,
+				Version: ostree_struct.InstalledPackage[n].Version, Epoch: ostree_struct.InstalledPackage[n].Epoch,
+			},
+		}
+		installed_package.InstalledPackage = installed_pkgs
```
biancarosa, line 357: I think we might need append here: https://tour.golang.org/moretypes/15

biancarosa (changes requested):

pkg/imagebuilder/client.go
Saurabh-Thakre: sure, makes sense

pkg/imagebuilder/client.go
Saurabh-Thakre: Yeah we would need **installed_package**, as we can not put **installed_pkg** in **ostree_struct.InstalledPackages** as it'd be of a different type.

biancarosa: But I think this should be created inside of the loop, right? Otherwise you're creating an empty object here and adding it to the list over an over.

biancarosa (approved):
Looks really good!

PR #163, opened 2021-08-04
return error status code on log; add psk to playbook-dispatcher call

acosferreira:


deploy/clowdapp.yaml
```diff
@@ -4,6 +4,12 @@ kind: Template
 metadata:
   name: edge-api
 objects:
+- apiVersion: cloud.redhat.com/v1alpha1
+  kind: ClowdApp
```
chambridge, line 8: Secret not defined correctly.

```suggestion
- apiVersion: v1
  kind: Secret
```

chambridge (changes requested):
Minor fix needed.

PR #179, opened 2021-08-06
add validation from rhc status when call the playbook-dispacther, sto…

acosferreira:
…re the playbook execution id

### DIFF
--- a/pkg/commits/repobuilder.go
+++ b/pkg/commits/repobuilder.go
@@ -3,6 +3,7 @@ package commits
 import (
 	"errors"
 	"fmt"
+	"net/http"
 
 	"bytes"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/commits/repobuilder.go
```diff
@@ -178,17 +181,32 @@ func (rb *RepoBuilder) BuildUpdateRepo(ut *models.UpdateTransaction) (*models.Up
...
-			log.Error(err)
+			log.Errorf("Error on playbook-dispatcher-executuin: %#v ", err)
 			return nil, err
 		}
-		//              Update/Create UpdateRecord.DispatchRecord
-		dispatchRecord := models.DispatchRecord{
-			Device:      &device,
-			PlaybookURL: "", // FIXME - need to populate this
-			Status:      exc,
+		for _, excPlaybook := range exc {
```
maxamillion, line 189: it looks like we're doing this exact logic here and in `pkg/updates/updates.go` below ... would it make sense to break this out into a function that call or is there contextual data that would be difficult to shepherd around?

maxamillion (approved):

maxamillion:
LGTM 👍

biancarosa (approved):

PR #190, opened 2021-08-11
Update how to generate API docs

biancarosa:


README.md
```diff
@@ -250,15 +250,17 @@ Golang also provides a unit test infrastructure `test`. Run unit tests with the
...
 ```
 
-## Generating API docs
+## API docs
 
-[go-swagger](https://github.com/go-swagger) is a tool that, amongst other things, parses comments in the code to generate a file in the [Open API 2.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md).
+[kin-openapi](https://github.com/getkin/kin-openapi) is a tool that helps us handle docs in the format of [Open API spec](https://github.com/OAI/OpenAPI-Specification). Sadly, it does not generate docs *automagically*. We have a [script](cmd/spec/main.go) that generates the docs and you have to add your model there to be picked by the code generation.
 
-To generate the API docs, you need to [install go-swagger](https://goswagger.io/install.html). Then you can use the following command:
+The [openapi3gen](https://github.com/getkin/kin-openapi/tree/v0.65.0/openapi3gen) package generates the docs for the models in the project and we have to update the routes by hand on the [path.yml](cmd/spec/path.yaml) file. Our generation [script](cmd/spec/main.go) adds the routes that you wrote by hand, creating an openapi.json and a openapi.yaml.
```
acosferreira, line 257: We should include the path.yml.

biancarosa: In what part exactly?

acosferreira (approved):
LGTM

PR #191, opened 2021-08-11
add create image to api doc

acosferreira:


cmd/spec/path.yaml
```diff
@@ -153,4 +153,31 @@ paths:
                 $ref: '#/components/schemas/v1.InternalServerError'
           description: There was an internal server error.
       summary: Get Image OSTree repo.
-      description: Returns the information of the OSTree Repository of a particular Image.
\ No newline at end of file
+      description: Returns the information of the OSTree Repository of a particular Image.
+  /updates:
+    post:
+      operationId: Update a device
```
biancarosa, line 159:
```suggestion
      operationId: updateDevice
```

Just to keep the operationId's simple, they're used on the URLs

biancarosa (commented):
Small coment, other than that we're good

biancarosa (approved):

PR #194, opened 2021-08-12
added constant for reserved usernames to send to frontend

resoluteCoder:
Created constants file to hold reserved usernames on a RHEL system and sending those to the frontend. Changed from previous way due to endpoint not running in a container environment.


biancarosa (approved):
LGTM

PR #198, opened 2021-08-16
Move playbook dispatcher code to clients

biancarosa:
This one breaks tests, because of how we're injecting dependencies so I wanna quickly go through the tests at the end and forget them for now.


maxamillion (approved):

maxamillion:
LGTM :+1:

PR #201, opened 2021-08-18
Create and move all things services and routes

biancarosa:


acosferreira (commented):
What do you think to aggregate the services by responsibilities?
I.E: 
"Files"-> Uploader, Downloader, Tar, Untar, etc
"Display" -> Filter, pagination, etc
"Template" -> playbook_template, etc

What do you think to move the StatusOk and the success responses to a handler instead let on common?

pkg/clients/playbookdispatcher/client.go
```diff
@@ -54,12 +54,12 @@ func (c *Client) ExecuteDispatcher(payload DispatcherPayload) ([]PlaybookDispatc
 
 	req.Header.Add("Content-Type", "application/json")
 
-	headers := common.GetOutgoingHeaders(c.ctx)
-	req.Header.Add("Authorization", fmt.Sprintf("PSK %s", c.psk))
+	headers := clients.GetOutgoingHeaders(c.ctx)
```
acosferreira, line 57: in this case, Don't need pass the headers along the way because will be available in the context?

pkg/routes/images.go
```diff
@@ -13,30 +14,35 @@ import (
...
+func MakeImagesRouter(sub chi.Router) {
+	sub.With(validateGetAllSearchParams).With(services.Paginate).Get("/", GetAll)
 	sub.Get("/reserved-usernames", GetReservedUsernames)
-	sub.Post("/", Create)
+	sub.Post("/", CreateImage)
 	sub.Route("/{imageId}", func(r chi.Router) {
 		r.Use(ImageCtx)
-		r.Get("/", GetByID)
+		r.Get("/", GetImageByID)
 		r.Get("/status", GetStatusByID)
```
acosferreira, line 46: Should we change this to GetImageStatusByID?

pkg/routes/images.go
```diff
@@ -692,12 +689,12 @@ func downloadISO(isoName string, url string) error {
...
 	cfg := config.Get()
-	var uploader files.Uploader
-	uploader = &files.FileUploader{
+	var uploader services.Uploader
+	uploader = &services.FileUploader{
 		BaseDir: "./",
 	}
 	if cfg.BucketName != "" {
-		uploader = files.NewS3Uploader()
+		uploader = services.NewS3Uploader()
```
acosferreira, line 697: Shouldn't be only NewS3Uploader?

pkg/clients/playbookdispatcher/client.go
biancarosa: In the way that I refactored clients, they context is passed during client initilization. We have ctx for each client and InitClient method that fills the context with the request context while we still have it. So we don't need to have the headers on method signature as pass as parameters because we will always have the original request context available in all moments since it is a variable in the client object itself.

pkg/routes/images.go
biancarosa: Not in this case because it has been moved to services package.

biancarosa:
@acosferreira I've adressing the renaming comment.

I also agree about splitting responsabilities, can I do this in another PR though? This is move of a "moving things around" kind of PR, I'd like to have a separate one for more, ahem, creative work, otherwise things will get messy to review.

acosferreira:
> @acosferreira I've adressing the renaming comment.
> 
> I also agree about splitting responsabilities, can I do this in another PR though? This is move of a "moving things around" kind of PR, I'd like to have a separate one for more, ahem, creative work, otherwise things will get messy to review.

For sure 👍

acosferreira (approved):
LGTM

PR #204, opened 2021-08-20
add endpoint to update image and create the parent information; add u…

acosferreira:
…pdate tag on image related

### DIFF
--- a/pkg/images/images.go
+++ b/pkg/images/images.go
@@ -284,29 +284,8 @@ func postProcessImage(id uint, headers map[string]string) {
 	}
 }
 func CreateImageUpdate(w http.ResponseWriter, r *http.Request) {
-	Create(w, r)
-}
-
-// Create creates an image on hosted image builder.
-// It always creates a commit on Image Builder.
-// We're creating a update on the background to transfer the commit to our repo.
-func Create(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
-	var image *models.Image
-	if err := json.NewDecoder(r.Body).Decode(&image); err != nil {
-		log.Error(err)
-		err := errors.NewInternalServerError()
-		w.WriteHeader(err.Status)
-		json.NewEncoder(w).Encode(&err)
-		return
-	}
-	if err := image.ValidateRequest(); err != nil {
-		log.Info(err)
-		err := errors.NewBadRequest(err.Error())
-		w.WriteHeader(err.Status)
-		json.NewEncoder(w).Encode(&err)
-		return
-	}
+	image := initImageCreateRequest(w, r)
 	account, err := common.GetAccount(r)
 	if err != nil {
 		log.Info(err)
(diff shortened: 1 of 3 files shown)


pkg/images/images.go
```diff
@@ -482,9 +498,18 @@ func GetStatusByID(w http.ResponseWriter, r *http.Request) {
 
 // GetByID obtains a image from the database for an account
 func GetByID(w http.ResponseWriter, r *http.Request) {
-	if image := getImage(w, r); image != nil {
-		json.NewEncoder(w).Encode(image)
+	image := getImage(w, r)
+	if image != nil {
+		updatesAvailable := getUpdateAvailable(int(image.ID))
+		json.NewEncoder(w).Encode(struct {
```
biancarosa, line 504: I don't think we should return this here, only at the devices endpoint.

biancarosa: (which is a future task, and not on the scope of this one as far as I could understand)

pkg/images/images.go
```diff
@@ -43,6 +43,7 @@ func MakeRouter(sub chi.Router) {
 		r.Post("/installer", CreateInstallerForImage)
 		r.Post("/repo", CreateRepoForImage)
 		r.Post("/kickstart", CreateKickStartForImage)
+		r.Post("/update", CreateImageUpdate)
```
biancarosa, line 46: It might be nice to sync with @Saurabh-Thakre and the work he was doing on this endpoint already.

pkg/models/images.go
```diff
@@ -10,17 +10,18 @@ import (
...
+	Distribution  string     `json:"Distribution"`
+	Description   string     `json:"Description"`
+	Status        string     `json:"Status"`
+	Version       int        `json:"Version" gorm:"default:1"`
+	ImageType     string     `json:"ImageType"`
+	CommitID      uint       `json:"CommitID"`
+	Commit        *Commit    `json:"Commit"`
+	InstallerID   *uint      `json:"InstallerID"`
+	Installer     *Installer `json:"Installer"`
+	ParentImageId int        `json:"ParentImage"`
```
biancarosa, line 24: There should be some sort of Gorm annotation for us to make sure this if a foreign key pointing to this table so we can assure that the data is consistent on the database.

I'd particularly call it ParentID, too, assuming that it's a FK to the same table it'll always point to a Image.

pkg/images/images.go
```diff
@@ -312,6 +316,17 @@ func Create(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	headers := common.GetOutgoingHeaders(r)
+
+	ctx := r.Context()
+	previous_image, ok := ctx.Value(imageKey).(*models.Image)
+	if !ok {
```
biancarosa, line 322: I'd rather add the code that creates a parent image just on the specific endpoint and reuse the function that are common in both, instead of using a one function that does it all in both. Don't know if that was clear.

PR #207, opened 2021-08-23
fixed spec-file syntax 

tomershinhar:
when creating an API client i ran into some syntax issues and i fixed them in order to generate the client


biancarosa (approved):
Looks really good! This is a generated file though, I don't know the fields are wrong. I'm just gonna checkout your PR and check something real quick to see if there's anything wrong with the generator.

biancarosa:
I'm merging this, but I'll take a look at how we're generating this because I just saw something really wrong :) Tks for the fix tho!

PR #208, opened 2021-08-23
add image set and endpoint to update an existent image0

acosferreira:


biancarosa (approved):
LGTM!

pkg/models/images.go
```diff
@@ -21,6 +21,7 @@ type Image struct {
 	Commit       *Commit    `json:"Commit"`
 	InstallerID  *uint      `json:"InstallerID"`
 	Installer    *Installer `json:"Installer"`
+	ParentId     int        `gorm:"foreignKey:Image"`
```
biancarosa, line 24:
```suggestion
	ParentId     *uint        `gorm:"foreignKey:Image"`
```

I think we need to make this nullable and make sure it's unsigned

biancarosa (changes requested):
I think.... we need to make this nullable

biancarosa (approved):
LGTM!

PR #211, opened 2021-08-24
Add output types as a list

biancarosa:
Related to https://github.com/RedHatInsights/edge-frontend/pull/140 and this one has to be merged first otherwise validation will break on image creation.

When the frontend uses the OutputType to display on the table, we can remove the ImageType field.


loadtheaccumulator:
/lgtm

Question: switching to snake case in the refactor?

loadtheaccumulator (commented):
/lgtm

Question: switching to snake case in the refactor?

biancarosa:
@loadtheaccumulator maybe we should add a separate task for that, it'll break more things 😬 i have a new one on the board already

PR #212, opened 2021-08-24
add doc to new endpoint

acosferreira:


cmd/spec/openapi.yaml
```diff
@@ -181,12 +193,201 @@ components:
...
+              type: array
+            Status:
+              type: string
+          type: object
+        CommitID: {}
+        Devices:
+          items: {}
+          type: array
+        DispatchRecords:
+          items: {}
```
biancarosa, line 261: Should we also add DispatchRecords?

PR #215, opened 2021-08-26
Update to openapi spec file

tomershinhar:
a small fix for the spec file for creating an api client.
also beautify the json file


tomershinhar:
updated for the new client

biancarosa (changes requested):
@tomershinhar these files are generated, so maybe we need to figure out how to add this to the file that actually generates them. Can you apply these changes the path.yml file in this same repo?

biancarosa (approved):
LGTM

PR #217, opened 2021-08-30
Add endpoint to proxy request to get playbook URL through Edge API

biancarosa:
- Removes the proxy to the OSTree repo since we are not using it. We can retrieve the code in the history if needed in the future.


acosferreira (approved):
/lgtm

PR #218, opened 2021-08-31
New api endpoint for image-sets

Saurabh-Thakre:
Signed-off-by: Saurabh Thakre sthakre@redhat.com


pkg/routes/imagesets.go
```diff
@@ -0,0 +1,45 @@
...
+
+// GetAllImageSets image objects from the database for ParentID
+func GetAllImageSets(w http.ResponseWriter, r *http.Request) {
+	var count int64
+	var images []models.Image
+	var image models.Image
+	result := imageFilters(r, db.DB)
+	pagination := common.GetPagination(r)
+
+	countResult := imageFilters(r, db.DB.Model(&models.Image{})).Where("images.parent_id is ?", image.ParentId).Count(&count)
```
biancarosa, line 29: Did you test those filters....? It sounds like things will get tricker with filters and image sets. If it doesn't work, I'd rather leave them out of the equation and come up with them in a second pull-request (after double-checking if we want the same filtering capabilities)

pkg/services/imagesets.go
```diff
@@ -0,0 +1,31 @@
...
+
+// NewImageSetsService gives a instance of the main implementation of a ImageSetsServiceInterface
+func NewImageSetsService(ctx context.Context) ImageSetsServiceInterface {
+	return &ImageSetsService{imageBuilder: imagebuilder.InitClient(ctx)}
+}
+
+// ImageSetsService is the main implementation of a ImageSetsServiceInterface
+type ImageSetsService struct {
+	ctx          context.Context
+	imageBuilder imagebuilder.ClientInterface
```
biancarosa, line 24: I don't think we need imageBuilder as a dependency in this service.

biancarosa (changes requested):
Couple of comments there, and we can add tests, what do you say?

pkg/routes/imagesets.go
acosferreira, line 29: I think we can use the relationship between images and imageSet.

pkg/services/imagesets.go
acosferreira: Agreed with @biancarosa 
the imagebuilder will continue be handled on image creation. The imageSet is a kind of "images' aggregation"

pkg/routes/imagesets.go
Saurabh-Thakre: I've tested those and it gives exact results for images with NULL ParentID and count of that. But I'd be happy to clean those and will put it another PR. It could get tricky at some point.

Saurabh-Thakre: yeah, do we already have the relationship between those in place?

pkg/routes/imagesets.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// MakeImageSetsRouter adds support for operations on image-sets
+func MakeImageSetsRouter(sub chi.Router) {
+	sub.With(validateGetAllImagesSearchParams).With(common.Paginate).Get("/", GetAllImageSets)
+	sub.Get("/reserved-usernames", GetReservedUsernames)
+}
+
+// GetAllImageSets image objects from the database for ParentID
+func GetAllImageSets(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
Saurabh-Thakre: I liked the thought

nasa-nasagif-nasahistory-eft1-238gVegOCUAYlR7jOR

pkg/routes/imagesets.go
Saurabh-Thakre: I think we will once your PR gets merged

Saurabh-Thakre:
> Couple of comments there, and we can add tests, what do you say?

Yeah, sounds good.

biancarosa (approved):
This is amazing! GJ! :)

PR #219, opened 2021-08-31
Improve upload speed

AaronH88:
Created a pool of workers to upload files and improve the speed of uploading files. Jira ticket THEEDGE-973

Everything works locally, will need to calibrate the number of workers in stage.

### DIFF
--- a/pkg/services/files/uploader.go
+++ b/pkg/services/files/uploader.go
@@ -126,6 +126,7 @@ func (u *S3Uploader) UploadRepo(src string, account string) (string, error) {
 	}
 
 	filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
+		workWg.Add(1)
 		if err != nil {
 			log.Warnf("incoming error!: %#v", err)
 		}
@@ -157,7 +158,6 @@ func (u *S3Uploader) UploadFile(fname string, uploadPath string) (string, error)
 	if err != nil {
 		return "", fmt.Errorf("failed to open file %q, %v", fname, err)
 	}
-	defer f.Close()
 	// Upload the file to S3.
 	result, err := u.Client.PutObject(&s3.PutObjectInput{
 		Bucket: aws.String(u.Bucket),
@@ -170,6 +170,7 @@ func (u *S3Uploader) UploadFile(fname string, uploadPath string) (string, error)
 	if err != nil {
 		return "", err
 	}
+	f.Close()
 	region := *u.Client.Config.Region
 	s3URL := fmt.Sprintf("https://%s.s3.%s.amazonaws.com/%s", u.Bucket, region, uploadPath)
 	return s3URL, nil
--- a/pkg/services/files/uploader.go
+++ b/pkg/services/files/uploader.go
@@ -131,10 +131,10 @@ func (u *S3Uploader) UploadRepo(src string, account string) (string, error) {
 	log.Debugf("S3Uploader::UploadRepo::account: %#v", account)
 
 	createWorkers()
-	var workWg sync.WaitGroup
+	var repoWaitGroup sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


pkg/services/files/uploader.go
```diff
@@ -89,6 +90,24 @@ func newS3Uploader() *S3Uploader {
 	}
 }
 
+type filePathVar struct {
```
biancarosa, line 93: Thinking about code readability, can we rename this? I thought about repoSettings, localRepo, repoToUpload, a bunch of things that I'm not particularly happy either but at least it's clearer 😬  @maxamillion @acosferreira @loadtheaccumulator if you have better ideas that's the time

pkg/services/files/uploader.go
```diff
@@ -89,6 +90,24 @@ func newS3Uploader() *S3Uploader {
...
+	uploadPath string
+}
+
+func worker(wg *sync.WaitGroup, queue <-chan filePathVar, u *S3Uploader) {
+	defer wg.Done()
+
+	for p := range queue {
+		fname := p.fileName
+		uploadDest := p.uploadPath
+		_, err := u.UploadFile(fname, uploadDest)
```
biancarosa, line 104:
```suggestion
		_, err := u.UploadFile(p.filename, p.uploadDest)
```

Or just pass p right away, but that's a major refactor

pkg/services/files/uploader.go
```diff
@@ -97,10 +116,17 @@ func (u *S3Uploader) UploadRepo(src string, account string) (string, error) {
...
 	log.Debugf("S3Uploader::UploadRepo::account: %#v", account)
 
-	// FIXME: might experiment with doing this concurrently but I've read that
-	//		  that can get you rate limited by S3 pretty quickly so we'll mess
-	//		  with that later.
+	//queue of jobs
+	q := make(chan filePathVar)
+
+	numberOfWorkers := 100
+	var workWg sync.WaitGroup
```
biancarosa, line 123: workWg could also be something easier to read 😬 is this wait group used for files?

biancarosa (changes requested):
@AaronH88 awesome work, minor comments, other than that we're good!

pkg/services/files/uploader.go
AaronH88: How about "uploadDetails" ?

pkg/services/files/uploader.go
AaronH88: Yeah, and its a unique group for the provided repo,
perhaps "repoWaitGroup" ?

pkg/services/files/uploader.go
```diff
@@ -89,6 +95,33 @@ func newS3Uploader() *S3Uploader {
...
 }
 
+type uploadDetails struct {
+	fileName   string
+	uploadPath string
+	uploader   *S3Uploader
+	wg         *sync.WaitGroup
+}
+
+func worker() {
```
AaronH88, line 105: Rename to uploadWorkers() ??

biancarosa (approved):
Most of my comments were regarding readability and I feel like they've been addressed, I'm good with this and I think it'll be an amazing improvement!

pkg/services/files/uploader.go
biancarosa: nah, in this case maybe we don't need to worry too much, it's pretty clear to me

pkg/services/files/uploader.go
AaronH88: Changed to repoWaitGroup

pkg/services/files/uploader.go
AaronH88: I agree that P is a major refactor, however I changed this fucntion call to use the "p." notation you suggested

PR #223, opened 2021-09-01
Feat/image set model

acosferreira:


pkg/models/images.go
```diff
@@ -8,6 +8,14 @@ import (
 	"gorm.io/gorm"
 )
 
+type ImageSet struct {
+	gorm.Model
+	Name    string  `json:"Name"`
+	Version int     `json:"Version" gorm:"default:1"`
+	Account string  `json:"Account"`
+	Images  []Image `json:"Images"`
```
biancarosa, line 16: I don't know GORM well enough, but I am not sure if we want this field in this model unless we can ignore it on the table since the relationship is owned by the Image model.

pkg/routes/images.go
```diff
@@ -129,6 +130,15 @@ func CreateImage(w http.ResponseWriter, r *http.Request) {
 		json.NewEncoder(w).Encode(&err)
 		return
 	}
+	var imageSet models.ImageSet
```
biancarosa, line 133: Can we add this to the image service method?

pkg/routes/images.go
```diff
@@ -172,7 +182,18 @@ func CreateImageUpdate(w http.ResponseWriter, r *http.Request) {
 		json.NewEncoder(w).Encode(&err)
 	}
 	if previous_image != nil {
+		var currentImageSet models.ImageSet
```
biancarosa, line 185: How do you feel about creating a UpdateImage method on ImageService  that does this?

pkg/services/images.go
```diff
@@ -27,6 +28,7 @@ var WaitGroup sync.WaitGroup
 // the business logic of creating RHEL For Edge Images
 type ImageServiceInterface interface {
 	CreateImage(image *models.Image, account string) error
+	UpdateImage(image *models.Image, account string, previous_image *models.Image) error
```
biancarosa, line 31:
```suggestion
	UpdateImage(image *models.Image, account string, previousImage *models.Image) error
```

just being strangely attached to Go's camelCase suggestion here 😺

biancarosa (commented):
This is looking sooooooooo good 👍🏻  One last comment and we're good to go!

Saurabh-Thakre:
> This is looking sooooooooo good 👍🏻 One last comment and we're good to go!

Really looks good, great work Ana

biancarosa (approved):

PR #224, opened 2021-09-01
Refactor devices routes to be on a separate file

biancarosa:
Yesterday while coding the update playbook method I noticed some improvements and talked to @acosferreira about it today.

I want to do some unit tests for the code pushed yesterday, so this should make things easier to test too.

- Creates devices route file.
- Separates device and update context
- Makes getUpdate work again
- Delete code that is not used or could be reused

### DIFF
--- a/pkg/routes/devices.go
+++ b/pkg/routes/devices.go
@@ -1,0 +1,107 @@
+package routes
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/redhatinsights/edge-api/pkg/clients/inventory"
+	"github.com/redhatinsights/edge-api/pkg/db"
+	"github.com/redhatinsights/edge-api/pkg/dependencies"
+	"github.com/redhatinsights/edge-api/pkg/models"
+	"github.com/redhatinsights/edge-api/pkg/routes/common"
+	log "github.com/sirupsen/logrus"
+)
+
+// MakeDevicesRouter adds support for operations on update
+func MakeDevicesRouter(sub chi.Router) {
+	sub.Route("/device/", func(r chi.Router) {
+		r.Use(DeviceCtx)
+		sub.Get("/{DeviceUUID}", GetDeviceStatus)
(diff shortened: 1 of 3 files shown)


pkg/routes/devices.go
```diff
@@ -0,0 +1,107 @@
...
+	// var devices []models.Device
+	var results []models.Device
+	//var results []models.UpdateTransaction
+	account, err := common.GetAccount(r)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	uuid := chi.URLParam(r, "DeviceUUID")
+	result := db.DB.
```
acosferreira, line 64: Is not possible use Gorm to do this query/relation?

pkg/routes/devices.go
```diff
@@ -0,0 +1,107 @@
...
+	DeviceUUID string
+	Tag        string
+}
+
+//  DeviceCtx is a handler for Device requests
+func DeviceCtx(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var uCtx DeviceContext
+		uCtx.DeviceUUID = chi.URLParam(r, "DeviceUUID")
+		uCtx.Tag = chi.URLParam(r, "Tag")
```
acosferreira, line 42: we're handling the tag, but we're not using, what do you think to let is commented until we start to work with the tags?

pkg/routes/devices.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+}
+
+// GetDeviceStatus returns the device with the given UUID that is associate to the account.
+// This is being used for the inventory table to determine whether the current device image
+// is the latest or older version.
+func GetDeviceStatus(w http.ResponseWriter, r *http.Request) {
+	// var devices []models.Device
+	var results []models.Device
+	//var results []models.UpdateTransaction
```
acosferreira, line 57: what do you think about remove this commented var to keep the code clean?

pkg/routes/devices.go
biancarosa: I'm not sure but I'd rather not mess with this one since I just moved the query to a different function

acosferreira (approved):
LGTM

PR #226, opened 2021-09-02
Add tests & docs for new playbook endpoint

biancarosa:


cmd/spec/openapi.yaml
```diff
@@ -620,6 +618,35 @@ paths:
...
       summary: Executes a device update.
       description: Executes a device update.
+  /updates/{updateID}/update-playbook.yml:
+    get:
+      operationId: GetUpdatePlaybook
+      parameters:
+        - name: updateID
+          in: path
+          required: true
+          description: An unique ID to identify the update that you want to retrieve your playbook for
```
AaronH88, line 628: "An unique ID to identify the update playbook" ?

cmd/spec/path.yaml
```diff
@@ -220,6 +220,35 @@ paths:
...
       summary: Executes a device update.
       description: Executes a device update.
+  /updates/{updateID}/update-playbook.yml:
+    get:
+      operationId: GetUpdatePlaybook
+      parameters:
+        - name: updateID
+          in: path
+          required: true
+          description: An unique ID to identify the update that you want to retrieve your playbook for
```
AaronH88, line 230: Same as other file: "An unique ID to identify the update playbook" ?

AaronH88 (commented):
Small re-wording suggestion.

Other than that, looks great! Ok to merge with or without the suggestion.

acosferreira (approved):
LGTM

PR #227, opened 2021-09-02
Revert "Improve upload speed"

AaronH88:
Reverts RedHatInsights/edge-api#219

Nil pointer error, unable to re-create locally, will revert and figure it out.


acosferreira (approved):
lgtm

PR #231, opened 2021-09-02
fix: concurrency stuff on upload repo

biancarosa:


maxamillion (approved):

pkg/services/files/uploader.go
```diff
@@ -101,26 +97,29 @@ type uploadDetails struct {
...
 //If the error "Too many open files" appears, lower the number of workers
-func createWorkers() {
+func createWorkers(uploadQueue chan *uploadDetails) {
 	if !workersCreated {
 		numberOfWorkers := 100
 		for i := 0; i < numberOfWorkers; i++ {
-			go worker()
+			go worker(uploadQueue)
 		}
 		workersCreated = true
```
AaronH88, line 124: Have you testing this locally?
I think this bool will prevent more than one repo to be uploaded, have you tried creating two or three upload requests?

This whole func can be removed, and creating 100 workers can fall back into the UploadRepo() func

pkg/services/files/uploader.go
```diff
@@ -19,9 +18,6 @@ import (
 //Bool used to control if uploading workers need to be created
 var workersCreated = false
```
AaronH88, line 19: I think this bool is no longer needed

pkg/services/files/uploader.go
```diff
@@ -153,12 +155,21 @@ func (u *S3Uploader) UploadRepo(src string, account string) (string, error) {
...
+		res.done = make(chan bool)
+		uploadQueue <- res
+		uploadDetailsList = append(uploadDetailsList, res)
+		count++
 		return nil
 	})
-
-	repoWaitGroup.Wait()
+	log.Infof("Files are being uploaded.... %d files to upload", len(uploadDetailsList))
+	for i, u := range uploadDetailsList {
```
AaronH88, line 166: "u" is already defined in this function, can we rename this var please? I got confused for a min wondering why we are looking for "uploader"

AaronH88: Do we want to do some sort of conditional to ensure that i is equal to count (-1) at the end of this loop?

AaronH88 (changes requested):
I think the bool "workersCreated" needs to be removed, and some small var name changes.

Apart from that, it looks good to me :smile:

pkg/services/files/uploader.go
biancarosa: I tested but just with one request. I can see how the bool will be a problem and I'm removing it :)

pkg/services/files/uploader.go
```diff
@@ -134,13 +118,20 @@ func (u *S3Uploader) UploadRepo(src string, account string) (string, error) {
 	log.Debugf("S3Uploader::UploadRepo::src: %#v", src)
 	log.Debugf("S3Uploader::UploadRepo::account: %#v", account)
 
-	createWorkers()
+	uploadQueue := make(chan *uploadDetails)
+
+	numberOfWorkers := cfg.UploadWorkers
```
AaronH88, line 123: NICE!

AaronH88:
Looks good to me, lets give it a go and see what happens

AaronH88 (approved):
Great!

biancarosa:
I'm thinking IQE is not happy with some latest changes so I'm gonna merge this and see what happens.

PR #235, opened 2021-09-03
create workers after we know the size of uploadslist

biancarosa:


AaronH88 (approved):
Lets give it a shot, ok to merge

PR #240, opened 2021-09-05
update path.yaml and image model for spec-file generation

tomershinhar:
[THEEDGE-937](https://issues.redhat.com/browse/THEEDGE-937)
most of the issues was that path.yaml was outdated.
regarding the missing elements from the spec file, this is because kin-openapi is using MarshalStrictStruct which only takes elements with a json tag while the image model was using gorm.Model for the ID which is missing the json tags


biancarosa (approved):
Awesome! Thank you ❤️

PR #241, opened 2021-09-07
add validation before to get the results of query on device

acosferreira:
Validate if has result on inventory before get the last checksum
Validate if has result on images before continue loading the associations


maxamillion (approved):

maxamillion:
LGTM :+1:

PR #243, opened 2021-09-08
Update app-common-go version

tpapaioa:
With `app-common-go` version 1.5.0, updating the OpenAPI spec gives a benign but confusing error message:

```
$ go run cmd/spec/main.go 
open : no such file or directory
```

This PR updates the dependency to version 1.5.2, which still outputs a message, but at least it's more understandable:

```
$ go run cmd/spec/main.go 
Clowder is not enabled, skipping init...
```

(The clowder config package prints this message during import if the `ACG_CONFIG` environment variable is not set, but `cmd/spec/main.go` doesn't use that package, so the message can be ignored.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/go-openapi/runtime v0.19.29
 	github.com/golang/mock v1.3.1
 	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/google/uuid v1.1.1
 	github.com/jackc/pgproto3/v2 v2.0.7 // indirect
 	github.com/lib/pq v1.3.0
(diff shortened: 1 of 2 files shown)


Saurabh-Thakre:
Looks good to me, I think we were discussing something around this but never really dig much. Thanks for this.

Saurabh-Thakre (commented):
Looks good to me, I think we were discussing something around this but never really dig much. Thanks for this.

biancarosa (approved):

biancarosa:
This is great! Thank you so much <3

biancarosa:
@tpapaioa d'you have permissions to merge?

PR #246, opened 2021-09-08
Feat/use update diff based on img builder info - part 1: refactor

biancarosa:


pkg/services/devices.go
```diff
@@ -59,45 +66,61 @@ type DeltaDiff struct {
...
+	lastDeployment := lastDevice.Ostree.RpmOstreeDeployments[lastDeploymentIdx]
+
 	var images []models.Image
 	var currentImage models.Image
-	result := db.DB.Joins("Commit").Where("OS_Tree_Commit = ?", currentCheckSum)
-	if result.Error != nil {
-		return nil, errors.NewInternalServerError()
+
+	result := db.DB.Where("OS_Tree_Commit = ?", lastDeployment.Checksum).First(&currentImage.Commit)
+	if result.Error != nil || result.RowsAffected == 0 {
```
acosferreira, line 89: result.RowsAffected is not working as expected, what do you think to change to results==nil?

pkg/services/devices.go
```diff
@@ -59,45 +66,61 @@ type DeltaDiff struct {
...
+		log.Error(result.Error)
+		return nil, new(DeviceNotFoundError)
 	}
-	if result.RowsAffected == 0 {
-		return nil, errors.NewNotFound("Record not found")
-	} else {
-		result.First(&currentImage)
+	result = db.DB.Where("commit_id = ?", currentImage.Commit.ID).First(&currentImage)
+	if result.Error != nil || result.RowsAffected == 0 {
```
acosferreira, line 94: same as above

pkg/services/devices.go
```diff
@@ -59,45 +66,61 @@ type DeltaDiff struct {
...
-			diff := GetDiffOnUpdate(currentImage, upd)
-			delta.Image = upd
-			delta.PackageDiff = diff
-			imageDiff = append(imageDiff, delta)
-		}
-
-		return imageDiff, nil
-	} else {
-		return nil, updates.Error
+	if updates.Error != nil || result.RowsAffected == 0 {
```
acosferreira, line 106: I think here you should validate the updates == nil instead result.RowsAffected because is related to the updates query, What do you think?

pkg/services/devices.go
biancarosa: I think in this case, with the unit tests, I was able to nail down the error and fix it, but let's talk later and find out what's the best approach :)

acosferreira (approved):
LGTM

PR #254, opened 2021-09-10
Feat/rebase image info

acosferreira:
Rebase with refactor
* added endpoint to get image info for a device
* added tests
* added doc


pkg/services/devices.go
```diff
@@ -153,3 +165,40 @@ func contains(s []string, searchterm string) bool {
...
+	if result.Error != nil || result == nil {
+		log.Error(result.Error)
+		return ImageInfo, new(ImageNotFoundError)
+	} else {
+		if currentImage.ParentId != nil {
+			db.DB.Where("ID = ?", currentImage.ParentId).First(&rollback)
+		}
+	}
+	//MUST FIX DeviceServiceInterface.GetUpdateAvailableForDeviceByUUID
+	// updateAvailable, err := DeviceServiceInterface.GetUpdateAvailableForDeviceByUUID(deviceUUID)
```
acosferreira, line 196: @biancarosa needing some halp with this get, for some reason I'm not able to query directly due to wrong type of var 🤔  Do you have any suggestion?

biancarosa (approved):

pkg/services/devices.go
biancarosa: Alright, yes! Here you don't wanna query the interface directly, you wanna query the service available to the route by the dependency injection through the context.

On the GetUpdateAvailableEndpoint, I did:

```

	contextServices, _ := r.Context().Value(dependencies.Key).(*dependencies.EdgeAPIServices)
	result, err := contextServices.DeviceService.GetUpdateAvailableForDeviceByUUID(dc.DeviceUUID)
```

biancarosa (approved):

PR #257, opened 2021-09-13
Feat: Get diff of packages from metadata

biancarosa:


pkg/services/devices.go
```diff
@@ -131,38 +132,45 @@ func (s *DeviceService) GetUpdateAvailableForDeviceByUUID(deviceUUID string) ([]
...
+	for _, pkg := range old {
+		oldPkgs[pkg.Name] = pkg
 	}
-	var removed []models.Package
-	for _, pkg := range initialCommit {
-		if !contains(updateString, pkg.Name) {
-			removed = append(removed, pkg)
+	for _, pkg := range new {
+		if oldPkg, ok := oldPkgs[pkg.Name]; ok {
+			v1 := version.NewVersion(oldPkg.Version)
```
acosferreira, line 157: what do you think to rename v1/v2 to previous and current?

PR #259, opened 2021-09-13
Feat: Get all device info

biancarosa:


pkg/services/devices.go
```diff
@@ -59,6 +68,34 @@ func (s *DeviceService) GetDeviceByUUID(deviceUUID string) (*models.Device, erro
...
+	if device.ID != 0 {
+		updates, err = s.updateService.GetUpdateTransactionsForDevice(device)
+		if err != nil {
+			return nil, err
+		}
+	}
+	details := &DeviceDetails{
+		Device:  device,
+		Image:   imageInfo,
+		Updates: updates,
```
acosferreira, line 94: I think we should do explicit that this is the update transaction, once inside Image we also have the updateAvailable. however I'm not sure how we can do that :/

biancarosa: Maybe call it `UpdateTransactions`? I like that name a lot better, it maintains a consistency.

biancarosa:
@acosferreira changed some names there, take a look whenever possible :)

pkg/services/devices.go
```diff
@@ -118,13 +154,14 @@ func (s *DeviceService) GetUpdateAvailableForDeviceByUUID(deviceUUID string) ([]
...
 	}
-
 	var imageDiff []ImageUpdateAvailable
 	for _, upd := range images {
 		db.DB.First(&upd.Commit, upd.CommitID)
 		db.DB.Model(&upd.Commit).Association("InstalledPackages").Find(&upd.Commit.InstalledPackages)
+		db.DB.Model(&upd.Commit).Association("Packages").Find(&upd.Commit.Packages)
 		var delta ImageUpdateAvailable
 		diff := getDiffOnUpdate(currentImage, upd)
+		upd.Commit.InstalledPackages = nil // otherwise the frontend will get the whole list of installed packages
```
acosferreira, line 164: should we ask them to remove this dependency on their end once we are returning nil?

pkg/services/devices.go
```diff
@@ -118,13 +154,14 @@ func (s *DeviceService) GetUpdateAvailableForDeviceByUUID(deviceUUID string) ([]
 	if updates.Error != nil || updates.RowsAffected == 0 {
 		return nil, new(UpdateNotFoundError)
 	}
-
 	var imageDiff []ImageUpdateAvailable
 	for _, upd := range images {
 		db.DB.First(&upd.Commit, upd.CommitID)
 		db.DB.Model(&upd.Commit).Association("InstalledPackages").Find(&upd.Commit.InstalledPackages)
```
acosferreira, line 160: Should we remove this once is not being used?

pkg/services/updates.go
```diff
@@ -197,3 +198,20 @@ func (s *UpdateService) writeTemplate(templateInfo TemplateRemoteInfo, account s
 	log.Infof("::WriteTemplate: ENDs")
 	return playbookURL, nil
 }
+
+func (s *UpdateService) GetUpdateTransactionsForDevice(device *models.Device) (*[]models.UpdateTransaction, error) {
+	var updates *[]models.UpdateTransaction
+
+	result := db.DB.
```
acosferreira, line 205: any reason to dont use gorm structure query?

pkg/services/devices.go
biancarosa: the frontend doesn't depend on the installed packages at all already, just the packages but not for that endpoint

pkg/services/devices.go
biancarosa: it actually is being used on the diff calculation, I'm removing after the diff calculation is done since after it's calculated, we don't need the list anymore.

pkg/services/updates.go
biancarosa: for this one I grabbed Boaz's query that was created already, but I'm gonna take a look on how to refactor since some things changed on the data we needed.

biancarosa: Okay, we are using GORM structured query as best as I know for a Many2Many join. I don't know how to do differently and spent a lot of time researching this a couple of days ago without being able to find it, so I'm happy to accept suggestions. Added unit tests for this bit of code meanwhile.

acosferreira (approved):
LGTM

PR #260, opened 2021-09-14
fixed devices endpoints on path.ymal

tomershinhar:


tomershinhar:
changed the return type for /updates/device/{DeviceUUID}/image to be a dict with image field in it 
and /updates/device/{DeviceUUID}/updates to be a list containing a dict with image field in it to mach the  API response

biancarosa (approved):

PR #263, opened 2021-09-15
init reboot device

acosferreira:


pkg/services/template_playbook_dispatcher_reboot_device.yml
```diff
@@ -0,0 +1,5 @@
+- name: Reboot the device after execute the update transaction
```
biancarosa, line 1: Thoughts and feelings about moving this, alongside with the other template, to a templates specific folder?

pkg/routes/updates.go
```diff
@@ -311,6 +312,12 @@ func AddUpdate(w http.ResponseWriter, r *http.Request) {
 	go repoService.CreateUpdate(update)
 	w.WriteHeader(http.StatusOK)
 	json.NewEncoder(w).Encode(update)
+
+	timer := time.AfterFunc(time.Minute*10, func() {
```
biancarosa, line 316: Can we have this 10 set up as a config env var in case we wanna tweak? We have 10 as default for local, and on the Clowder YAMLs, and if we need to tweak we only change on app-interface

pkg/services/updates.go
```diff
@@ -197,3 +199,66 @@ func (s *UpdateService) writeTemplate(templateInfo TemplateRemoteInfo, account s
...
+		log.Errorf("Error on playbook-dispatcher-executuin: %#v ", err)
+
+	}
+	for _, excPlaybook := range exc {
+		if excPlaybook.StatusCode == http.StatusCreated {
+			update.Devices[len(update.Devices)-1].Connected = true
+			dispatchRecord := &models.DispatchRecord{
+				Device:               &update.Devices[len(update.Devices)-1],
+				PlaybookURL:          playbookURL,
+				Status:               models.DispatchRecordStatusCreated,
```
biancarosa, line 247: For those that we created the dispatch record, is it worthy to check in a call to inventory 5-10 minutes later to see if the update was successful based on the OSTree Commit running on the device?

acosferreira: you mean not record this call to playbook dispatcher and rely on OSTree Commit? And about check the execution, should we not worry about it?

pkg/routes/updates.go
acosferreira: I think this will be a temporary solution, once we could get the response on playbook dispatcher execution as success, this should be move to another way (queue, event, etc)

biancarosa: As a constant at least? Looks like a magic number

pkg/services/updates.go
biancarosa: yes, at least for now yes.

i think we have to worry about setting correct status as success when the execution finishes as soon as possible, but I might be wrong

biancarosa (approved):
LGTM

biancarosa:
Merging this since IQE Plugin current has a bug that's on the way of getting this check to pass.

PR #265, opened 2021-09-20
Added validation to image is already in use

adeliaferreira:
### what's is inside?
- Added validation to check if name is already in use.

### How is it build?
- Added function to retrieve image register by name.
- Check if result data from database.
- Added valiadation on ValidateRequest function.
- Create new Error string.


pkg/models/images.go
```diff
@@ -125,3 +132,13 @@ func (i *Image) HasOutputType(imageType string) bool {
 	}
 	return false
 }
+
+//ImageNameExist checks if name to image is already in use
+func ImageNameExist(imageName string) bool {
```
biancarosa, line 137: I'd make this private for this package, by using lowercase instead of the uppercase initial letter. This is how Go deal with private methods :) 

Also, I like method names more as verbs / actions: checkIfImageNameExists, for example. Thoughts on it?

pkg/models/images.go
```diff
@@ -125,3 +132,13 @@ func (i *Image) HasOutputType(imageType string) bool {
...
+	var imageFindByName []Image
```
biancarosa, line 138: Why a list of images instead of just one image?

pkg/models/images.go
```diff
@@ -125,3 +132,13 @@ func (i *Image) HasOutputType(imageType string) bool {
...
+	result := db.DB.Where("Name = ?", imageName).First(&imageFindByName)
+	if result.Error != nil {
+		return true
```
biancarosa, line 141: if it errors out, won't it mean that it doesn't exist instead?

pkg/models/images.go
```diff
@@ -125,3 +132,13 @@ func (i *Image) HasOutputType(imageType string) bool {
...
+	}
+	return imageFindByName != nil
```
biancarosa, line 143: I think .First will always return error, but I don't see a problem in keeping this here just for sanity check :)

biancarosa (approved):
LGTM!

biancarosa (approved):

PR #266, opened 2021-09-20
Add update status in the beggining of an update

biancarosa:


acosferreira (approved):
LGTM

biancarosa:
Merging this since IQE Plugin current has a bug that's on the way of getting this check to pass.

PR #267, opened 2021-09-21
Use the edge iqe plugin image for smoke tests, instead of the full iqe test suite image

tpapaioa:
The GitHub PR checks were previously using the container images that contain the full test suite from all IQE plugins. To avoid potential problems caused by changes to other plugins, this PR updates `clowdapp.yaml` to use the images containing only the edge plugin.


acosferreira (approved):
LGTM

PR #269, opened 2021-09-21
Feat/use image set

acosferreira:


pkg/services/devices.go
```diff
@@ -226,8 +226,8 @@ func (s *DeviceService) GetDeviceImageInfo(deviceUUID string) (*ImageInfo, error
 		log.Error(result.Error)
 		return nil, new(ImageNotFoundError)
 	} else {
-		if currentImage.ParentId != nil {
-			db.DB.Where("ID = ?", currentImage.ParentId).First(rollback)
+		if currentImage.ImageSetID != nil {
+			db.DB.Where("Image_Set_Id = ?", currentImage.ImageSetID).First(rollback)
```
biancarosa, line 230: Do we really want first here? I think we want the immediate previous version 🤔

biancarosa: If I have image 1, 2, 3, and device is on image 3, the output for this should be 2? 🤔

pkg/services/devices.go
```diff
@@ -147,7 +147,7 @@ func (s *DeviceService) GetUpdateAvailableForDeviceByUUID(deviceUUID string) ([]
 		return nil, new(DeviceNotFoundError)
 	}
 
-	updates := db.DB.Where("Parent_Id = ? and Images.Status = ?", currentImage.ID, models.ImageStatusSuccess).Joins("Commit").Find(&images)
+	updates := db.DB.Where("Image_set_id = ? and Images.Status = ? and Images.Id != ?", currentImage.ImageSetID, models.ImageStatusSuccess, currentImage.ID).Joins("Commit").Find(&images)
```
biancarosa, line 150: Instead of Images.ID !=, don't we want only the ones with IDs that are > than Image.ID?

If I have Images 1, 2, 3, device is on image 2, we only wanna show 3 in this scenario, right?

biancarosa (approved):

PR #272, opened 2021-09-22
Update openapi spec cmd to pretty-print json file

tpapaioa:
This PR updates the OpenAPI spec generation tool `cmd/spec/main.go` so that the resulting `openapi.json` output file is in pretty-print format. This will make it easier to verify the differences in git commits and in the QE automated test suite that uses this file to generate its REST API client.


biancarosa (approved):
This is awesome! Thanks!

PR #274, opened 2021-09-28
Feat/return image sets

acosferreira:
adjust context on image sets
created service to get info by id
re-generate mock services


pkg/routes/imagesets.go
```diff
@@ -13,9 +15,35 @@ import (
...
+		if imageSetID := chi.URLParam(r, "imageSetId"); imageSetID != "" {
+			_, err := strconv.Atoi(imageSetID)
+			if err != nil {
+				err := errors.NewBadRequest(err.Error())
+				w.WriteHeader(err.Status)
+				json.NewEncoder(w).Encode(&err)
+				return
+			}
+
+			ctx := context.WithValue(r.Context(), imageSetKey, imageSetID)
```
biancarosa, line 43: We could, instead of saving the image set ID, do the DB query and save the actual ImageSet :) We've been doing that to avoid doing the same query over and over in all of the routes.

pkg/services/imagesets.go
```diff
@@ -72,3 +74,15 @@ func (s *ImageSetsService) ListAllImageSets(w http.ResponseWriter, r *http.Reque
 	json.NewEncoder(w).Encode(map[string]interface{}{"data": &images, "count": count})
 	return nil
 }
+
+func (s *ImageSetsService) GetImageSetsByID(imageSetID int) (*models.ImageSet, error) {
+	var imageSet models.ImageSet
+	result := db.DB.Where("Image_sets.id = ?", imageSetID).Find(&imageSet)
+	db.DB.Where("image_set_id = ?", imageSetID).Find(&imageSet.Images)
+	fmt.Printf("::image_set_id:: %v", result.Error)
```
biancarosa, line 82: remove that line ❌

pkg/services/imagesets_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+func TestGetImageSetByID(t *testing.T) {
+	imageSetService := ImageSetsService{
+		ctx: context.Background(),
+	}
+	imageSet, err := imageSetService.GetImageSetsByID(1)
+	if err != nil {
+		t.Errorf("Expected nil image set available, got %#v", err)
+	}
+	fmt.Print(imageSet)
```
biancarosa, line 17: ❌

pkg/routes/imagesets.go
```diff
@@ -13,9 +16,43 @@ import (
...
+
+		if imageSetID := chi.URLParam(r, "imageSetId"); imageSetID != "" {
+			_, err := strconv.Atoi(imageSetID)
+			if err != nil {
+				err := errors.NewBadRequest(err.Error())
+				w.WriteHeader(err.Status)
+				json.NewEncoder(w).Encode(&err)
+				return
+			}
+			result := db.DB.Where("Image_sets.id = ?", imageSetID).Find(&imageSet)
```
biancarosa, line 44: I think this query gets the ImageSet images, but not the image set itself 🤔  Am I right?

acosferreira: no, gets the image sets and after I'm loading the images related to this set.

biancarosa (approved):

pkg/routes/imagesets.go
maxamillion: This should also include the account ID too, yes?

maxamillion (approved):

PR #277, opened 2021-09-29
add save repo; remove code, we should always create a new one

acosferreira:


pkg/routes/updates.go
```diff
@@ -262,7 +262,7 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) (*models.UpdateTrans
 		}
 	}
 	update.OldCommits = oldCommits
-
+	db.DB.Save(&update)
```
biancarosa, line 265: I think DB.Save returns and error, and we should check it

pkg/services/repobuilder.go
```diff
@@ -128,23 +128,23 @@ func (rb *RepoBuilder) BuildUpdateRepo(ut *models.UpdateTransaction) (*models.Up
...
-				log.Errorf("updateFromHTTP::GetRepoByCommitID::repo: %#v, %#v", repo, err)
-			} else {
-				log.Infof("Old Repo not found in database for CommitID, creating new one: %d", update.CommitID)
-				update.Repo = &models.Repo{
-					Commit: update.Commit,
-				}
-			}
-		}
-	}
+	// if update.Repo == nil {
```
biancarosa, line 131: Just delete it? We can recover over git history if needed

biancarosa: Or better: if update.Repo is nil here, we might want to return an error in the whole method, just to avoid a nil pointer that will be hard to track

biancarosa (changes requested):

pkg/services/repobuilder.go
```diff
@@ -128,23 +128,23 @@ func (rb *RepoBuilder) BuildUpdateRepo(ut *models.UpdateTransaction) (*models.Up
...
+	// 			log.Infof("Old Repo not found in database for CommitID, creating new one: %d", update.CommitID)
+	// 			update.Repo = &models.Repo{
+	// 				Commit: update.Commit,
+	// 			}
+	// 		}
+	// 	}
+	// }
 	update.Repo.URL = repoURL
 	update.Repo.Status = models.RepoStatusSuccess
 	db.DB.Save(&update)
```
biancarosa, line 150: since we're here, let's add a error check for this babe?

biancarosa (approved):
Awesome! Good finding here.

biancarosa:
@acosferreira the error on the check seems to be related to not finding imgs on quay.io, we need to fix this before the merge.

PR #278, opened 2021-09-30
pre-prod playbook verification workaround

maxamillion:
We are not currently signing Ansible playbooks for use in playbook
dispatcher for local development nor in stage (yet) so we need to
configure the rhc-worker-playbook plugin accordingly at device
onboarding time.


acosferreira (approved):
LGTM

PR #280, opened 2021-09-30
add validation on rpm_ostree to check current booted image on device

acosferreira:


biancarosa (approved):

pkg/services/devices.go
```diff
@@ -211,14 +211,19 @@ func (s *DeviceService) GetDeviceImageInfo(deviceUUID string) (*ImageInfo, error
...
 	if err != nil || device.Total != 1 {
 		return nil, new(DeviceNotFoundError)
 	}
 
 	lastDevice := device.Result[len(device.Result)-1]
-	lastDeployment := lastDevice.Ostree.RpmOstreeDeployments[len(lastDevice.Ostree.RpmOstreeDeployments)-1]
+
+	for _, rpm_ostree := range lastDevice.Ostree.RpmOstreeDeployments {
+		if rpm_ostree.Booted {
+			lastDeployment = rpm_ostree
```
biancarosa, line 224: We might wanna add a return here to resume the loop otherwise we'll get the first

biancarosa (changes requested):

pkg/services/devices.go
biancarosa: In fact, I'm not sure, because that will depend on the order. 

Nice thing for a test case.

Scenario:

Deployment 1 - booted true
Deployment 2 - booted true
Deployment 3 - booted false

We want it to return the 2nd

acosferreira: I added the break, however I think we can have only one as true

biancarosa (approved):

PR #286, opened 2021-10-04
if no updates found is not an erro, is update unavailable

acosferreira:


pkg/services/devices.go
```diff
@@ -151,9 +151,12 @@ func (s *DeviceService) GetUpdateAvailableForDeviceByUUID(deviceUUID string) ([]
 	}
 
 	updates := db.DB.Where("Image_set_id = ? and Images.Status = ? and Images.Id > ?", currentImage.ImageSetID, models.ImageStatusSuccess, currentImage.ID).Joins("Commit").Find(&images)
-	if updates.Error != nil || updates.RowsAffected == 0 {
+	if updates.Error != nil {
 		return nil, new(UpdateNotFoundError)
 	}
+	if updates.RowsAffected == 0 {
+		return nil, nil
```
biancarosa, line 158: do we want nil or a empty list here?

biancarosa (approved):

PR #287, opened 2021-10-04
Feat/get image by ostree

acosferreira:


cmd/spec/openapi.json
```diff
@@ -2772,6 +2772,55 @@
         "summary": "Composes an Update for a image"
       }
     },
+    "/images/{imageOstree}/info": {
```
biancarosa, line 2775: can we call it... `ostreeCommitHash`? Or something like that?

cmd/spec/openapi.json
```diff
@@ -2772,6 +2772,55 @@
...
+      "get": {
+        "operationId": "GetImageByOstree",
+        "parameters": [
+          {
+            "description": "OstreeCommitId",
+            "in": "path",
+            "name": "ostree",
```
biancarosa, line 2782: I feel like the name should match what's on the URL path

biancarosa (approved):
lgtm

PR #288, opened 2021-10-04
Settings endpoint

resoluteCoder:
Added route to send DDF data to application settings. Partially completes this ticket [THEEDGE-962](https://issues.redhat.com/browse/THEEDGE-962)


maxamillion:
@resoluteCoder the endpoint `settings` seems a bit generic and lacking context .... what types of settings will be stored here? would it make sense to namespace the endpoint (`api/edge/v1/namespace/settings/`)? or otherwise namespace the endpoint itself with a hyphenated name (`api/edge/v1/namespace-settings/`)?

resoluteCoder:
@maxamillion I agree but I'm not sure if we are able to do something like that based on the prerequisites listed [here](https://github.com/RedHatInsights/settings-frontend#prerequisites). It looks like the path has to be `/api/{appName}/{version}/settings`

If there is a way or if I'm missing something please let me know :smile:

maxamillion:
@resoluteCoder omg, nevermind me .... I completely mixed up the context of what this PR was for. Carry on! :+1:

PR #290, opened 2021-10-04
Create check image name service i

adeliaferreira:
### What is inside?
A new method in Image service to check if an image name is already in use for the account

### How is build?
Modify Service interface and implement method to check database based in a query per image name and account


biancarosa:
@adeliaferreira I think we're just missing tests + regenerate the mocks for the image service.

`mockgen -source=pkg/services/images.go > pkg/services/mock_services/images.go`

If you don't have mockgen installed please search for this part on the README.md

pkg/services/images.go
```diff
@@ -488,6 +489,15 @@ func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, er
 	return image, nil
 }
 
+func (s *ImageService) CheckImageName(image *models.Image) (bool, error) {
```
acosferreira, line 492: Can you add some test for it?

pkg/routes/images.go
```diff
@@ -49,6 +49,7 @@ func MakeImagesRouter(sub chi.Router) {
 		r.Post("/repo", CreateRepoForImage)
 		r.Post("/kickstart", CreateKickStartForImage)
 		r.Post("/update", CreateImageUpdate)
+		r.Post("/checkImageName", CheckImageName)
```
acosferreira, line 52: As new endpoint can you please add the documentation?

PR #296, opened 2021-10-06
call reboot inside create update

acosferreira:


pkg/services/updates.go
```diff
@@ -139,6 +144,8 @@ func (s *UpdateService) CreateUpdate(update *models.UpdateTransaction) (*models.
 	update.Status = models.UpdateStatusSuccess
 	db.DB.Save(update)
 	log.Infof("Repobuild::ends: update record %#v ", update)
+
+	s.RebootDevice(update)
```
biancarosa, line 148: I think you're missing a WaitGroupReboot.Wait after this call.

And we can also declare the wait group inside of the RebootDevice function and return it, instead of using as package-var, as its intent is a bit different from the image wait group

acosferreira: should I add the WaitGroupReboot here or inside of RebootDevice? Cause, I'm tryng to make sure that will call this func, and add the assync inside of it, thoughts?

PR #297, opened 2021-10-08
Fix: Add packages to img instead of commit

biancarosa:
Goes together with https://github.com/RedHatInsights/edge-frontend/pull/179

Missing IQE tests fix, will mark as ready for review when this is done to avoid merging. Please give me reviews beforehand (while on draft) so I can update the code.


pkg/models/images.go
```diff
@@ -32,7 +32,7 @@ type Image struct {
 	InstallerID  *uint          `json:"InstallerID"`
 	Installer    *Installer     `json:"Installer"`
 	ImageSetID   *uint          `json:"ImageSetID"` // TODO: Wipe staging database and set to not nullable
-	ID           uint           `gorm:"primarykey" json:"ID"`
+	Packages     []Package      `json:"Packages" gorm:"many2many:images_packages;"`
```
acosferreira, line 35: Should this be nullable?

acosferreira: do we show this required packages in somewhere?
In my opinion, as they are required, should not count as "packages", once the list of packages are chosen by the user and there is no visibility around those required.

biancarosa: actually, i don`t think we save this, so we are cool, we don't send to the user, you're correct.

anyways, i still don`t think it is useful to keep this nullable as this is as a list, it makes more sense to keep it empty to avoid checking for null values while looping. an empty list and a null list are semantically the same, but this doesn't happen with primary data types.

biancarosa:
Speaking with Tomer, it makes more sense to merge the changes on the API and then regenerate the client.

PR #301, opened 2021-10-12
Creating ThirdyPartyRepo endpoint

Saurabh-Thakre:


tpapaioa:
FYI, "Thirdy" should be "Third" everywhere.

PR #306, opened 2021-10-14
Add context properly to services

biancarosa:
Fixes the error on the get image by id method because the ctx was nil on the method


AaronH88:
Adding the context looks good,

want to add comments in this PR to satisfy the build bot?

PR #316, opened 2021-10-20
Adding ov_de package

Avielyo10:
* Unmarshal CBOR OV into golang structs without going through C.
* Support marshal the whole OV header to JSON.
* Support receiving only relevant data from the OV header.
* Adding some tests without uploading binary files.


pkg/models/ov_de/ov_de.go
```diff
@@ -0,0 +1,74 @@
+// Ownershipvoucher deserialization package from CBOR
+// As for our needs we'll deserialize its header only
+package ov_de
```
loadtheaccumulator, line 3: The dreaded package underscore and comment format linting messages.

Avielyo10: I changed it all, still think this should be ov_de & not ovde

pkg/models/ov_de/ov_resolvers.go
```diff
@@ -0,0 +1,105 @@
...
+	case 0:
+		s = "Crypto"
+	case 1:
+		s = "X509"
+	case 2:
+		s = "COSEX509"
+	case 3:
+		s = "Cosekey"
+	default:
+		fmt.Fprintln(os.Stderr, s)
```
acosferreira, line 22: can we change the fmt to log?

pkg/models/ov_de/ov_de.go
```diff
@@ -0,0 +1,77 @@
...
+	var ovh OwnershipVoucherHeader
+	err := cbor.Unmarshal(ovhb, &ovh)
+	return &ovh, err
+}
+
+// If CBOR unmarshal fails => panic
+// Something might be wrong with OV
+func unmarshalCheck(e error, ovORovh string) {
+	if e != nil {
+		panic(fmt.Sprintf("Can't unmarshal %s from bytes", ovORovh))
```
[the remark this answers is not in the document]
Avielyo10: Great idea, thanks!

PR #320, opened 2021-10-21
add total count and data as reutn on image-sets list

acosferreira:


biancarosa (commented):
Couple of minor things: we could have tests to test that we're using the proper data structure for paginated lists, remove comment and turn this map interface into a struct

biancarosa (approved):

PR #330, opened 2021-10-28
changed path to query

tomershinhar:
changed the filtration and sorting parameters from path to query


tomershinhar:
I also changed the parameter names on the response scheme to match the actual API response (Count and Data vs count and data).
since this is inconsistent with the other list endpoints (listImages and GetAllThirdPartyRepo) we might want to change the API response itself to match the others

maxamillion (approved):

PR #335, opened 2021-11-01
Read from playbook-dispatcher's Kafka topic and logs the message

biancarosa:


pkg/services/consumers.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
+
+	"github.com/segmentio/kafka-go"
```
loadtheaccumulator, line 12: Curious on reasoning for going with segmentio vs the github.com/confluentinc/confluent-kafka-go package?

biancarosa: No particular preference. Went for the first search, has 4k stars and looked safe, and by accident ended up finding more code at Red Hat using the same lib. Apparently we have other people using sarama, too, but didn't even knew about confluent-kafka-go. Any arguments in favor of it instead of kafka-go? Happy to change.

biancarosa: No particular reason. Found in my first search, tried and liked and it worked, and found some red hat services using it and sarama too.

Any reasons why I should change? I'm happy to change if needed.

loadtheaccumulator: The confluent lib is what I've used, seen other c.rh.c apps use, and the RH managed service leverages in their examples. No reason other than that. I haven't dived into a comparison and was curious.

thom-at-redhat: /lgtm

pkg/services/consumers.go
```diff
@@ -0,0 +1,76 @@
...
+	topic := "platform.playbook-dispatcher.runs"
+	brokers := make([]string, len(s.config.Brokers))
+	for i, b := range s.config.Brokers {
+		brokers[i] = fmt.Sprintf("%s:%d", b.Hostname, b.Port)
+	}
+	// make a new reader that consumes from topic from this consumer group
+	r := kafka.NewReader(kafka.ReaderConfig{
+		Brokers: brokers,
+		Topic:   topic,
+		GroupID: "edge-fleet-management",
```
loadtheaccumulator, line 44: We'll need to discuss groupid's as we consume for various functions. It might be useful to make it more specific now.

loadtheaccumulator (changes requested):
Dropping this here to prompt a quick discussion.

pkg/services/consumers.go
biancarosa: makes sense. let's add that discussion to event based meeting or arch meeting?

loadtheaccumulator (approved):

maxamillion (approved):

PR #337, opened 2021-11-03
remove reboot playbook from dockerfile

loadtheaccumulator:


biancarosa (approved):
Might be worthy to also remove the code that calls it.

loadtheaccumulator:
> Might be worthy to also remove the code that calls it.

Ah, yes. The call was removed before, but I can include removing the orphaned functions in this PR.

maxamillion (approved):

PR #343, opened 2021-11-04
Enhance error handling during update creation

biancarosa:
This will be helpful while handling playbook dispatcher's response through the kafka topic and I captured some errors while testing this, so I did some enhancement and added the updates on the waitgroup to avoid having them hanging around with building status.


acosferreira (approved):
LGTM

PR #344, opened 2021-11-05
Fix internal server errors in custom repo API due to missing context.

tpapaioa:
This PR fixes the following issues:
- Remove `/delete` and `/update` from the `DELETE` and `PUT` API endpoints. The new endpoints are now:
```
DELETE /thirdpartyrepo/{ID}
PUT /thirdpartyrepo/{ID}
```
- Add missing context variable to the `ThirdPartyRepoService` instance, to resolve internal server errors for `DELETE|PUT|GET /thirdpartyrepo/{ID}` calls when the edge-api service is run with `AUTH=true` (as is the case when deployed via clowder).

### DIFF
--- a/cmd/spec/openapi.json
+++ b/cmd/spec/openapi.json
@@ -2845,45 +2845,6 @@
       }
     },
     "/thirdpartyrepo/{ID}": {
-      "get": {
-        "operationId": "GetThirdPartyRepoByID",
-        "parameters": [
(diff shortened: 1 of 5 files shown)


biancarosa (approved):
TY @tpapaioa :) This is really great :)

omaciel (approved):

PR #347, opened 2021-11-05
Poc/inventory endpoint

acosferreira:


main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 		s.Route("/updates", routes.MakeUpdatesRouter)
 		s.Route("/image-sets", routes.MakeImageSetsRouter)
 		s.Route("/devices", routes.MakeDevicesRouter)
+		s.Route("/inventory", routes.MakeInventoryRouter)
```
biancarosa, line 97: Can't we have this under our current devices route rather than creating a new one?

pkg/clients/inventory/client.go
```diff
@@ -39,8 +41,11 @@ type Response struct {
 
 // Devices represents the struct of a Device on Inventory API
 type Devices struct {
-	ID     string        `json:"id"`
-	Ostree SystemProfile `json:"system_profile"`
+	ID              string `json:"id"`
+	DisplayName     string `json:"display_name"`
+	LastSeen        string `json:"updated"`
+	UpdateAvailable bool
```
biancarosa, line 47: Can we keep the camel_case contract and add a `update_available` json tag here?

pkg/clients/inventory/client.go
```diff
@@ -59,13 +64,59 @@ const (
...
-	filterParams = "?filter[system_profile][host_type]=edge&fields[system_profile]=host_type,operating_system,greenboot_status,greenboot_fallback_detected,rpm_ostree_deployments,rhc_client_id,rhc_config_state"
+	Fields       = "host_type,operating_system,greenboot_status,greenboot_fallback_detected,rpm_ostree_deployments,rhc_client_id,rhc_config_state"
+	FilterParams = "?staleness=fresh&filter[system_profile][host_type]=edge&fields[system_profile]=host_type,operating_system,greenboot_status,greenboot_fallback_detected,rpm_ostree_deployments,rhc_client_id,rhc_config_state"
 )
 
+type InventoryParams struct {
+	PerPage      string
+	Page         string
+	OrderBy      string
+	OrderHow     string
```
biancarosa, line 75: This is normally called Sort

pkg/clients/inventory/client.go
```diff
@@ -59,13 +64,59 @@ const (
...
+	HostnameOrId string
```
biancarosa, line 76: Identifier, maybe? With a comment saying it can be a Hostname or ID. We can also have the two fields for clarity and one is supposed to be always empty.

pkg/routes/devices.go
```diff
@@ -15,6 +15,7 @@ import (
 
 // MakeDevicesRouter adds support for operations on update
 func MakeDevicesRouter(sub chi.Router) {
+	sub.With(InventoyCtx).Get("/", GetInventory)
```
biancarosa, line 18: Do we need InventoryCtx function? This function is normally for get by id endpoints. If this is for filtering, I suggest keeping on the GetInventory method.

pkg/clients/inventory/client.go
biancarosa: I just saw that Inventory uses OrderHow. We can rename it or keep it like that 🤷🏻

pkg/routes/devices.go
acosferreira: we need to build the parameters that we will send to inventory api call

pkg/clients/inventory/client.go
acosferreira: I'm using the exactly parameters that we send to inventory api call

pkg/clients/inventory/client.go
acosferreira: this is the response that we are re-building to FE add the info with intentory and edgeapi joined

pkg/routes/devices.go
biancarosa: Can we do that directly on the GetInventory function?

biancarosa (commented):
To sum up, my suggestions are: 
- fix vet comments
- get rid of the InventoryCtx function and get the params straight on GetInventory for a more simpler logic on params retrieval since it's not used by another route

pkg/routes/devices.go
```diff
@@ -40,6 +42,7 @@ func DeviceCtx(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var uCtx DeviceContext
 		uCtx.DeviceUUID = chi.URLParam(r, "DeviceUUID")
+		fmt.Printf(":::: Device CTX %v\n ::::", uCtx)
```
biancarosa, line 45: a lost print has been found

pkg/routes/inventory.go
```diff
@@ -0,0 +1,98 @@
...
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/redhatinsights/edge-api/pkg/clients/inventory"
+	"github.com/redhatinsights/edge-api/pkg/dependencies"
+	"github.com/redhatinsights/edge-api/pkg/errors"
+	"github.com/redhatinsights/edge-api/pkg/services"
+)
+
+// MakeDevicesRouter adds support for operations on invetory
```
biancarosa, line 15: comment doesnt match function name

biancarosa: also see github actions comments about for other structs :)

pkg/routes/inventory.go
```diff
@@ -0,0 +1,98 @@
...
+	InventoryData.Results = results
+
+	json.NewEncoder(w).Encode(InventoryData)
+}
+
+func GetUpdateAvailableInfo(param *inventory.InventoryParams, r *http.Request, inventoryResp inventory.Response) (IvtResponse []InventoryResponse) {
+	var ivt []InventoryResponse
+	services, _ := r.Context().Value(dependencies.Key).(*dependencies.EdgeAPIServices)
+	deviceService := services.DeviceService
+	// ctx := r.Context()
```
biancarosa, line 72: can we get rid of the comments?

biancarosa (changes requested):

PR #349, opened 2021-11-08
Create FDO client

Avielyo10:
- Create FDO client - aligned with [ov_management_api.md](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/specs/ov_management_api.md)
- Move ownershipvoucher deserializer out
- Create test container for libfdo_data
- Use libfdo_data test container for coverage testing
- Add tests for FDO client


config/config.go
```diff
@@ -87,6 +90,9 @@ func Init() {
 	options.SetDefault("TemplatesPath", "/usr/local/etc/")
 	options.SetDefault("EdgeAPIBaseURL", "http://localhost:3000")
 	options.SetDefault("UploadWorkers", 100)
+	options.SetDefault("FDOHostURL", "fdo.redhat.com")
```
biancarosa, line 93: Can we set those variables on the clowapp.yaml already to avoid further commits in order to customize it for other environments?

biancarosa (changes requested):

config/config.go
Avielyo10: Yes, done

Avielyo10:
@biancarosa Well for now the secret is empty, do you know where `PSK_DISPATCHER` value get received?

maxamillion (approved):
LGTM :+1:

PR #352, opened 2021-11-09
Fix: Set image status as error while creating installer if it fails

biancarosa:
**The problem**

When creating an installer on image builder, if that request fails, we were not setting the image status with an error.

**The solution**

Set the error status properly when an error occurs while sending this request, alongside with some refactors on tests to use the new ginko and gomega libraries.


acosferreira (approved):
LGTM

PR #353, opened 2021-11-10
Bug fix for 1310, 1200

Saurabh-Thakre:
**Now, PUT response returns the correct value for CreatedAt and UpdatedAt**
```
curl --request PUT --url http://localhost:3000/api/edge/v1/thirdpartyrepo/8/  --data '{                                                      
    "URL": "http://www.hmm.com/in/repo/hmmm",
    "Description": "I will have a nice sleep now",
    "Name": "Sleep time"                                                                                                         
}' | jq

{
  "ID": 8,
  "CreatedAt": "2021-11-10T15:52:33.683344213+05:30",
  "UpdatedAt": "2021-11-11T00:55:14.174617616+05:30",
  "DeletedAt": null,
  "Name": "Sleep time",
  "URL": "http://www.hmm.com/in/repo/hmmm",
  "Description": "I will have a nice sleep now",
  "Account": "0000000"
}

```
**Now, we have a better response for DELETE and GET requests for non-existent and deleted objects**
```
curl --request DELETE --url http://localhost:3000/api/edge/v1/thirdpartyrepo/8/ | jq      
 
"Successfully deleted third party repository with ID 8"

```
```
curl --request GET --url http://localhost:3000/api/edge/v1/thirdpartyrepo/10 | jq    

{
  "Code": "NOT_FOUND",
  "Status": 404,
  "Title": "Third Party Repository is not found"
}

```

### DIFF
--- a/cmd/spec/openapi.json
+++ b/cmd/spec/openapi.json
@@ -2859,8 +2859,15 @@
           }
         ],
         "responses": {
-          "204": {
-            "description": "Successfully deleted third party repository with ID {ID}"
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/v1.CreateThirdPartyRepo"
+                }
+              }
+            },
+            "description": "OK"
           },
           "404": {
             "content": {
@@ -2870,7 +2877,7 @@
                 }
               }
             },
-            "description": "Third Party Repository is not found."
+            "description": "third party repository was not found."
           },
           "500": {
             "content": {
@@ -2917,7 +2924,7 @@
                 }
               }
             },
-            "description": "Third Party Repository is not found."
+            "description": "third party repository was not found."
           },
           "500": {
             "content": {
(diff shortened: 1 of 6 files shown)


cmd/spec/openapi.json
```diff
@@ -2859,15 +2859,18 @@
...
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/v1.CreateThirdPartyRepo"
+                  "$ref": "#/components/schemas/v1.NotFound"
                 }
               }
             },
-            "description": "OK"
+            "description": "Third Party Repository is not found."
```
omaciel, line 2873: s/is/was?

omaciel:
Change `is` to `was` and perhaps keep the same case for `third party repositories` as you have some written with capital letters and some in lowercase.

biancarosa (changes requested):
I think that, just for that DELETE method, we could use Unscoped to return the deleted data and return a proper 200 for the operation:

https://gorm.io/docs/delete.html#Soft-Delete

biancarosa (approved):

PR #354, opened 2021-11-10
Fix/image tar repository

acosferreira:
add upload repo.tar on our bucket and use this file instead ImageBuildTarURL to avoid problem due apply updates using commits older that 7 days

### DIFF
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -17,7 +17,6 @@ type Commit struct {
 	ImageBuildHash       string             `json:"ImageBuildHash"`
 	ImageBuildParentHash string             `json:"ImageBuildParentHash"`
 	ImageBuildTarURL     string             `json:"ImageBuildTarURL"`
-	TarRepoURL           string             `json:"TarRepoURL"`
 	OSTreeCommit         string             `json:"OSTreeCommit"`
 	OSTreeParentCommit   string             `json:"OSTreeParentCommit"`
 	OSTreeRef            string             `json:"OSTreeRef"`
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -17,6 +17,7 @@ type Commit struct {
 	ImageBuildHash       string             `json:"ImageBuildHash"`
 	ImageBuildParentHash string             `json:"ImageBuildParentHash"`
 	ImageBuildTarURL     string             `json:"ImageBuildTarURL"`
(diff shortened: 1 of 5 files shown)


biancarosa (changes requested):
This will upload the file every time we get the repo, and we only want to it on the first time, my suggestion is to break this function in order to call the upload tar on image service only during image creation

pkg/models/commits.go
```diff
@@ -17,6 +17,7 @@ type Commit struct {
 	ImageBuildHash       string             `json:"ImageBuildHash"`
 	ImageBuildParentHash string             `json:"ImageBuildParentHash"`
 	ImageBuildTarURL     string             `json:"ImageBuildTarURL"`
+	TarRepoURL           string             `json:"TarRepoURL"`
```
biancarosa, line 20: Maybe just TarURL? The tar itself is just a commit and not yet a repo. Or we can use the same ImageBuildTarURL.

pkg/services/images.go
```diff
@@ -692,3 +695,17 @@ func (s *ImageService) RetryCreateImage(image *models.Image) error {
 
 	return nil
 }
+
+func uploadTarRepo(imageName string, repoId int) (string, error) {
+	fmt.Printf("::uploadTarRepo::\n")
```
biancarosa, line 700: a print has been found and check github actions build comment below

pkg/services/images.go
```diff
@@ -692,3 +695,17 @@ func (s *ImageService) RetryCreateImage(image *models.Image) error {
...
+	uploadPath := fmt.Sprintf("/tar/%v/%s", repoId, imageName)
+	fmt.Printf("::uploadPath:: %v\n", uploadPath)
```
biancarosa, line 702: a print has been found

pkg/services/images.go
```diff
@@ -692,3 +695,17 @@ func (s *ImageService) RetryCreateImage(image *models.Image) error {
...
+	filesService := NewFilesService()
+	url, err := filesService.GetUploader().UploadFile(imageName, uploadPath)
+	fmt.Printf("::TAR REPO URL:: %v\n", url)
```
biancarosa, line 705: a print has been found

biancarosa (approved):
I think this should be fine, let's give it a shot :)

PR #357, opened 2021-11-12
fix/checkImageName route on OpenAPI files

de1987:
The **checkImageName** route is wrong on the OpenAPI file and it's affecting the tests.


biancarosa (approved):
lgtm

PR #359, opened 2021-11-16
Feat/details image set

acosferreira:
add package and installer info to be used on new image set detail screen


biancarosa (approved):
LGTM

PR #361, opened 2021-11-16
Add libfdo_data shared object into edge-api

Avielyo10:
This is an interim solution, till libfdo_data will start shipping with RHEL

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,11 @@ COPY --from=builder /src/mypackage/myapp/pkg/services/templateKickstart.ks /usr/
 # template to playbook dispatcher
 COPY --from=builder /src/mypackage/myapp/pkg/services/template_playbook/template_playbook_dispatcher_ostree_upgrade_payload.yml /usr/local/etc
 
+# interim FDO requirements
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -34,6 +34,11 @@ COPY --from=builder /src/mypackage/myapp/pkg/services/templateKickstart.ks /usr/
 # template to playbook dispatcher
 COPY --from=builder /src/mypackage/myapp/pkg/services/template_playbook/template_playbook_dispatcher_ostree_upgrade_payload.yml /usr/local/etc
 
+# interim FDO requirements
+ENV LD_LIBRARY_PATH /usr/local/lib
+COPY --from=quay.io/ayosef/libfdo-data:v1 /usr/local/lib/libfdo_data.so /usr/local/lib/libfdo_data.so
```
AaronH88, line 39: Should we set up a team quay and not use ones linked to our account?

Avielyo10: This lib will be shipped with RHEL so for now it's just interim 

Also I do agree with having a team account

PR #368, opened 2021-11-18
Bug fixes

Saurabh-Thakre:
Fixed :
- missing the ID in response, when creating a third party repository. fix to return correct value of created_at and updated_at
- fixing third party duplicate name issues

**Before :**

> curl --request POST --url http://localhost:3000/api/edge/v1/thirdpartyrepo/  --data '{                                                     
    "URL": "http://www.hmm.mmh/in/repo/hmmm",
    "Description": "hmm mmh hmm mmh",
    "Name": "X"                                                                                                         
}' | jq
{
  "CreatedAt": "0001-01-01T00:00:00Z",
  "UpdatedAt": "0001-01-01T00:00:00Z",
  "DeletedAt": null,
  "Name": "X",
  "URL": "http://www.hmm.mmh/in/repo/hmmm",
  "Description": "hmm mmh hmm mmh",
  "Account": ""
}

**After :** 

> curl --request POST --url http://localhost:3000/api/edge/v1/thirdpartyrepo/  --data '{                                                     
    "URL": "http://www.hmm.mmh/in/repo/hmmm",
    "Description": "hmm mmh hmm mmh",
    "Name": "XX"                                                                                                         
}' | jq
{
  "ID": 38,
  "CreatedAt": "2021-11-18T15:59:31.067563425+05:30",
  "UpdatedAt": "2021-11-18T15:59:31.067563425+05:30",
  "DeletedAt": null,
  "Name": "XX",
  "URL": "http://www.hmm.mmh/in/repo/hmmm",
  "Description": "hmm mmh hmm mmh",
  "Account": "0000000"
}


pkg/routes/thirdpartyrepo.go
```diff
@@ -69,6 +69,10 @@ func CreateThirdPartyRepo(w http.ResponseWriter, r *http.Request) {
 		json.NewEncoder(w).Encode(&err)
 		return
 	}
+	result := db.DB.Where("name = ? and account = ?", tprepo.Name, account).First(&tprepo)
+	if result.Error != nil {
+		errors.NewInternalServerError()
```
biancarosa, line 74: Will this return the error? I think it just creates the variable, but doesn't change the response code or body

biancarosa (changes requested):

pkg/services/thirdpartyrepo.go
```diff
@@ -40,6 +40,10 @@ func (s *ThirdPartyRepoService) CreateThirdPartyRepo(tprepo *models.ThirdPartyRe
 			Description: tprepo.Description,
 			Account:     account,
 		}
+		repoDetails := db.DB.Where("name = ? and account = ?", tprepo.Name, account).First(&tprepo)
+		if tprepo.Name == repoDetails.Name() {
+			return errors.NewBadRequest("third party repository with already exists with same")
```
biancarosa, line 45: Can't we achieve this with some sort of unique index on the database? Check if GORM has some json tag for uniqueness, and then the db.Create should fail without any further logic

Saurabh-Thakre: I've to look if GORM has something for us. I'll see if I can find anything

pkg/routes/thirdpartyrepo.go
Saurabh-Thakre: yeah actually the result doesn't return error.

omaciel:
Do we need a DeletedAt field?

Saurabh-Thakre:
> Do we need a DeletedAt field?

Good question, it comes from the Model struct, so we have it in all our endpoints. So when we delete an object, say a third party repo, we get a response with the DeletedAt field updated saying you've deleted this now at this time.

biancarosa (changes requested):
Just realized what's going on there!

Suggestion is to make the method CreateThirdPartyRepo return the newly saved tprepo object and an error instead of just an error, in that case that query to retrieve the value is not necessary on the route.

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

pkg/routes/thirdpartyrepo.go
Saurabh-Thakre: > Just realized what's going on there!
> 
> Suggestion is to make the method CreateThirdPartyRepo return the newly saved tprepo object and an error instead of just an error, in that case that query to retrieve the value is not necessary on the route.



Your suggestion was perfect, and its done now :)

pkg/routes/thirdpartyrepo.go
```diff
@@ -39,7 +39,7 @@ type CreateTPRepoRequest struct {
 
 // CreateThirdPartyRepo creates Third Party Repository
 func CreateThirdPartyRepo(w http.ResponseWriter, r *http.Request) {
-	services, _ := r.Context().Value(dependencies.Key).(*dependencies.EdgeAPIServices)
+	s, _ := r.Context().Value(dependencies.Key).(*dependencies.EdgeAPIServices)
```
Avielyo10, line 42: I don't see a reason for not being explicit

pkg/routes/thirdpartyrepo.go
```diff
@@ -60,7 +60,7 @@ func CreateThirdPartyRepo(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = services.ThirdPartyRepoService.CreateThirdPartyRepo(tprepo, account)
+	tprepo, err = s.ThirdPartyRepoService.CreateThirdPartyRepo(tprepo, account)
```
Avielyo10, line 63: Can we change it to something readable like `thirdPartyRepo`?or at least `tpRepo` as Golang camel case notation?

Avielyo10 (changes requested):

pkg/routes/main_test.go
```diff
@@ -14,6 +14,7 @@ import (
 var (
 	testImage models.Image
 	testRepo  models.Repo
+	tprepo    models.ThirdPartyRepo
```
Avielyo10, line 17: I don't see were this been used at the test

Saurabh-Thakre: `mockThirdPartyRepoService.EXPECT().CreateThirdPartyRepo(gomock.Any(), gomock.Any()).Return(&tprepo, nil)`

This is getting used here :) @Avielyo10

pkg/routes/thirdpartyrepo.go
Saurabh-Thakre: thirdPartyRepo, sure I can make that happen o/

pkg/routes/main_test.go
Avielyo10: Does this teats your fix?

Saurabh-Thakre: yeah kinda, I had to modify services to return 2 values so yeah :)

pkg/services/thirdpartyrepo.go
```diff
@@ -32,7 +32,7 @@ type ThirdPartyRepoService struct {
 }
 
 // CreateThirdPartyRepo creates the ThirdPartyRepo for an Account on our database
-func (s *ThirdPartyRepoService) CreateThirdPartyRepo(tprepo *models.ThirdPartyRepo, account string) error {
+func (s *ThirdPartyRepoService) CreateThirdPartyRepo(tprepo *models.ThirdPartyRepo, account string) (*models.ThirdPartyRepo, error) {
 	if tprepo.URL != "" && tprepo.Name != "" {
 		tprepo = &models.ThirdPartyRepo{
```
Avielyo10, line 37: ditto

pkg/services/thirdpartyrepo.go
```diff
@@ -42,12 +42,12 @@ func (s *ThirdPartyRepoService) CreateThirdPartyRepo(tprepo *models.ThirdPartyRe
...
 		result := db.DB.Create(&tprepo)
 		if result.Error != nil {
-			return result.Error
+			return nil, result.Error
 		}
 		log.Infof("Getting ThirdPartyRepo info: repo %s, %s", tprepo.URL, tprepo.Name)
 
 	}
-	return nil
+	return tprepo, nil
```
Avielyo10, line 50: ditto

pkg/services/thirdpartyrepo.go
Saurabh-Thakre: cool , but then other functions will look different having tprepo haha. I will open new PR for those as well, tprepo is all around this endpoint

biancarosa (approved):
LGTM

Avielyo10 (approved):
lgtm

PR #378, opened 2021-11-21
FDO service & route

Avielyo10:
- CGO_ENABLED=1 - FDO uses cgo with libfdo_data, needed for compilation 
- mod tidy - clean
- Use vet inside of test container - vet will try to use libfdo_data, the test-container has it inside
- Improve test-container - not every OS support libfdo_data.so, this will help build & test things from your repo & branch 
- FDO service & route - created


Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

Avielyo10:
@Mergifyio refresh

main.go
```diff
@@ -95,6 +95,7 @@ func main() {
 		s.Route("/image-sets", routes.MakeImageSetsRouter)
 		s.Route("/devices", routes.MakeDevicesRouter)
 		s.Route("/thirdpartyrepo", routes.MakeThirdPartyRepoRouter)
+		s.Route("/fdo", routes.MakeFDORouter)
```
AaronH88, line 98: Are we ok calling the endpoint `fdo`? 

Wondering as the spec references it as `fido`

Avielyo10: FDO = FIDO Device Onboarding
FIDO = Fast ID Online

Avielyo10:
@maxamillion @biancarosa

Avielyo10:
@Mergifyio refresh

.github/workflows/go.yml
```diff
@@ -31,7 +31,11 @@ jobs:
         run: make lint
 
       - name: Vet
-        run: make vet
+        uses: addnab/docker-run-action@v3
+        with:
+          image: quay.io/ayosef/libfdo-data:v1
```
biancarosa, line 36: Just wondering if we should use private images here. Is that image def open source and on GitHub? Can we create the repo under RedHatInsights org?

Dockerfile
```diff
@@ -10,11 +10,17 @@ ENV GO111MODULE=on
 # Using go get requires root.
 USER root
 RUN go get -d -v
+
+# interim FDO requirements
+ENV LD_LIBRARY_PATH /usr/local/lib
+COPY --from=quay.io/ayosef/libfdo-data:v1 ${LD_LIBRARY_PATH}/libfdo_data.so.0 ${LD_LIBRARY_PATH}/libfdo_data.so.0
```
biancarosa, line 16: Same as above. I'd avoid using private images hosted on quay that the team is not sure where the source code lives, because if needed we'll have to change it and it is important that everyone has the knowledge and access to update it.

.github/workflows/go.yml
Avielyo10: The `Dockerfile` for this image is under `edge-api`, `test-container` directory
Here I also updating the `Dockerfile`, I was wondering how can we push it to cloudservices but I don't have the permissions to push

Dockerfile
Avielyo10: The `Dockerfile` for this image is under `edge-api`, `test-container` directory
Here I also updating the `Dockerfile`, I was wondering how can we push it to cloudservices but I don't have the permissions to push

Avielyo10:
@biancarosa I've created quay.io/cloudservices/edge-api:libfdo-data which is the internal version.
Github actions can't pull it without secrets so for Github actions we can use the private for fast tests && coverage, and for building edge-api binary we will use the internal version. WDYT?

biancarosa (approved):
It LGTM! :)

We will need the spec on the openapi json regardless of the markdown spec written, afaik, but this can come in a different PR. Maintaning both might be tricky so I'd suggest sticking with just the one on the openapi file.

Avielyo10:
I created the spec for FE, so this will be much easier for them :)
I'll create a separate PR for openapi, thanks !

PR #388, opened 2021-11-23
ci(Mergify): configuration update

Avielyo10:
This change has been made by @Avielyo10 from https://mergify.com config editor.


Avielyo10:
@mergify has bug, remove the `ready` label as for now

PR #395, opened 2021-11-24
Set error status during migration

biancarosa:
# Description

Due to a issue where we are discussing how to handle retries and updates to images with error status, I came up with this strategy to avoid images to get stuck on building status. With our current architecture the migration is the moment where we should set the status with an error and when we migrate to Kafka topics we can revisit that.

Fixes # (issue)

## Type of change

What is it?

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Tests update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I run `go fmt ./...` to check that my code is properly formatted
- [x] I run `go vet ./...` to check that my code is free of common Go style mistakes

PR #391, opened 2021-11-23
Fix: Allow users to update image with error

biancarosa:
# Description

This pull-requests adds the ability to update a failed version of an image, as the user might change something on the next version and/or retry without changing anything due to internal server errors.

Fixes #THEEDGE-1278

## Type of change

What is it?

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Tests update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I run `go fmt ./...` to check that my code is properly formatted
- [x] I run `go vet ./...` to check that my code is free of common Go style mistakes

PR #396, opened 2021-11-24
Fix: retry image build

biancarosa:
# Description

On top of #395  and #391, THEEDGE-1278 also requires the retry endpoint to be fully working and it wasn't. This was mostly due to a logic on the post process that considered the image done due to its error status, so I merely did some status changes on the image to allow the postProcessImage method to retry the build. The commit itself doesn't get built again on img builder, just the installer, but I assume this is okay for now as most errors are on our side.

Fixes #THEEDGE-1278

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update

# Checklist:

- [x] I have made corresponding changes to the documentation


cmd/spec/openapi.json
```diff
@@ -3,9 +3,7 @@
     "schemas": {
       "v1.AddUpdate": {
         "properties": {
-          "CommitID": {
-            "type": "integer"
-          },
+          "CommitID": {},
```
acosferreira, line 6: any reason to change it?
The commit Id should be an integer when we call the endpoint and wasn't being displayed correctly

biancarosa: I just regenerated, the fact that is putting this is mostly because of bug on the openapi lib we are using. I contributed to the code by adding it: https://github.com/getkin/kin-openapi/pull/379 but I don't think it has been released...? I'm gonna check if we can update this lib.

biancarosa: Actually, on a second thought, this shouldn't affect the type (I guess) so I'll take a closer look.

biancarosa: @acosferreira thanks for pointing out, updated the lib and the generate docs came out with the type :)

acosferreira (approved):
LGTM

PR #400, opened 2021-11-25
Make FK nullable

Avielyo10:
# Description

Bug fix, make FK nullable

Fixes `insert or update on table "devices" violates foreign key constraint "fk_devices_ownership_voucher_data"`

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update

# Checklist:

### DIFF
--- a/pkg/models/updates.go
+++ b/pkg/models/updates.go
@@ -15,7 +15,7 @@ type Device struct {
 	DesiredHash            string                `json:"DesiredHash"`
 	RHCClientID            string                `json:"RHCClientID"`
 	Connected              bool                  `gorm:"default:true" json:"Connected"`
-	OwnershipVoucherDataID int                   `json:"-"`
+	OwnershipVoucherDataID *uint                 `json:",omitempty"`
 	OwnershipVoucherData   *OwnershipVoucherData `json:"fdo,omitempty"`
 }
 


pkg/models/updates.go
```diff
@@ -15,7 +15,7 @@ type Device struct {
 	DesiredHash            string                `json:"DesiredHash"`
 	RHCClientID            string                `json:"RHCClientID"`
 	Connected              bool                  `gorm:"default:true" json:"Connected"`
-	OwnershipVoucherDataID int                   `json:"-"`
+	OwnershipVoucherDataID uint                  `json:"-,omitempty"`
```
biancarosa, line 18: I think you need a pointer of uint here

biancarosa (changes requested):

pkg/models/updates.go
Avielyo10: How so?other stucts has only `uint`

Avielyo10: Changed, I don't get why the error is only on `Device`

biancarosa (approved):

pkg/models/updates.go
biancarosa: We found when inserting a device, but it might happen somewhere else that has an FK. Be aware that maybe testing with sqlite locally might cause the error to not show up, only postgres complains about this

PR #421, opened 2021-12-05
spec-file validation workflows

tomershinhar:
# Description

added jobs to the github workflows to validate the spec file is matching the openapi specifications.
note that this will only validate that the syntax is correct, it will not validate that the spec-file matches the API behavior

What is it?

- [ ] Documentation update
- [ x] Tests update


Avielyo10 (approved):
I'd move it to another place but so far so good

PR #466, opened 2021-12-15
Security check

Avielyo10:
# Description

Fix all of security alerts on our project & implement a mechanism to catch them in the future 

Fixes security possible risks

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update

# Checklist:

- [x] I have made corresponding changes to the documentation


pkg/clients/fdo/client.go
```diff
@@ -69,7 +69,7 @@ func (c *Client) BatchDelete(fdoUUIDList []string) (interface{}, error) {
 // Decode response body into json
 func decodeResBody(body *io.ReadCloser) interface{} {
 	var j interface{}
-	json.NewDecoder(*body).Decode(&j)
+	_ = json.NewDecoder(*body).Decode(&j)
```
biancarosa, line 72: I think we'd benefit from actually handling the errors in all of those cases, what do you think?

Avielyo10: I think this can be discussed within the team, since we control our data we decode (for most cases) we can catch errors on our testsuites  (e.g. a request without a response will alert).
If we will try to catch those decoding errors for most cases we won't be able to cover it, (e.g. creating an input that will trigger a decoding error) but I don't mind to try.

biancarosa: But what if a customer starts using the API? I understand that we control the requests from the front-end, but the API should still handle the errors separately as a completely separate product as we might have some customers willing to automate things on their side. +1 for discussing with the team!

Avielyo10: I meant, we construct the struct, we can check validity while constructing

PR #499, opened 2021-12-23
Make private Context edge-api services key.

ldjebran:
# Description
Make the context key of edge api services private, this will prevent future bugs. and simplify code reading.

Fixes possible future bugs.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

# Checklist:

### DIFF
--- a/pkg/dependencies/main.go
+++ b/pkg/dependencies/main.go
@@ -43,16 +43,27 @@ func Init(ctx context.Context) *EdgeAPIServices {
 	}
 }
 
-type key int
+type servicesKeyType int
 
-// Key is the context key for dependencies on the request context
-const Key key = iota
+// servicesKey is the context key for dependencies on the request context
+const servicesKey servicesKeyType = iota
+
+// ContextWithServices add edge apis services to context
+func ContextWithServices(ctx context.Context, services *EdgeAPIServices) context.Context {
+	return context.WithValue(ctx, servicesKey, services)
+}
+
+// ServicesFromContext return the edge api services from context
+func ServicesFromContext(ctx context.Context) (*EdgeAPIServices, bool) {
+	edgeAPIServices, ok := ctx.Value(servicesKey).(*EdgeAPIServices)
+	return edgeAPIServices, ok
+}
 
 // Middleware is the dependencies Middleware that serves all Edge API services on the current request context
 func Middleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		services := Init(r.Context())
-		ctx := context.WithValue(r.Context(), Key, services)
+		edgeAPIServices := Init(r.Context())
+		ctx := ContextWithServices(r.Context(), edgeAPIServices)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
--- a/pkg/dependencies/main.go
+++ b/pkg/dependencies/main.go
@@ -43,10 +43,10 @@ func Init(ctx context.Context) *EdgeAPIServices {
 	}
 }
 
-type servicesKeyType int
+type servicesKeyType string
 
 // servicesKey is the context key for dependencies on the request context
-const servicesKey servicesKeyType = iota
+const servicesKey = servicesKeyType("services")
 
 // ContextWithServices add edge apis services to context
 func ContextWithServices(ctx context.Context, services *EdgeAPIServices) context.Context {
@@ -54,9 +54,13 @@ func ContextWithServices(ctx context.Context, services *EdgeAPIServices) context
 }
 
 // ServicesFromContext return the edge api services from context
-func ServicesFromContext(ctx context.Context) (*EdgeAPIServices, bool) {
+func ServicesFromContext(ctx context.Context) *EdgeAPIServices {
 	edgeAPIServices, ok := ctx.Value(servicesKey).(*EdgeAPIServices)
-	return edgeAPIServices, ok
+	if !ok {
+		log.Fatal("Could not get EdgeAPIServices from Context")
+	}
+
+	return edgeAPIServices
 }
 
 // Middleware is the dependencies Middleware that serves all Edge API services on the current request context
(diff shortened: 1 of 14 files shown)


pkg/dependencies/main.go
```diff
@@ -43,16 +43,27 @@ func Init(ctx context.Context) *EdgeAPIServices {
...
+// ContextWithServices add edge apis services to context
+func ContextWithServices(ctx context.Context, services *EdgeAPIServices) context.Context {
+	return context.WithValue(ctx, servicesKey, services)
+}
+
+// ServicesFromContext return the edge api services from context
+func ServicesFromContext(ctx context.Context) (*EdgeAPIServices, bool) {
+	edgeAPIServices, ok := ctx.Value(servicesKey).(*EdgeAPIServices)
+	return edgeAPIServices, ok
+}
```
Avielyo10, line 60: We should catch any errors as we can, I have 2 suggestions:
1. Return only `*EdgeAPIServices` and ignore the check of its existence, then on each `services, _ := dependencies.ServicesFromContext(r.Context())` we can write `services := dependencies.ServicesFromContext(r.Context())` & check `services` for nil.
  ```suggestion
  func ServicesFromContext(ctx context.Context) *EdgeAPIServices {
	      edgeAPIServices, _ := ctx.Value(servicesKey).(*EdgeAPIServices)
	      return edgeAPIServices
  }
  ```
2. Keep this function as it is & for each `services, _ := dependencies.ServicesFromContext(r.Context())` we should write `services, ok := dependencies.ServicesFromContext(r.Context())`, then try to see if `!ok`

I tend to go with the first suggestion, but this is up to you.

Avielyo10 (changes requested):
Thanks for it, this looks much better than before!

pkg/dependencies/main.go
ldjebran: if context value does not exist a nil will be returned , and the cast will fail with ok = false , so the check will be from the side of the caller function.

ldjebran: Fixing caller function is not the purpose of this PR.

ldjebran: @Avielyo10 what do you think?

Avielyo10: You are correct saying this is not the PR intention, but if we do a refactor don't you think we should make it all through it is done?so this won't need more refactoring on the same place?
You can create a separate PR for the check if you'd like, this will be very helpful for everyone

You are also correct that its the caller responsibility to check the output of a function, but then I'd suggest sending back an `error` instead of a boolean, this way static code analysis & security checks we want to add will fail since we have an unhandled error. (returning `bool` won't do that)

ldjebran: I think we should not modify ServicesFromContext to return an error , the caller function should generate it's proper error when receiving ok = false , perhaps the caller function will call init function and write a new value , some other function will generate a fatal error, and other function can pass silent if it's not critical, we cannot decide at this place.

ldjebran: Maybe there is a sync-up misunderstanding :) let me explain my point of view

When accessing Context values the code reader is waiting to see a boolean as in all the project we are using that pattern when accessing context values.

Returning errors is a good a idea , as we the caller function will have a ready to use error to return when necessary , but the error should be an instance so the caller can make comparisons.

But we not want to make services context value a special case.

I think If we should return an error let's do it by writing a global public function that access context and return standard project errors like 
ErtContextVaslueNotFound  - in case of key not found
ErrContextValueMismatch  - In case the cast fail

Let's discuss about it after the holidays, Happy Holidays :)

Avielyo10: @ldjebran I think you are missing Golang's error concept, returning an error doesn't mean throwing an error. Errors are values, so each caller can handle it differently but as it comes to code quality check then the caller must handle it & can't ignore it.
```suggestion
func ServicesFromContext(ctx context.Context) (edgeAPIServices *EdgeAPIServices, err error) {
	edgeAPIServices, ok = ctx.Value(servicesKey).(*EdgeAPIServices)
	if !ok {
	    err = errors.New("Couldn't get services from Context")
	}
}
```

Avielyo10:
@biancarosa why do we use `iota`? As each request has its own `Context` which derived from background we can have it a `const string`. I do afraid this can become a problem in the future (2^64 - 1 is actually infinity) do we want to pass those large numbers for each request? (they gets bigger & bigger without an actual necessity)

Avielyo10:
I don't think we should use `iota` in this case, sure this should be unique but:

1. Different packages can still collide. (for example using hardcoded predefined values by other packages)
2. 2^64 - 1 is practically infinity, are we sure we want to pass those big integers as keys?

Referring https://pkg.go.dev/context#WithValue example, the best practice should follows something like:

```go
// edgeAPIKey is the context key for dependencies on the request context
type edgeAPIKey string

// Middleware is the dependencies Middleware that serves all Edge API services on the current request context
func Middleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		services := Init(r.Context())
		ctx := context.WithValue(r.Context(), edgeAPIKey("services"), services)
		next.ServeHTTP(w, r.WithContext(ctx))
	})
}

func ServicesFromContext(ctx context.Context) (edgeAPIServices *EdgeAPIServices, err error) {
	edgeAPIServices, ok := ctx.Value(edgeAPIKey("services")).(*EdgeAPIServices)
	if !ok {
		err = errors.New("couldn't get services from Context")
	}
	return
}

```
@ldjebran @biancarosa WDYT?

pkg/dependencies/main.go
biancarosa: I think we can catch the error here and log as log.Fatal. If something happens, the caller shouldn't have to handle it because this is core functionality.

biancarosa:
I'm good with the suggestion about not using iota, I think that was done before and we just kept the pattern but it's better indeed.

pkg/dependencies/main.go
Avielyo10: @biancarosa I was thinking maybe if there is an error there, then we can return immediately on `Middleware` without forwarding.

ldjebran:
@Avielyo10 @biancarosa all comments allied please review.

Avielyo10 (approved):
lgtm

PR #532, opened 2022-01-10
Pass Third-Party Repositories URLs to /compose API

Saurabh-Thakre:
# Description

Fixes # (issue) THEEDGE-1084

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


biancarosa (commented):
We also need to make sure that the repository being passed belongs to the logged in account

Saurabh-Thakre:
> We also need to make sure that the repository being passed belongs to the logged in account

yeah added the account to DB query

pkg/clients/imagebuilder/client.go
```diff
@@ -47,7 +48,19 @@ type OSTree struct {
...
+
+// Repository is the record of Third Party Repository
+type Repository struct {
+	Baseurl    []string `json:"baseurl,omitempty"`
+	CheckGpg   *bool    `json:"check_gpg,omitempty"`
+	GpgKey     *string  `json:"gpg_key,omitempty"`
+	IgnoreSsl  *bool    `json:"ignore_ssl,omitempty"`
+	Metalink   *string  `json:"metalink,omitempty"`
+	Mirrorlist *string  `json:"mirrorlist,omitempty"`
+	Rhsm       bool     `json:"rhsm"`
```
Avielyo10, line 63: Let's stick with CamelCase

```suggestion
	BaseURL  []string `json:"base_url,omitempty"`
	CheckGPG   *bool    `json:"check_gpg,omitempty"`
	GPGKey     *string  `json:"gpg_key,omitempty"`
	IgnoreSSL  *bool    `json:"ignore_ssl,omitempty"`
	Metalink   *string  `json:"metalink,omitempty"`
	Mirrorlist *string  `json:"mirrorlist,omitempty"`
	RHSM       bool     `json:"rhsm"`
```

biancarosa: This is for image builder's request so depends on how @ldjebran implements there.

biancarosa: I do like with the underscore better for what's worth.

Saurabh-Thakre: yeah for imagebuilder we need to keep base_url as baseurl here for example.

pkg/clients/imagebuilder/client.go
```diff
@@ -47,7 +48,19 @@ type OSTree struct {
...
 // Customizations is made of the packages that are baked into an image
 type Customizations struct {
-	Packages *[]string `json:"packages"`
+	Packages            *[]string     `json:"packages"`
+	PayloadRepositories *[]Repository `json:"payload_repositories,omitempty"`
+}
+
+// Repository is the record of Third Party Repository
+type Repository struct {
+	Baseurl    []string `json:"baseurl,omitempty"`
```
ldjebran, line 57: baseurl is of type "string" and not "[]string".

Saurabh-Thakre: oh yeah, PR was still in draft thats I left couple things for finishing. Now opening the PR for review

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +425,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
 	c.log.Infof("Done with metadata for image")
 	return image, nil
 }
+
+// GetThirdPartyURL finds the url of Third Party Repository using the name
+func (c *Client) GetThirdPartyURL(image *models.Image) ([]Repository, error) {
```
ldjebran, line 430:
```suggestion
func (c *Client) GetThirdPartyRepos(image *models.Image) ([]Repository, error) {
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +425,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	var thirdpartyrepo *models.ThirdPartyRepo
+	account, err := common.GetAccountFromContext(c.ctx)
+	if err != nil {
+		c.log.WithField("error", err).Error("error retrieving account")
+		return nil, errors.New("error retrieving account")
+	}
+	pkgs := make([]Repository, len(image.ThirdPartyRepositories))
```
ldjebran, line 437:
```suggestion
	repos := make([]Repository, len(image.ThirdPartyRepositories))
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +425,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	for i := range image.ThirdPartyRepositories {
+		tprepoURL := db.DB.Model(&models.ThirdPartyRepo{}).Select("url").Where("account = ? and id = ?", account, image.ThirdPartyRepositories[i].ID).Find(&thirdpartyrepo)
+		if tprepoURL.Error != nil {
+			log.Error(tprepoURL.Error)
+		}
+		pkgs[i] = Repository{
```
ldjebran, line 443:
```suggestion
		repos[i] = Repository{
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +425,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+			BaseURL: thirdpartyrepo.URL,
+		}
+	}
+
+	return pkgs, err
```
ldjebran, line 448:
```suggestion
	return repos, err
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +425,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+func (c *Client) GetThirdPartyURL(image *models.Image) ([]Repository, error) {
+	var thirdpartyrepo *models.ThirdPartyRepo
+	account, err := common.GetAccountFromContext(c.ctx)
+	if err != nil {
+		c.log.WithField("error", err).Error("error retrieving account")
+		return nil, errors.New("error retrieving account")
+	}
+	pkgs := make([]Repository, len(image.ThirdPartyRepositories))
+	for i := range image.ThirdPartyRepositories {
+		tprepoURL := db.DB.Model(&models.ThirdPartyRepo{}).Select("url").Where("account = ? and id = ?", account, image.ThirdPartyRepositories[i].ID).Find(&thirdpartyrepo)
```
ldjebran, line 439: why we should do this ? 
we already have the repo url with  image.ThirdPartyRepositories[i].URL why we have to search it again ?

Saurabh-Thakre: With image.ThirdPartyRepositories[i].ID we only have the repository ID which the user will provide, so this query will find the repo URL for that user-provided ID and for the associated account

pkg/clients/imagebuilder/client.go
```diff
@@ -187,11 +200,15 @@ func (c *Client) compose(composeReq *ComposeRequest) (*ComposeResult, error) {
 
 // ComposeCommit composes a Commit on ImageBuilder
 func (c *Client) ComposeCommit(image *models.Image) (*models.Image, error) {
+	payloadURL, err := c.GetThirdPartyRepos(image)
```
ldjebran, line 203:
```suggestion
	payloadRepos, err := c.GetThirdPartyRepos(image)
```

pkg/clients/imagebuilder/client.go
```diff
@@ -187,11 +200,15 @@ func (c *Client) compose(composeReq *ComposeRequest) (*ComposeResult, error) {
...
+	if err != nil {
+		return nil, errors.New("error getting information on third Party repository")
+	}
 	req := &ComposeRequest{
 		Customizations: &Customizations{
-			Packages: image.GetPackagesList(),
+			Packages:            image.GetPackagesList(),
+			PayloadRepositories: &payloadURL,
```
ldjebran, line 210:
```suggestion
			PayloadRepositories: &payloadRepos,
```

pkg/clients/imagebuilder/client.go
Avielyo10: I fixed my suggestion for `baseurl` :)
@biancarosa @Saurabh-Thakre I still think we should stick with CamelCase on edge-api variable names. I don't see where it's related to image builder's request as those are internal used variables.

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
 	c.log.Infof("Done with metadata for image")
 	return image, nil
 }
+
+// GetThirdPartyRepos finds the url of Third Party Repository using the name
+func (c *Client) GetThirdPartyRepos(image *models.Image) ([]Repository, error) {
+	var thirdpartyrepo *models.ThirdPartyRepo
```
ldjebran, line 433: May be it should be 
```go
var thirdpartyrepo models.ThirdPartyRepo
```
Think we should move this to the loop

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	account, err := common.GetAccountFromContext(c.ctx)
+	if err != nil {
+		c.log.WithField("error", err).Error("error retrieving account")
+		return nil, errors.New("error retrieving account")
+	}
+	repos := make([]Repository, len(image.ThirdPartyRepositories))
+	for i := range image.ThirdPartyRepositories {
+		tprepoURL := db.DB.Model(&models.ThirdPartyRepo{}).Select("url").Where("account = ? and id = ?", account, image.ThirdPartyRepositories[i].ID).Find(&thirdpartyrepo)
```
ldjebran, line 441: may be 
```go
result := db.DB.Where("account = ? and id = ?", account, image.ThirdPartyRepositories[i].ID).First(&thirdpartyrepo)
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,25 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+		if tprepoURL.Error != nil {
+			log.Error(tprepoURL.Error)
+		}
```
ldjebran, line 444: Are we sure , we should continue in the loop on first error ?

ldjebran (commented):
Put some comments

Avielyo10:
@Saurabh-Thakre Tests must be added

ldjebran (commented):
Suggest to refactor function GetThirdPartyRepos to make one request to db and not in the loop for each i, something like : 

```go
db.DB.Where("accounts = ?", account).Find(&repos, reposIDS)
```

Avielyo10:
@Saurabh-Thakre needs rebase

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+		thirdpartyrepoID[i] = int(image.ThirdPartyRepositories[i].ID)
+	}
+	var count int64
+	result := db.DB.Where("account = ?", account).Find(&thirdpartyrepo, thirdpartyrepoID).Count(&count)
+	if result.Error != nil {
+		log.Error(result.Error)
+		return nil, result.Error
+	}
+
+	if count != int64(len(thirdpartyrepoID)) {
```
ldjebran, line 453: why we need to test this ?
for me here we are testing SQL and GORM to work properly or you have something other in mind?

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+func (c *Client) GetThirdPartyRepos(image *models.Image) ([]Repository, error) {
+	account, err := common.GetAccountFromContext(c.ctx)
+	if err != nil {
+		c.log.WithField("error", err).Error("error retrieving account")
+		return nil, errors.New("error retrieving account")
+	}
+
+	repos := make([]Repository, len(image.ThirdPartyRepositories))
+	thirdpartyrepo := make([]models.ThirdPartyRepo, len(image.ThirdPartyRepositories))
+	thirdpartyrepoID := make([]int, len(image.ThirdPartyRepositories))
```
ldjebran, line 441: thirdpartyrepoIDS

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+
+	for i := range image.ThirdPartyRepositories {
+		thirdpartyrepoID[i] = int(image.ThirdPartyRepositories[i].ID)
+	}
+	var count int64
+	result := db.DB.Where("account = ?", account).Find(&thirdpartyrepo, thirdpartyrepoID).Count(&count)
```
ldjebran, line 447: We may give a try to
```go
db.Model(&image).Association("ThirdPartyRepositories").Find(&repos)
```
@biancarosa @Avielyo10 what do you think?

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+// GetThirdPartyRepos finds the url of Third Party Repository using the name
+func (c *Client) GetThirdPartyRepos(image *models.Image) ([]Repository, error) {
+	account, err := common.GetAccountFromContext(c.ctx)
+	if err != nil {
+		c.log.WithField("error", err).Error("error retrieving account")
+		return nil, errors.New("error retrieving account")
+	}
+
+	repos := make([]Repository, len(image.ThirdPartyRepositories))
+	thirdpartyrepo := make([]models.ThirdPartyRepo, len(image.ThirdPartyRepositories))
```
ldjebran, line 440: thirdpartyrepos

pkg/clients/imagebuilder/client.go
Saurabh-Thakre: suppose a user has three inputs like {1,2,3] and thirdpartyrepoID 3 does not exist in the database, then it should give error and should not create the image with other two ids.

ldjebran: In this case you just created the image with repo ids {1, 2, 3} the database has a reference integrity check, if 3 does not exist , the image will not be saved.
But this check is simply checking that count returned is the same as the len of the slice also returned. 
If the repo was deleted meanwhile building it should be also deleted from image thirdparty repos collection (because of the reference integrity check) and the last query will be without that repo 3 anyway.

ldjebran (commented):
Missing verifying "account" of the inserted repo into image repos collection.

We have to be sure that the image account and repo account are the same when inserting to image image collection.

pkg/clients/imagebuilder/client.go
ldjebran: eg: We should not be able to delete the repo, if repo is in a collection.

ldjebran: Ok got the purpose, but we need to verify account before landing in image collection as in the review comment

pkg/clients/imagebuilder/client.go
Avielyo10: @ldjebran Much better approach, I like it.

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,37 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+		return nil, errors.New("error retrieving account")
+	}
+
+	repos := make([]Repository, len(image.ThirdPartyRepositories))
+	thirdpartyrepos := make([]models.ThirdPartyRepo, len(image.ThirdPartyRepositories))
+	thirdpartyrepoIDS := make([]int, len(image.ThirdPartyRepositories))
+
+	for i := range image.ThirdPartyRepositories {
+		thirdpartyrepoIDS[i] = int(image.ThirdPartyRepositories[i].ID)
+	}
```
Avielyo10, line 445:
```suggestion
	for repo := range image.ThirdPartyRepositories {
		thirdpartyrepoIDS[repo] = int(image.ThirdPartyRepositories[repo].ID)
	}
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,37 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+
+	if count != int64(len(thirdpartyrepoIDS)) {
+		return nil, errors.New("enter valid third party repository id")
+	}
+	for i := range thirdpartyrepoIDS {
+		repos[i] = Repository{
+			BaseURL: thirdpartyrepos[i].URL,
+		}
+	}
```
Avielyo10, line 460:
```suggestion
	for repoID := range thirdpartyrepoIDS {
		repos[repoID] = Repository{
			BaseURL: thirdpartyrepos[repoID].URL,
		}
	}
```

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,37 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+
+	repos := make([]Repository, len(image.ThirdPartyRepositories))
+	thirdpartyrepos := make([]models.ThirdPartyRepo, len(image.ThirdPartyRepositories))
+	thirdpartyrepoIDS := make([]int, len(image.ThirdPartyRepositories))
+
+	for i := range image.ThirdPartyRepositories {
+		thirdpartyrepoIDS[i] = int(image.ThirdPartyRepositories[i].ID)
+	}
+	var count int64
+	result := db.DB.Where("account = ?", account).Find(&thirdpartyrepos, thirdpartyrepoIDS).Count(&count)
```
Avielyo10, line 447: I do think that the approach suggested by @ldjebran  to use Associations is much more efficient

Saurabh-Thakre: Yes, makes sense to me as well. had a talk with @ldjebran  and now I'm working on using the query and validating the accounts packages in images.go

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	if result.Error != nil {
+		log.Error(result.Error)
+		return nil, result.Error
+	}
+
+	if count != int64(len(thirdpartyrepoIDS)) {
+		return nil, errors.New("enter valid third party repository id")
+	}
+	for repoID := 0; repoID < len(thirdpartyrepos); repoID++ {
```
ldjebran, line 456: This is not repoID just ind 
```suggestion
	for int := 0; ind < len(thirdpartyrepos); ind++ {
```

pkg/services/images.go
```diff
@@ -71,6 +71,23 @@ type ImageService struct {
 	RepoService  RepoServiceInterface
 }
 
+// ValidateAllImagesPackagesAreFromAccount validates the account for Third Party Repositories
+func ValidateAllImagesPackagesAreFromAccount(account string, repos []models.ThirdPartyRepo) (bool, error) {
```
ldjebran, line 75: missing the case when len(repos) == 0  and when account = ""

when account == "" validation is false 
when  len(repos) == 0 validation is true

pkg/clients/imagebuilder/client.go
```diff
@@ -410,3 +427,38 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	result := db.DB.Where("account = ?", account).Find(&thirdpartyrepos, thirdpartyrepoIDS).Count(&count)
+	if result.Error != nil {
+		log.Error(result.Error)
+		return nil, result.Error
+	}
+
+	if count != int64(len(thirdpartyrepoIDS)) {
+		return nil, errors.New("enter valid third party repository id")
+	}
+	for int := 0; ind < len(thirdpartyrepos); ind++ {
```
ldjebran, line 456:
```suggestion
	for repoID := 0; repoID < len(thirdpartyrepos); repoID++ {
```

pkg/clients/imagebuilder/client.go
ldjebran: Just tried that and seems we cannot use this approche as the image is not saved to db yet at this stage

ldjebran (commented):
We should also validate on UpdateImage.

pkg/clients/imagebuilder/client.go
ldjebran: We investigated on this and the image is not saved yet at this stage thus unfortunately we cannot use this method.

PR #553, opened 2022-01-19
Use a method to get image by ostree hash and rollback image

biancarosa:
# Description

Use a method to get image by ostree hash and rollback image. Tests pending for the new method, that's why I left on draft.

I did this by centralizing the way we get images by ostree hash and the rollback image, so that I could test both method separately, and making sure we are doing all the joins to retrieve all the documented nested associations.

It has been discussed that we might want to, in the future, review how our API handles that so that we don't cause performance issues by returning all of these objects. We want the docs to reflect these lazy loading associations though, and need to find a way to make openapi spec take that into consideration.

Fixes #THEEDGE-1507

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


acosferreira (approved):
LGTM

PR #581, opened 2022-01-31
Device groups: Create DB structure to "groups".

ldjebran:
# Description

Fixes # THEEDGE-1590

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


pkg/models/groups.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"errors"
+	"regexp"
+)
+
+// Group is a record of Edge Devices Groups
+// Account is the account associated with the group
+// Type is the group type and must be "static" or "dynamic"
+type Group struct {
```
biancarosa, line 11: Maybe we should call it DeviceGroups?

ldjebran: I was also thinking this way, but there is also devices_groups , so then we will have too much devices groups and will became confusing , 
so we have devices_groups table for the collection 
and if this one will be DeviceGroup  we will have a device_groups 
naming is hard :)

we can rename but we will be confused with device_groups and devices_groups.
what do you think ?

ldjebran: Or rename this to DeviceGroup   and the collection to device_groups_devices , what do you think ?

ldjebran: anyway going to rename to DeviceGroup and the collection to device_groups_devices

seems more logical

biancarosa (approved):
LGTM!

PR #582, opened 2022-01-31
Replace the call to /hosts so we can get the exact device for ID

Avielyo10:
# Description

The following API call: /v1/devices/45 doesn't return error 404 as expected (since there is no device with uuid 45). This is also true for `uuid = 4`  as  well as for other numbers. There are numbers that truly return 404 like `uuid = 1`.

The response has only the image-info attribute while all the other attributes are empty it seems like the image-info is taken from the latest device.

Fixes THEEDGE-1617

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


Avielyo10:
@biancarosa I really need your eyes on it since I don't know how this will take affect on others

biancarosa (approved):
I think this makes sense as it wasn't an optional parameter from the call;

PR #595, opened 2022-02-02
Don't duplicate ImageSets when an Image with the same name attempts to be created twice

maxamillion:

# Description

Removes the ability to create a duplicate an ImageSet.

## Type of change

What is it?

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

- [-] I have made corresponding changes to the documentation


Avielyo10 (approved):
ltgm

pkg/services/images.go
```diff
@@ -75,15 +75,34 @@ type ImageService struct {
...
+		Find(&imageSetExists).
+		Error
+
+	if (err == nil) && imageSetExists {
+		result := db.DB.Where("name = ? AND account = ?", image.Name, account).First(&imageSet)
+		if result.Error != nil {
+			s.log.WithField("error", result.Error.Error()).Error("Error checking for previous image set existence")
+			return result.Error
+		}
+		log.Errorf("ImageSet %s already exists, UpdateImage transaction expected and not CreateImage", image.Name)
```
biancarosa, line 95:
```suggestion
		s.log.WithField("imageSetName", name).Error("ImageSet already exists, UpdateImage transaction expected and not CreateImage", image.Name)
```

biancarosa (changes requested):

pkg/services/images.go
```diff
@@ -75,15 +75,34 @@ type ImageService struct {
...
-	imageSet.Account = account
-	imageSet.Name = image.Name
+
+	err := db.DB.Model(imageSetModel).
+		Select("count(*) > 0").
+		Where("name = ? AND account = ?", image.Name, account).
+		Find(&imageSetExists).
+		Error
+
+	if (err == nil) && imageSetExists {
```
biancarosa, line 89: shouldn't we check if err != nil before that block? if returning an image fails i think we should give an internal error, no?

Avielyo10:
@Mergifyio refresh

pkg/services/images.go
maxamillion: oh, yes ... .yes we should, thanks!

maxamillion:
@Mergifyio refresh

PR #606, opened 2022-02-03
Refactor code to avoid duplicate calls to ReturnDeviceByID on inventoryAPI

biancarosa:
# Description

After @matoval  pointed out that the call to /devices was giving a error 500, we got together and realized that the reason why it was giving an error was because the system_profile was not being returned.

We realized together that the /hosts/<id> endpoint by inventory API doesn't return the system profile and another call is needed for that (/hosts/<id>/system_profile). We started calling /hosts/<id> in #582  and that fixed another bug.

The /devices endpoint  didn't need to call that endpoint - this PR removes the need to call /hosts/<id> on the /devices endpoint.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

- [x]  New and existing unit tests pass locally with my changes

### DIFF
--- a/pkg/services/devices.go
+++ b/pkg/services/devices.go
@@ -235,6 +235,7 @@ func (s *DeviceService) GetDeviceImageInfo(device inventory.Device) (*models.Ima
 	var lastDeployment *inventory.OSTree
 
 	for _, rpmOstree := range device.Ostree.RpmOstreeDeployments {
+		rpmOstree := rpmOstree
 		if rpmOstree.Booted {
 			lastDeployment = &rpmOstree
 			break
--- a/pkg/services/devices.go
+++ b/pkg/services/devices.go
@@ -12,12 +12,17 @@ import (
 
 // DeviceServiceInterface defines the interface to handle the business logic of RHEL for Edge Devices
 type DeviceServiceInterface interface {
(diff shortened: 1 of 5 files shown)


pkg/services/devices.go
```diff
@@ -198,28 +217,34 @@ func GetDiffOnUpdate(oldImg models.Image, newImg models.Image) models.PackageDif
...
+// GetDeviceImageInfo returns the information of a the running image for a device
+func (s *DeviceService) GetDeviceImageInfo(device inventory.Device) (*models.ImageInfo, error) {
+	var ImageInfo models.ImageInfo
+	var currentImage *models.Image
+	var rollback *models.Image
+	var lastDeployment *inventory.OSTree
 
-	for _, rpmOstree := range lastDevice.Ostree.RpmOstreeDeployments {
+	for _, rpmOstree := range device.Ostree.RpmOstreeDeployments {
+		rpmOstree := rpmOstree
```
thom-at-redhat, line 238: Is this line necessary?

thom-at-redhat (approved):

pkg/services/devices.go
biancarosa: Yes, this is needed to avoid implicit memory aliasing and its fixing what gosec complained about in the last build

PR #610, opened 2022-02-04
Revert "Replace the call to /hosts so we can get the exact device for ID"

biancarosa:
Reverts RedHatInsights/edge-api#582


Avielyo10:
Hi @biancarosa, please explain why

PR #620, opened 2022-02-07
Device groups: Add devices.

ldjebran:
# Description

Fixes # THEEDGE-1594)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/devicegroups.go
+++ b/pkg/routes/devicegroups.go
@@ -36,6 +36,7 @@ func MakeDeviceGroupsRouter(sub chi.Router) {
 		r.Get("/", GetDeviceGroupByID)
 		r.Put("/", UpdateDeviceGroup)
 		r.Delete("/", DeleteDeviceGroupByID)
+		r.Post("/devices", AddDeviceGroupDevices)
 	})
 }
 
@@ -52,8 +53,8 @@ func DeviceGroupCtx(next http.Handler) http.Handler {
 				err := errors.NewBadRequest(err.Error())
 				w.WriteHeader(err.GetStatus())
 				if err := json.NewEncoder(w).Encode(&err); err != nil {
-					services := dependencies.ServicesFromContext(r.Context())
-					services.Log.WithField("error", err.Error()).Error("Error while trying to encode")
+					ctxServices := dependencies.ServicesFromContext(r.Context())
+					ctxServices.Log.WithField("error", err.Error()).Error("Error while trying to encode")
 				}
 				return
 			}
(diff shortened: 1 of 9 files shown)


pkg/services/devicegroups.go
```diff
@@ -108,3 +111,59 @@ func (s *DeviceGroupsService) GetDeviceGroupByID(ID string) (*models.DeviceGroup
 	}
 	return &deviceGroup, nil
 }
+
+const (
```
acosferreira, line 115: What do you think to move it for the Error class?

pkg/routes/devicegroups.go
```diff
@@ -22,6 +22,19 @@ import (
 type deviceGroupTypeKey int
 
 const deviceGroupKey deviceGroupTypeKey = iota
+const contextDeviceGroupNotFound = "context device group not found"
+
+func setContextDeviceGroup(ctx context.Context, deviceGroup *models.DeviceGroup) context.Context {
+	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
+}
+
+func getContextDeviceGroup(ctx context.Context) (*models.DeviceGroup, error) {
```
acosferreira, line 31: I think we are doing this duplicated, https://github.com/RedHatInsights/edge-api/pull/620/files#diff-48f6502941a54bb08d271267b7002cd524e07fade2ffb11ba7437baca89f781dR253
Maybe can just get it once. 
What do you think?

pkg/routes/devicegroups.go
```diff
@@ -22,6 +22,19 @@ import (
 type deviceGroupTypeKey int
 
 const deviceGroupKey deviceGroupTypeKey = iota
+const contextDeviceGroupNotFound = "context device group not found"
+
+func setContextDeviceGroup(ctx context.Context, deviceGroup *models.DeviceGroup) context.Context {
+	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
```
acosferreira, line 28: we are already setting the context in the route, should we set it again?
https://github.com/RedHatInsights/edge-api/pull/620/files#diff-48f6502941a54bb08d271267b7002cd524e07fade2ffb11ba7437baca89f781dR44

pkg/routes/devicegroups.go
ldjebran: cannot get it where we are duplicating this  ?

acosferreira: https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/devicegroups.go#L224

pkg/routes/devicegroups.go
acosferreira: https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/devicegroups.go#L31

pkg/routes/devicegroups.go
ldjebran: Think the name getContextDeviceGoup is more adequate to the usage what do you think ?

acosferreira: I think we can change the name, but avoid the duplicity ;)

pkg/routes/devicegroups.go
acosferreira: @biancarosa any thougth ?

biancarosa: looks like the call was just encapsulated to be called on a test, nothing wrong with it

acosferreira (approved):
lgtm

PR #647, opened 2022-02-09
Migrate device account

acosferreira:
# Description

Setting a default account to the existent devices, cClowdapp Should be rollback 

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

- [ X] I run `go vet ./...` to check that my code is free of common Go style mistakes


cmd/db/set_account_on_device.go
```diff
@@ -0,0 +1,39 @@
...
+		"DefaultOSTreeRef":         cfg.DefaultOSTreeRef,
+		"InventoryURL":             cfg.InventoryConfig.URL,
+		"PlaybookDispatcherConfig": cfg.PlaybookDispatcherConfig.URL,
+		"TemplatesPath":            cfg.TemplatesPath,
+		"DatabaseType":             cfg.Database.Type,
+		"DatabaseName":             cfg.Database.Name,
+	}).Info("Configuration Values:")
+	db.InitDB()
+
+	log.Info("Setting default account to existent devices...")
```
ldjebran, line 36: I was thinking that we are going to query inventory for the devices using their uuid and get the account info for each device and save it locally

Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN go build -tags=fdo -o /go/bin/edge-api
 # Build the migration binary.
 RUN go build -o /go/bin/edge-api-migrate cmd/migrate/main.go
 RUN go build -o /go/bin/edge-api-wipe cmd/db/wipe.go
+RUN go build -o /go/bin/edge-api-wipe cmd/db/set_account_on_device.go
```
ldjebran, line 26: may be 
```go
RUN go build -o /go/bin/edge-api-migrate-device cmd/db/set_account_on_device.go
```

cmd/db/set_account_on_device.go
acosferreira: I discussed in a thread, as stage data, will be enough just set a default value to the old devices ;)

acosferreira: do you have any concern on this action?

ldjebran: I would prefer if we add "WHERE account IS NULL"

ldjebran (approved):

PR #668, opened 2022-02-14
Delete DeviceGroup

Avielyo10:
# Description

Deleting a DeviceGroup by its ID, since there are no orphan devices - delete those too.

- Adding the service
- Adding the route
- Refactor testing 

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:


pkg/models/devicegroups_test.go
```diff
@@ -88,3 +89,58 @@ func TestGroupCreateUpdateConstraint(t *testing.T) {
...
+	// Create the DeviceGroup
+	err := db.DB.Create(&deviceGroup).Error
+	if err != nil {
+		t.Error("Error saving device group to DB")
+	}
+	// Check all in DB
+	var deviecsFromDB []Device
+	err = db.DB.Where("account = ?", account).Find(&deviecsFromDB).Error
+	if err != nil {
+		t.Error("Error retrieving devices from DB")
```
ldjebran, line 123: same here with t.Fatal

pkg/models/devicegroups_test.go
```diff
@@ -88,3 +89,58 @@ func TestGroupCreateUpdateConstraint(t *testing.T) {
...
+	deviceGroup := &DeviceGroup{
+		Name:    deviceGroupName,
+		Type:    DeviceGroupTypeDefault,
+		Account: account,
+		Devices: devices,
+	}
+	// Create the DeviceGroup
+	err := db.DB.Create(&deviceGroup).Error
+	if err != nil {
+		t.Error("Error saving device group to DB")
```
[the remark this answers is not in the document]
Avielyo10: Why so?

pkg/models/devicegroups_test.go
Avielyo10: why so?

pkg/models/devicegroups_test.go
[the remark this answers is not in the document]
ldjebran: it make no sense now that you have changed all the tests :)

pkg/routes/devicegroups.go
```diff
@@ -272,7 +272,29 @@ func UpdateDeviceGroup(w http.ResponseWriter, r *http.Request) {
 
 // DeleteDeviceGroupByID deletes an existing device group
 func DeleteDeviceGroupByID(w http.ResponseWriter, r *http.Request) {
-
+	ctxServices := dependencies.ServicesFromContext(r.Context())
+	deviceGroup := getContextDeviceGroup(w, r)
+	if deviceGroup == nil {
+		return // error handled by getContextDeviceGroup already
+	}
+	ctxServices.Log = ctxServices.Log.WithField("device_group_id", deviceGroup.ID)
```
ldjebran, line 280: I found very weird to re-allocataire the services Log, why we are doing so (as found this in many places, this may be a source of bugs ).
what do you think?

pkg/routes/devicegroups_test.go
```diff
@@ -321,3 +325,73 @@ func TestUpdateDeviceGroup(t *testing.T) {
...
+				Expect(err).To(BeNil())
+				Expect(deviceGroupBytes).NotTo(BeNil())
+			})
+		})
+
+		When("all is valid", func() {
+			url := fmt.Sprintf("/%d", deviceGroup.ID)
+			req, err := http.NewRequest(http.MethodDelete, url, bytes.NewBuffer(deviceGroupBytes))
+			It("should create new request", func() {
+				Expect(err).To(BeNil())
```
ldjebran, line 371: The operation we test should be in the test it self.
```go
It("should create new request", func() {
    req, err := http.NewRequest(http.MethodDelete, url, bytes.NewBuffer(deviceGroupBytes))
    Expect(err).To(BeNil())
})
```

ldjebran: not a blocker

pkg/routes/devicegroups_test.go
```diff
@@ -321,3 +325,73 @@ func TestUpdateDeviceGroup(t *testing.T) {
...
+				DeviceGroupsService: mockDeviceGroupsService,
+				Log:                 log.NewEntry(log.StandardLogger()),
+			})
+			req = req.WithContext(ctx)
+			rr := httptest.NewRecorder()
+			handler := http.HandlerFunc(DeleteDeviceGroupByID)
+
+			handler.ServeHTTP(rr, req)
+			It("should return status code 200", func() {
+				Expect(rr.Code).To(Equal(http.StatusOK))
```
ldjebran, line 393: same here the testing subject should be in the test (not a blocker)

pkg/models/devicegroups_test.go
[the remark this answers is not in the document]
Avielyo10: This is testing only the `BeforeDelete` functionality, not the DB

pkg/services/devicegroups.go
```diff
@@ -76,6 +77,24 @@ func (s *DeviceGroupsService) GetDeviceGroupsCount(account string, tx *gorm.DB)
 	return count, nil
 }
 
+// DeleteDeviceGroupByID deletes the device group by ID from the database
+func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
+	s.log = s.log.WithField("device_group_id", ID)
```
ldjebran, line 82: same , where we are re-allocating the service log ?

ldjebran: Note for the need of our function we can create a logger like :
```go
sLog := s.log.WithField("device_group_id", ID)
```

and use it in the function and do not reallocate the service log

pkg/routes/devicegroups.go
Avielyo10: This actually replaces the memory address to add functionality and this works only for the function scope so there is no way this can cause an issue on a different method.

pkg/services/devicegroups.go
```diff
@@ -76,6 +77,24 @@ func (s *DeviceGroupsService) GetDeviceGroupsCount(account string, tx *gorm.DB)
...
+	s.log.Info("Deleting device group")
+	deviceGroup, err := s.GetDeviceGroupByID(ID) // get the device group
+	if err != nil {
+		s.log.WithField("error", err.Error()).Error("Error getting device group")
+		return err
+	}
+	// delete the device group
+	result := db.DB.Delete(&deviceGroup)
```
ldjebran, line 90:
```go
if result := db.DB.Delete(&deviceGroup); result.Error != nil {
        s.log.WithField("error", result.Error.Error()).Error("Error deleting device group")
	return result.Error
}
```

pkg/routes/devicegroups.go
ldjebran: I am not sure about that as ctxServices is a pointer.

pkg/models/devicegroups_test.go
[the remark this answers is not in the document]
ldjebran: should work with resole, and we are upgrading test framework to ginkgo

PR #682, opened 2022-02-16
Removed unneeded verbose logging when uploading images.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


pkg/services/files/uploader.go
```diff
@@ -204,8 +202,7 @@ func (u *S3Uploader) UploadFile(fname string, uploadPath string) (string, error)
 		ACL:    aws.String("public-read"),
 	})
 
-	u.log.WithField("result", result).Info("Finished upload to AWS S3")
-	if err != nil {
+	if err != nil || result == nil {
```
ldjebran, line 205: if err == nil and resilut == nil we stil  enter this block and in the block we have err.Error() that will produce a panic

ldjebran:
```suggestion
	if err != nil {
```

ldjebran, line 205:
```suggestion
	if err != nil {
```

ldjebran:
```suggestion
	if err != nil {
```

AaronH88: I see what you mean.
Dropping result altogether.

PR #706, opened 2022-02-23
Main refactoring

Avielyo10:
# Description

* Use some functions for maintainability
* Remove unneeded channels & goroutines
* Register consumer closing in a graceful manner.
* Anonymouse consumers.
* Handle nil kafka config in the the service creation.
* Close() is a part of the ConsumerService

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

# Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	redoc "github.com/go-openapi/runtime/middleware"
 	"github.com/redhatinsights/edge-api/config"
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -40,33 +41,28 @@ func initDependencies() {
...
-		"DefaultOSTreeRef":         cfg.DefaultOSTreeRef,
-		"InventoryURL":             cfg.InventoryConfig.URL,
-		"PlaybookDispatcherConfig": cfg.PlaybookDispatcherConfig.URL,
-		"TemplatesPath":            cfg.TemplatesPath,
-		"DatabaseType":             cfg.Database.Type,
-		"DatabaseName":             cfg.Database.Name,
-		"IsKafkaEnabled":           cfg.KafkaConfig != nil,
-		"FDOHostURL":               cfg.FDO.URL,
-	}).Info("Configuration Values")
+func serveMetrics(port int) *http.Server {
```
AaronH88, line 44: Moving this to its own file might improve readability. Something like a `metrics.go`

AaronH88: I really like this idea, good job!

main.go
```diff
@@ -76,21 +72,21 @@ func main() {
 	)
 
 	// Unauthenticated routes
-	r.Get("/", routes.StatusOK)
-	r.Get("/api/edge/v1/openapi.json", func(w http.ResponseWriter, r *http.Request) {
+	route.Get("/", routes.StatusOK)
```
AaronH88, line 75: Love these more explicit var names

main.go
```diff
@@ -99,57 +95,88 @@ func main() {
...
-		sigint := make(chan os.Signal, 1)
-		signal.Notify(sigint, os.Interrupt, syscall.SIGTERM)
-		<-sigint
-		log.Info("Shutting down gracefully...")
-		if err := srv.Shutdown(context.Background()); err != nil {
-			l.LogErrorAndPanic("HTTP server (web port) shutdown failed", err)
+	server.RegisterOnShutdown(func() {
+		for _, consumer := range consumers {
+			if consumer != nil {
+				consumer.Close()
```
AaronH88, line 111: Very nice

main.go
```diff
@@ -99,57 +95,88 @@ func main() {
...
 	}
+	webServer := serveWeb(cfg, consumers)
+	metricsServer := serveMetrics(cfg.MetricsPort)
 
-	<-gracefulStop
+	if cfg.KafkaConfig != nil {
+		log.Info("Starting Kafka Consumers")
+		for _, consumer := range consumers {
+			if consumer != nil {
+				go consumer.Start()
```
AaronH88, line 173: Very very nice!

main.go
```diff
@@ -99,57 +95,88 @@ func main() {
...
+			}
+		}
+	}
+	<-sigint
```
AaronH88, line 177: Humm, can we think of a more verbose name here?

pkg/services/consumers.go
```diff
@@ -64,25 +62,28 @@ type PlatformInsightsCreateEventPayload struct {
...
-		topic:         topic,
-	}
-	if topic == "platform.playbook-dispatcher.runs" {
-		s.consumer = s.ConsumePlaybookDispatcherRuns
-	} else if s.topic == "platform.inventory.events" {
-		s.consumer = s.ConsumeInventoryCreateEvents
-	} else {
-		log.Errorf("No consumer for topic: %s", topic)
-		return nil
+	if config != nil {
```
AaronH88, line 65: If we change the condition here we can improve readability a bit, and also remove some nested `if`'s

e.g. `if config == nil { return nil }`

main.go
AaronH88: Also, where was this `sigint` channel created?

main.go
Avielyo10: You mean separate the metrics & the web servers?

main.go
Avielyo10: like `interruptSignal`?
Created at the beginning of `main()`

pkg/services/consumers.go
Avielyo10: Like that!

main.go
AaronH88: Exactly, I personally would expect to find this either under `pkg/routes/` or `pkg/services/` 
Just a thought :smile:

main.go
AaronH88: Great choice

main.go
Avielyo10: Can this take place in a new PR?

ldjebran (commented):
It looks good, waiting @AaronH88 response

loadtheaccumulator (changes requested):
Do all of these commits have to go in at the same time, or can they be broken into separate PRs. We've been running into breakage issues with large refactors.

I think we might need to sync more on refactor plans versus new feature now that we're in squads and might be touching the same code.

Avielyo10:
@loadtheaccumulator, actually they are all connected in some way.

This is not a feature, only improve readability & avoid goroutines leak on main (also implement a real graceful shutdown)

Avielyo10:
@mergify rebase

main.go
AaronH88: Yes it can, Ok move this in a new PR.

AaronH88:
Looks good to me, ok to merge as well as have that chat Jonathan is pointing out

Avielyo10:
@loadtheaccumulator ?

PR #723, opened 2022-03-01
Fix image-sets filtering

matoval:
# Description

Added check if filter has multiple params with the same key, if so the DB query has an OR to get both params being filtered 

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

- [ ] I run `go fmt ./...` to check that my code is properly formatted
- [ ] I run `go vet ./...` to check that my code is free of common Go style mistakes


Avielyo10:
Thanks @matoval, can you add a test that proves your fix?

matoval:
> Thanks @matoval, can you add a test that proves your fix?

I added an additional test to test multiple status filters to test the code I added.

pkg/routes/common/filters_test.go
```diff
@@ -79,6 +79,32 @@ func TestContainFilterHandler(t *testing.T) {
...
+	result.Find(&images)
+	hasBothStatus := 0
+	for _, image := range images {
+		if image.Status == "SUCCESS" {
+			hasBothStatus++
+		} else if image.Status == "ERROR" {
+			hasBothStatus++
+		}
+	}
+	fmt.Println(hasBothStatus, len(images))
```
Avielyo10, line 102: Please use `t.Log` or `t.Logf` here instead of `fmt`.
This will allow the text to be printed only if the test fails or the -test.v flag is set.

matoval: I should've removed this, it was a debugging log

pkg/routes/common/filters.go
```diff
@@ -20,10 +20,17 @@ type Filter struct {
...
+		if multipleStatusQuery := r.URL.Query()[filter.QueryParam]; len(multipleStatusQuery) > 1 {
+			for i, q := range multipleStatusQuery {
+				if i == 0 {
+					tx = tx.Where(fmt.Sprintf("%s LIKE ?", filter.DBField), q)
+				} else {
+					tx = tx.Or(fmt.Sprintf("%s LIKE ?", filter.DBField), q)
+				}
+			}
+		} else if val := r.URL.Query().Get(filter.QueryParam); val != "" {
+			tx = tx.Where(fmt.Sprintf("%s LIKE ?", filter.DBField), "%"+val+"%")
```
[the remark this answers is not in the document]
matoval: The else is just to return tx, I think it, makes sense to have an else if here to make sure that val is returning a value

pkg/routes/common/filters.go
```diff
@@ -20,10 +20,17 @@ type Filter struct {
 
 // ContainFilterHandler handles sub string values
 func ContainFilterHandler(filter *Filter) FilterFunc {
-	sqlQuery := fmt.Sprintf("%s LIKE ?", filter.DBField)
 	return FilterFunc(func(r *http.Request, tx *gorm.DB) *gorm.DB {
-		if val := r.URL.Query().Get(filter.QueryParam); val != "" {
-			tx = tx.Where(sqlQuery, "%"+val+"%")
+		if multipleStatusQuery := r.URL.Query()[filter.QueryParam]; len(multipleStatusQuery) > 1 {
+			for i, q := range multipleStatusQuery {
```
Avielyo10, line 25: we rather use `query` than just `q`

pkg/routes/common/filters.go
```diff
@@ -20,10 +20,17 @@ type Filter struct {
...
+				if i == 0 {
+					tx = tx.Where("? LIKE ?", filter.DBField, q)
+				} else {
+					tx = tx.Or("? LIKE ?", filter.DBField, q)
+				}
+			}
+		} else if val := r.URL.Query().Get(filter.QueryParam); val != "" {
```
Avielyo10, line 32: so if there is no `else` WDYT about setting 2 `if`s?or even a case/switch

Avielyo10 (approved):
lgtm

PR #782, opened 2022-03-14
Added more data to Device struct, inclu…

AaronH88:
…ding an ImageID.

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

- [ ] I run `go fmt ./...` to check that my code is properly formatted
- [ ] I run `go vet ./...` to check that my code is free of common Go style mistakes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,12 @@
 module github.com/redhatinsights/edge-api
 
 require (
-	github.com/aws/aws-sdk-go v1.43.16
+	github.com/aws/aws-sdk-go v1.43.18
 	github.com/bxcodec/faker/v3 v3.8.0
 	github.com/cavaliercoder/grab v2.0.0+incompatible
 	github.com/confluentinc/confluent-kafka-go v1.8.2 // indirect
 	github.com/fedora-iot/fido-device-onboard-rs/libfdo-data-go v0.0.0-20211210154920-5c241beb5c4e
-	github.com/getkin/kin-openapi v0.91.0
+	github.com/getkin/kin-openapi v0.92.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-chi/chi v4.1.2+incompatible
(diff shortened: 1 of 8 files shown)


pkg/models/devices.go
```diff
@@ -43,4 +43,7 @@ type Device struct {
 	LastSeen      EdgeAPITime `json:"LastSeen"`
 	CurrentHash   string      `json:"CurrentHash,omitempty"`
 	Account       string      `json:"Account"`
+	DeviceName    string      `json:"DeviceName"`
```
acosferreira, line 46: why we need the DeviceName if we already have the name?

AaronH88: Good spot, we dont need it. removing it now

acosferreira:
1. Should we use the currentHash to store the rpm_ostree_deployments.checksum or we'll go directly create the image relationship?
2. Will we use the booted to store which info? on inventory level this indicates the which image is applyed, in our scenario shouldn't be always true? 
3. On UpdateAvailable, should this be filled up by assync process?

AaronH88:
> 1. Should we use the currentHash to store the rpm_ostree_deployments.checksum or we'll go directly create the image relationship?
> 
>     2. Will we use the booted to store which info? on inventory level this indicates the which image is applyed, in our scenario shouldn't be always true?
> 
>     3. On UpdateAvailable, should this be filled up by assync process?

1. I never looked at that relationship yet. Happy to take direction from you if you think either is a good idea :smile: 

2. This was copy and pasted from the `EdgeDevices` struct. I think your correct, we can remove this var.

3. I was thinking something simpler. When a new Image is created then all devices that use the old image get this bool set to `true`. We could do this async, but it would be more of an implementation detail and not really affect the flow.

acosferreira (approved):
lgtm

PR #832, opened 2022-03-20
THEEDGE-1940 test fix for init container error versus interrupt status

loadtheaccumulator:
* two issues:
* 	1. problems if only one pod of many restarts
* 	2. they restart in a staggered manner...
*		so this steps on status set to INTERRUPTED

# Description

The init container starts in each pod so setting BUILDING to ERROR at start
  can error out valid builds running on other pods.
Second issue is related to when all pods are restarting gracefully, they do not
  stop and start at the same time. There is a cascade, so BUILDING status can
  be wiped before being set to INTERRUPTED.

Fixes # (issue) THEEDGE-1940

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/cmd/migrate/main.go
+++ b/cmd/migrate/main.go
@@ -33,40 +33,43 @@ func main() {
 	}).Info("Configuration Values:")
 	db.InitDB()
 
-	// If there any image builds in progress, in the current architecture, we need to set them as errors because this is a brand new deployment
-	var images []models.Image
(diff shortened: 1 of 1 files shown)


loadtheaccumulator:
Testing to see if this resolves the timing issue with BUILDING not being set to INTERRUPTED on graceful shutdown.

PR #883, opened 2022-03-30
Theedge 1950 refactoring

acosferreira:
# Description

Add imageInfo from device to a device group

Fixes # (THEEDGE-1949 & THEEDGE-1950)

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/devicegroups.go
+++ b/pkg/services/devicegroups.go
@@ -122,44 +122,55 @@ func (s *DeviceGroupsService) GetDeviceGroups(account string, limit int, offset 
 
 	for _, group := range deviceGroups {
 
-		listImageInfo := []models.DeviceImageInfo{}
+		var listImageInfo *[]models.DeviceImageInfo
+		var err error
 		setOfImages := make(map[int]int)
 		//Getting imageId from devices under the group
 		for _, device := range group.Devices {
 			setOfImages[int(device.ImageID)] = int(device.ImageID)
 		}
 		//Getting the imageInfo and calculate the update available based on imageSet
-		for _, imageID := range setOfImages {
-			var updAvailable bool
-			var deviceImage models.Image
-			var deviceImageSet models.ImageSet
-			var CommitID uint
-			if result := db.DB.Where(models.Image{Account: account}).
-				First(&deviceImage, imageID); result.Error != nil {
-				return nil, result.Error
-			}
-			if result := db.DB.Where(models.ImageSet{Account: account}).Preload("Images").
-				First(&deviceImageSet, deviceImage.ImageSetID).Order("ID desc"); result.Error != nil {
-				return nil, result.Error
-			}
-
-			latestImage := &deviceImageSet.Images[len(deviceImageSet.Images)-1]
-			latestImageID := latestImage.ID
-
-			if int(latestImageID) > imageID {
-				updAvailable = true
-				CommitID = deviceImageSet.Images[len(deviceImageSet.Images)-1].CommitID
-			}
-
-			listImageInfo = append(listImageInfo, models.DeviceImageInfo{Name: deviceImage.Name,
-				UpdateAvailable: updAvailable,
-				CommitID:        CommitID})
-
+		listImageInfo, err = GetDeviceImageInfo(setOfImages, account)
+		if err != nil {
+			s.log.WithField("error", err.Error()).Error("Error getting device image info")
+			return nil, res.Error
 		}
 		deviceGroupListDetail = append(deviceGroupListDetail,
-			models.DeviceGroupListDetail{DeviceGroup: group, DeviceImageInfo: &listImageInfo})
+			models.DeviceGroupListDetail{DeviceGroup: group, DeviceImageInfo: listImageInfo})
 	}
 	return &deviceGroupListDetail, nil
+}
+
+func GetDeviceImageInfo(images map[int]int, account string) (*[]models.DeviceImageInfo, error) {
+	listImageInfo := []models.DeviceImageInfo{}
+	for _, imageID := range images {
+		var updAvailable bool
+		var deviceImage models.Image
+		var deviceImageSet models.ImageSet
+		var CommitID uint
+		if result := db.DB.Where(models.Image{Account: account}).
+			First(&deviceImage, imageID); result.Error != nil {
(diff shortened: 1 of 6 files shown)


pkg/services/devicegroups.go
```diff
@@ -102,22 +102,76 @@ func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
...
+		deviceGroupListDetail = append(deviceGroupListDetail,
+			models.DeviceGroupListDetail{DeviceGroup: group, DeviceImageInfo: listImageInfo})
+	}
+	return &deviceGroupListDetail, nil
+}
+
+// GetDeviceImageInfo returns the image related to the groups
+func GetDeviceImageInfo(images map[int]int, account string) (*[]models.DeviceImageInfo, error) {
+	listImageInfo := []models.DeviceImageInfo{}
+	for _, imageID := range images {
```
AaronH88, line 147: Really like this function, its nice and clean

pkg/services/devicegroups.go
```diff
@@ -102,22 +102,76 @@ func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
...
+		if result := db.DB.Where(models.ImageSet{Account: account}).Preload("Images").
+			First(&deviceImageSet, deviceImage.ImageSetID).Order("ID desc"); result.Error != nil {
+			return nil, result.Error
+		}
+
+		latestImage := &deviceImageSet.Images[len(deviceImageSet.Images)-1]
+		latestImageID := latestImage.ID
+
+		if int(latestImageID) > imageID {
+			updAvailable = true
```
AaronH88, line 165: If we are using images like this to discover if a device has an update, then do we need the `UpdateAvailable` flag in devices struct at all?

pkg/services/devicegroups.go
```diff
@@ -102,22 +102,76 @@ func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
...
+		Preload("Devices").
+		Find(&deviceGroups)
 
 	if res.Error != nil {
 		s.log.WithField("error", res.Error.Error()).Error("Error getting device groups")
 		return nil, res.Error
 	}
 
-	return &deviceGroups, nil
+	for _, group := range deviceGroups {
```
AaronH88, line 123: I like the use of the set in these loops, really nice, 
Some questions though.

Can a device be in more than one group?
Does a group have more than one Image?
Can a customer have more than one group?


I think the answer to these 3 is yes, but want to double check.
The reason I am asking is that, as we build all the data we need for one group, a lot of this data may need to be re-discovered for the next group. So instead of having these nested `for` loops, instead in one loop we discover all Devices for all groups and add it to a `set`, then in a following loop discover all Images for all Devices and add to a `set` (like you have done below) and finally then we call your new `GetDeviceImageInfo` function. 

Does this make sense? I believe that any nested loops in data look up like this result in the same data being retrieved more than once

pkg/services/devicegroups.go
acosferreira: Good catch

pkg/services/devicegroups.go
acosferreira: I like that idea! build the data for the devices and aggregate it to dont need to get it again when the device belong to more than 1 group! Let me change it ;)

pkg/services/devicegroups.go
acosferreira: @AaronH88 Can we have the next commit to update on device level as we have the udate_available?
If yes, I'll be able to move all this code, once I'll have the value to send to the FE.

pkg/services/devicegroups.go
acosferreira: @AaronH88 can you take a look if it's in the right way now?

pkg/services/devicegroups.go
```diff
@@ -102,22 +102,99 @@ func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
...
 
-	return &deviceGroups, nil
+	//Getting all devices for all groups
+	setOfDevices := make(map[int][]models.Device)
+	for _, group := range deviceGroups {
+		setOfDevices[int(group.ID)] = group.Devices
+	}
+
+	//Getting all images for all devices
+	setOfImages := make(map[int]int)
```
AaronH88, line 129: So, instead of `setOfImages := make(map[int]int)` consider `setOfImages := make(map[int]*DeviceImageInfo)`

The set part is guaranteed as `image.ID` is the `key` in this map, 
And if we have `*DeviceImageInfo` as the value, and then fill in this value in the `GetDeviceImageInfo` function, then the `//Concat info` work you do below will be a lot easier

pkg/services/devicegroups.go
```diff
@@ -102,22 +102,99 @@ func (s *DeviceGroupsService) DeleteDeviceGroupByID(ID string) error {
...
+		var updAvailable bool
+		var deviceImage models.Image
+		var deviceImageSet models.ImageSet
+		var CommitID uint
+		if result := db.DB.Where(models.Image{Account: account}).
+			First(&deviceImage, imageID); result.Error != nil {
+			return nil, result.Error
+		}
+		if result := db.DB.Where(models.ImageSet{Account: account}).Preload("Images").
+			First(&deviceImageSet, deviceImage.ImageSetID).Order("ID desc"); result.Error != nil {
```
ldjebran, line 180: Think the Order should be done not by id DESC  but by created_at DESC

ldjebran (commented):
please take in consideration that each device has an image and update availability now 

and with this PR https://github.com/RedHatInsights/edge-api/pull/886 they will be also updated

pkg/services/devicegroups.go
acosferreira: The loop on imageSet can be by ID because we dont really update an image, we create a new version that will be the update from the previous one.

pkg/services/devicegroups.go
acosferreira: @AaronH88 let me know if it's better now ;) 
Also I've change  GetDeviceImageInfo to receive the map of models.Device to dont need to query to get the update information

PR #886, opened 2022-03-30
devices: Update db devices when creating an image (update) or when upgrading a device.

ldjebran:
# Description
We need to update devices update availability when an image created (for all the devices that use that image set). 
and When a device is upgraded we need to update the device image and update availability.

Fixes # THEEDGE-1913

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/errors.go
+++ b/pkg/services/errors.go
@@ -119,3 +119,10 @@ type DeviceHasImageUndefined struct{}
 func (e *DeviceHasImageUndefined) Error() string {
 	return "device has image undefined"
 }
+
+// UpdateTransactionHasCommitUndefined indicate that the transaction has no commit defined
+type UpdateTransactionHasCommitUndefined struct{}
+
+func (e *UpdateTransactionHasCommitUndefined) Error() string {
+	return "update transaction has not no defined commit"
+}
--- a/pkg/services/errors.go
+++ b/pkg/services/errors.go
@@ -1,5 +1,7 @@
 package services
 
+import "errors"
+
 // DeviceNotFoundError indicates the device was not found
 type DeviceNotFoundError struct{}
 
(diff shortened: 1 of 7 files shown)


pkg/services/updates.go
```diff
@@ -472,3 +476,60 @@ func (s *UpdateService) SendDeviceNotification(i *models.UpdateTransaction) (Ima
 	}
 	return notify, nil
 }
+
+// UpdateDevicesFromUpdateTransaction update device with new image and update availability
+func (s *UpdateService) UpdateDevicesFromUpdateTransaction(update models.UpdateTransaction) error {
```
AaronH88, line 481: Overall looks great!
This function name is a little confusing, do you have any alternatives?

Perhaps `SetUpdateFlagOnDevices` , or something else :smile:

ldjebran: we are updating only update availability , but also images, any way this must a general function to update a device from update transaction , really nothing  coming in mind than the direct function goal, what is "update the devices from update transaction" :)

Any other suggestion ?

AaronH88: Yeah, I dont have any clearer ideas :smile: 

This name will be OK

pkg/services/errors.go
```diff
@@ -119,3 +119,10 @@ type DeviceHasImageUndefined struct{}
 func (e *DeviceHasImageUndefined) Error() string {
 	return "device has image undefined"
 }
+
+// UpdateTransactionHasCommitUndefined indicate that the transaction has no commit defined
+type UpdateTransactionHasCommitUndefined struct{}
+
+func (e *UpdateTransactionHasCommitUndefined) Error() string {
+	return "update transaction has not no defined commit"
+}
```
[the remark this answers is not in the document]
ldjebran: @Avielyo10 that error is defined the same way all the module errors in that file,

pkg/services/updates.go
ldjebran: @Avielyo10 please propose a name :)

pkg/services/errors.go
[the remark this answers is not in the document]
ldjebran: make changes as per your recommendation.

PR #935, opened 2022-04-11
fixed /updates post response

tomershinhar:
# Description

the response for the post request for the /updates endpoint was changed to be an array of v1.UpdateTransaction
the spec file needs to be updated accordingly


ldjebran (approved):
LGMT

PR #962, opened 2022-04-19
Added UpdateTransaction status to the /devicesview endpoint.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/devices.go
+++ b/pkg/services/devices.go
@@ -617,7 +617,7 @@ func (s *DeviceService) GetDevicesView(limit int, offset int, tx *gorm.DB) (*mod
 	}
 
 	var storedDevices []models.Device
-	if res := tx.Limit(limit).Offset(offset).Where("account = ? AND image_id IS NOT NULL", account).Find(&storedDevices); res.Error != nil {
+	if res := tx.Limit(limit).Offset(offset).Where("account = ? AND image_id IS NOT NULL", account).Preload("UpdateTransaction").Find(&storedDevices); res.Error != nil {
 		return nil, res.Error
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/services/devices.go
```diff
@@ -629,8 +629,21 @@ func (s *DeviceService) GetDevicesView(limit int, offset int, tx *gorm.DB) (*mod
...
+			status = (*devices.UpdateTransaction)[len(*devices.UpdateTransaction)-1].Status
+			if status == models.UpdateStatusBuilding {
+				status = models.DeviceViewStatusUpdating
+			}
+			if status == models.UpdateStatusError {
+				status = models.DeviceViewStatusRunning
+			}
+			if status == models.UpdateStatusCreated {
+				status = models.DeviceViewStatusRunning
+			}
```
de1987, line 643: I think we could change `switch case` or `if else` here.
WDYT?

de1987: another point, I think `Running` should be status by default, the `Running` status is not related to update, but also related to that the device is running, or at least it should.
Maybe Andy could confirm that.

AaronH88: I agree, `Running` is the default. 
I however I think this should be done in the UI because, their is a field `UpdateAvailable` that needs to be checked before the device status is returned to the User.

Hence I have created 3 statuses here, and the `UpdateAvailable` flag :smile: 

I will add the `if else` now!

pkg/services/devices.go
```diff
@@ -629,8 +629,17 @@ func (s *DeviceService) GetDevicesView(limit int, offset int, tx *gorm.DB) (*mod
...
 	for _, devices := range storedDevices {
+		var status = ""
+		if devices.UpdateTransaction != nil && len(*devices.UpdateTransaction) > 0 {
+			status = (*devices.UpdateTransaction)[len(*devices.UpdateTransaction)-1].Status
+			if status == models.UpdateStatusBuilding {
+				status = models.DeviceViewStatusUpdating
+			} else {
+				status = models.DeviceViewStatusRunning
+			}
+		}
```
de1987, line 640:
```suggestion
		var status = models.DeviceViewStatusRunning
		if devices.UpdateTransaction != nil && len(*devices.UpdateTransaction) > 0 {
			status = (*devices.UpdateTransaction)[len(*devices.UpdateTransaction)-1].Status
			if status == models.UpdateStatusBuilding {
				status = models.DeviceViewStatusUpdating
			}
		}
```

WDYT with we go with that?

AaronH88: Seems ok, 

Ill add now
Tyvm!

ldjebran (approved):
LGMT

PR #963, opened 2022-04-19
[DRAFT ]first try to generate RHEL9 image

acosferreira:
# Description
First try to map the responses from image-builder and make sure we can do it.

### DIFF
--- a/pkg/clients/imagebuilder/client.go
+++ b/pkg/clients/imagebuilder/client.go
@@ -257,7 +257,7 @@ func (c *Client) ComposeInstaller(image *models.Image) (*models.Image, error) {
 				Architecture: image.Commit.Arch,
 				ImageType:    models.ImageTypeInstaller,
 				Ostree: &OSTree{
-					Ref: "rhel/8/x86_64/edge", //image.Commit.OSTreeRef,
+					Ref: "rhel/9/x86_64/edge", //image.Commit.OSTreeRef,
 					URL: image.Commit.Repo.URL,
 				},
 				UploadRequest: &UploadRequest{
--- a/pkg/models/images.go
+++ b/pkg/models/images.go
@@ -73,9 +73,9 @@ const (
 	NoOutputTypes = "an output type is required"
 
 	// ImageTypeInstaller is the installer image type on Image Builder
-	ImageTypeInstaller = "rhel-edge-installer"
+	ImageTypeInstaller = "edge-installer"
 	// ImageTypeCommit is the installer image type on Image Builder
-	ImageTypeCommit = "rhel-edge-commit"
+	ImageTypeCommit = "edge-commit"
 
 	// ImageStatusCreated is for when an image is created
 	ImageStatusCreated = "CREATED"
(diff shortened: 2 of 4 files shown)


pkg/clients/imagebuilder/client.go
```diff
@@ -257,7 +257,7 @@ func (c *Client) ComposeInstaller(image *models.Image) (*models.Image, error) {
 				Architecture: image.Commit.Arch,
 				ImageType:    models.ImageTypeInstaller,
 				Ostree: &OSTree{
-					Ref: "rhel/8/x86_64/edge", //image.Commit.OSTreeRef,
+					Ref: "rhel/9/x86_64/edge", //image.Commit.OSTreeRef,
```
ldjebran, line 260: we need to support both 8 and 9 rhel major version 
as 8.6 is in beta and 9.0 in beta too 
so we should support both refs "rhel/9/x86_64/edge"  and "rhel/8/x86_64/edge"

pkg/models/images.go
```diff
@@ -73,9 +73,9 @@ const (
 	NoOutputTypes = "an output type is required"
 
 	// ImageTypeInstaller is the installer image type on Image Builder
-	ImageTypeInstaller = "rhel-edge-installer"
```
omaciel, line 76: We don't need to do this anymore as Sanne has another PR to reintroduce the 'RHEL-' alias for RHEL 9 too.

templates/templateKickstart.ks
```diff
@@ -19,8 +19,8 @@ echo PRE
 # Auto-detect a dir at that location and inject it into the command list for install
 # Default to prior ostree/repo location in 8.4
 [[ -d /run/install/repo/ostree ]] \
-	&& echo "ostreesetup --nogpg --osname=rhel-edge --remote=rhel-edge --url=file:///run/install/repo/ostree/repo --ref=rhel/8/x86_64/edge" > /tmp/ostreesetup \
-	|| echo "ostreesetup --nogpg --osname=rhel-edge --remote=rhel-edge --url=file:///ostree/repo --ref=rhel/8/x86_64/edge" > /tmp/ostreesetup
+	&& echo "ostreesetup --nogpg --osname=rhel-edge --remote=rhel-edge --url=file:///run/install/repo/ostree/repo --ref=rhel/9/x86_64/edge" > /tmp/ostreesetup \
```
omaciel, line 22: This has to handle both `8` and `9`.

loadtheaccumulator: Yeah. this will break the install on 8, but it should be easily distinguishable with another line here...

loadtheaccumulator (changes requested):

templates/templateKickstart.ks
loadtheaccumulator: Line 22 should stay 8. Line 23 applies to both 8 and 9 and the full ref path can be grabbed with...
ostree refs --repo=/ostree/repo

```
e.g.,
**$ ostree refs --repo=ostree/repo**
rhel/9/x86_64/edge
```

loadtheaccumulator: SIDENOTE: I left off the leading slash in the example above, because I'm mounted to /mnt and cd'd into /mnt to run the command /ostree/repo is at the root of the ISO at kick.

PR #979, opened 2022-04-23
Theedge 2068 add devices info

acosferreira:
# Description

return devices details inside device groups as we have on device view endpoint to avoid different responses and avoid call inventory api 

Fixes # ([issue](https://issues.redhat.com/browse/THEEDGE-2068))

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/devicegroups.go
+++ b/pkg/routes/devicegroups.go
@@ -31,10 +31,6 @@ func setContextDeviceGroupDetails(ctx context.Context, deviceGroup *models.Devic
 	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
 }
 
-func setContextDeviceGroupDetailsView(ctx context.Context, deviceGroup *models.DeviceGroupDetailsView) context.Context {
-	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
-}
-
 type deviceGroupDeviceKeyType string
 
 const deviceGroupDeviceKey = deviceGroupDeviceKeyType("device_group_device_key")
@@ -396,18 +392,6 @@ func GetDeviceGroupByID(w http.ResponseWriter, r *http.Request) {
 func getContextDeviceGroupDetails(w http.ResponseWriter, r *http.Request) *models.DeviceGroupDetails {
 	ctx := r.Context()
 	deviceGroupDetails, ok := ctx.Value(deviceGroupKey).(*models.DeviceGroupDetails)
-
-	if !ok {
-		ctxServices := dependencies.ServicesFromContext(ctx)
-		respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest("Failed getting device group from context"))
-		return nil
-	}
-	return deviceGroupDetails
-}
-
-func getContextDeviceGroupDetailsView(w http.ResponseWriter, r *http.Request) *models.DeviceGroupDetailsView {
-	ctx := r.Context()
-	deviceGroupDetails, ok := ctx.Value(deviceGroupKey).(*models.DeviceGroupDetailsView)
 
 	if !ok {
 		ctxServices := dependencies.ServicesFromContext(ctx)
--- a/pkg/routes/devicegroups.go
+++ b/pkg/routes/devicegroups.go
@@ -31,6 +31,10 @@ func setContextDeviceGroupDetails(ctx context.Context, deviceGroup *models.Devic
 	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
 }
 
+func setContextDeviceGroupDetailsView(ctx context.Context, deviceGroup *models.DeviceGroupDetailsView) context.Context {
+	return context.WithValue(ctx, deviceGroupKey, deviceGroup)
+}
+
 type deviceGroupDeviceKeyType string
 
 const deviceGroupDeviceKey = deviceGroupDeviceKeyType("device_group_device_key")
@@ -54,6 +58,10 @@ func MakeDeviceGroupsRouter(sub chi.Router) {
 		r.Route("/details", func(d chi.Router) {
 			d.Use(DeviceGroupDetailsCtx)
 			d.Get("/", GetDeviceGroupDetailsByID)
+		})
+		r.Route("/view", func(d chi.Router) {
+			d.Use(DeviceGroupDetailsCtxView)
+			d.Get("/", GetDeviceGroupDetailsByIDView)
 		})
 		r.Route("/devices/{DEVICE_ID}", func(d chi.Router) {
 			d.Use(DeviceGroupDeviceCtx)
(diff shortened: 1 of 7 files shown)


InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
736636f2-c794-4976-a88e-9ba41a2802bc
8a5c25c8-2b4c-4fb3-bbc9-54fd94300168

InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
0d868cba-882e-47d2-a5d3-5a56dac29a22
ecd56369-bb87-4cb3-b726-44394fbc1435

InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
551a088a-48e3-4a75-86ec-f567effc0443
6e53a224-8884-44ad-a07b-80b59acf485a

pkg/routes/devicegroups.go
```diff
@@ -108,6 +116,57 @@ func DeviceGroupDetailsCtx(next http.Handler) http.Handler {
 	})
 }
 
+// DeviceGroupDetailsCtxView is a handler to Device Group Details requests
+func DeviceGroupDetailsCtxView(next http.Handler) http.Handler {
```
ldjebran, line 120: what make this function diffrent from DeviceGroupCtx ? why we need a new ctx function to get devicegroup ?

pkg/services/devicegroups.go
```diff
@@ -306,6 +307,44 @@ func (s *DeviceGroupsService) GetDeviceGroupDetailsByID(ID string) (*models.Devi
...
+		s.log.WithField("error", err.Error()).Error("Device details query error")
+		return nil, new(DeviceGroupNotFound)
+	}
+
+	var devicesIds []int
+	for _, device := range deviceGroupDetails.DeviceGroup.Devices {
+		devicesIds = append(devicesIds, int(device.ID))
+	}
+	var storedDevices []models.Device
+	if res := tx.Limit(limit).Offset(offset).Where("account = ? and ID in ?", account, devicesIds).
```
ldjebran, line 329: maybe we can pass ctx to the existing devices service 
```go
    newCtx := ctx.Where("ID in (?)", devicesIds)
   services.DeviceServices.GetDevicesView(limit, offset, newCtx) 

```

what do you think, rest to think how to get the devicesService

pkg/routes/devicegroups.go
```diff
@@ -108,6 +116,57 @@ func DeviceGroupDetailsCtx(next http.Handler) http.Handler {
...
+			if err != nil {
+				ctxServices.Log.Debug("ID is not an integer")
+				respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest(err.Error()))
+				return
+			}
+
+			tx := devicesFilters(r, db.DB).Where("image_id IS NOT NULL AND image_id != 0")
+			pagination := common.GetPagination(r)
+
+			fmt.Printf("\n ID >>> %v\n", ID)
```
ldjebran, line 137: not needed anymore

pkg/routes/devicegroups.go
ldjebran: think that resolving business logic in the context function , make the code very unreadable,

Think we should use DeviceGroupCtx  and resolve the business logic  in GetDeviceGroupDetailsByIDView

pkg/routes/devicegroups.go
```diff
@@ -108,6 +116,57 @@ func DeviceGroupDetailsCtx(next http.Handler) http.Handler {
...
+			_, err := strconv.Atoi(ID)
+			ctxServices.Log = ctxServices.Log.WithField("deviceGroupID", ID)
+			ctxServices.Log.Debug("Retrieving device group")
+			if err != nil {
+				ctxServices.Log.Debug("ID is not an integer")
+				respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest(err.Error()))
+				return
+			}
+
+			tx := devicesFilters(r, db.DB).Where("image_id IS NOT NULL AND image_id != 0")
```
AaronH88, line 134: Could we put `account` in this search query and just return an error here is the account dosent have the device group id

AaronH88, line 134: Is this re-using the filters made for devices?
Does this work ok? As their are some filters that are device specific, e.g. `update_available` and `device.uuid` 

Perhaps we need to make a specific `devicegroupFilters`

pkg/routes/devicegroups.go
```diff
@@ -108,6 +116,57 @@ func DeviceGroupDetailsCtx(next http.Handler) http.Handler {
...
+			fmt.Printf("\n ID >>> %v\n", ID)
+			deviceGroup, err := ctxServices.DeviceGroupsService.GetDeviceGroupDetailsByIDView(ID, pagination.Limit, pagination.Offset, tx)
+			if err != nil {
+				var responseErr errors.APIError
+				switch err.(type) {
+				case *services.DeviceGroupNotFound:
+					responseErr = errors.NewNotFound(err.Error())
+				default:
+					responseErr = errors.NewInternalServerError()
+					responseErr.SetTitle("failed getting third device group")
```
AaronH88, line 146: This title seems a little strange, why `third` device group?

pkg/routes/devicegroups.go
acosferreira: yes, the idea is be able to do the same filter for a device that belong a device group.
Given I have the list of devices for a group, I'd like to be able to perform the same filter and sort the list of device view
For the device groups the filters are applied in the list of groups not on details.

pkg/routes/devicegroups.go
acosferreira: At this moment there is no difference between the context that they can be handled by the logic.
We are adding this "duplication" to avoid break the FE in case we merge the BE first.
Instead use the /details we will use /view

pkg/routes/devicegroups.go
```diff
@@ -108,6 +116,57 @@ func DeviceGroupDetailsCtx(next http.Handler) http.Handler {
 	})
 }
 
+// DeviceGroupDetailsCtxView is a handler to Device Group Details requests
+func DeviceGroupDetailsCtxView(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctxServices := dependencies.ServicesFromContext(r.Context())
+
+		if ID := chi.URLParam(r, "ID"); ID != "" {
```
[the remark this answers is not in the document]
acosferreira: its following the validation needed after the ID check

matoval:
Tested this PR with Edge-frontend [PR#405](https://github.com/RedHatInsights/edge-frontend/pull/405) and everything is working correctly on the UI.

pkg/routes/devicegroups.go
ldjebran: I understand , but you will break nothing if the logic will be in handler it self and not in the context handler.

ldjebran: + we are overwriting the devicegroup context value , with details one when using the same key.

ldjebran: so the solution may be like this (me and @AaronH88 was looking at this)

file pkg/routes/devicegroups.go
```go
func MakeDeviceGroupsRouter(sub chi.Router) {
        ....
	sub.Route("/{ID}", func(r chi.Router) {
	       ...
		r.Get("/devices-view", GetDeviceGroupDetailsByIDView)
                ....
	})
}

// GetDeviceGroupDetailsByIDView return devices groups details for an account and Id
func GetDeviceGroupDetailsByIDView(w http.ResponseWriter, r *http.Request) {
	ctxServices := dependencies.ServicesFromContext(r.Context())
	deviceGroup := getContextDeviceGroup(w, r)
	if deviceGroup == nil {
		return // error handled by getContextDeviceGroup already
	}

	devicesIDS := make([]uint, 0, len(deviceGroup.Devices))
	for _, device := range deviceGroup.Devices {
		devicesIDS = append(devicesIDS, device.ID)
	}
	tx := devicesFilters(r, db.DB).Where("image_id IS NOT NULL AND image_id != 0 AND ID IN (?)", devicesIDS)
	pagination := common.GetPagination(r)

	devicesCount, err := ctxServices.DeviceService.GetDevicesCount(tx)
	if err != nil {
		respondWithAPIError(w, ctxServices.Log, errors.NewNotFound("No devices found"))
		return
	}
	deviceGroupDevices, err := ctxServices.DeviceService.GetDevicesView(pagination.Limit, pagination.Offset, tx)
	if err != nil {
		responseErr := errors.NewInternalServerError()
		responseErr.SetTitle("failed getting device group")

		respondWithAPIError(w, ctxServices.Log, responseErr)
		return
	}
	var deviceGroupDetails models.DeviceGroupDetailsView
	deviceGroupDetails.DeviceGroup = deviceGroup
	deviceGroupDetails.DeviceDetails.Devices = deviceGroupDevices.Devices
	deviceGroupDetails.DeviceDetails.Total = int(devicesCount)
	respondWithJSONBody(w, ctxServices.Log, &deviceGroupDetails)
}


```
in conclusion we we added one route , and one handler function

AaronH88: This is a lot smaller code, and dosent require changes to the context. The model is the same as Ana proposed,

We think this may be a simpler solution, we tested locally and it works.
Let us know what you think Ana :smile:

acosferreira: :+1:

ldjebran (approved):
LGMT

PR #1067, opened 2022-05-18
Fix incorrect ImageSet version for Updated Image 

Saurabh-Thakre:
# Description

[BE] EDGE-API sending incorrect ImageSet version for the latest image.

Fixes # [THEEDGE-2164]

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -221,40 +221,39 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
 	image.ImageSetID = previousImage.ImageSetID
 	image.Account = previousImage.Account
 
-	if previousImage.Status == models.ImageStatusSuccess {
-		// Previous image was built successfully
-		var currentImageSet models.ImageSet
-		result := db.DB.Where("Id = ?", previousImage.ImageSetID).First(&currentImageSet)
-
-		if result.Error != nil {
-			s.log.WithField("error", result.Error.Error()).Error("Error retrieving the image set from parent image")
-			return result.Error
-		}
-		currentImageSet.Version = currentImageSet.Version + 1
-		if err := db.DB.Save(currentImageSet).Error; err != nil {
-			return result.Error
-		}
-
-		// Always get the repo URL from the previous Image's commit
-		repo, err := s.RepoService.GetRepoByID(previousImage.Commit.RepoID)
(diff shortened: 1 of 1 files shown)


InsightsDroid:
Test results are available in [Ibutsu](https://url.corp.redhat.com/ibutsu-runs). The test run IDs are:
686d2164-5afb-4d7e-b662-04b011bc5105
da3255c5-80d5-4134-a400-2313a66ba211

loadtheaccumulator:
Does it matter if the previous image version errored?

loadtheaccumulator (commented):
Does it matter if the previous image version errored?

Saurabh-Thakre:
> Does it matter if the previous image version errored?

Yeah kinda, If the previous image version is errored then it might possible most of the time it doesn't have the repo url. so the next will also not have.

pkg/services/images.go
```diff
@@ -221,20 +221,18 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
...
+	var currentImageSet models.ImageSet
+	result := db.DB.Where("Id = ?", previousImage.ImageSetID).First(&currentImageSet)
+	if result.Error != nil {
+		s.log.WithField("error", result.Error.Error()).Error("Error retrieving the image set from parent image")
+		return result.Error
+	}
+	currentImageSet.Version = currentImageSet.Version + 1
+	if err := db.DB.Save(currentImageSet).Error; err != nil {
+		return result.Error
+	}
```
ldjebran, line 233: at this stage imageSet is saved with version = previousVersion + 1 (but the image is not saved yet , with new version).
what will happen if there was an error and we return before saving the image with new version = previousVersion +1

Propose to not calculate the new imageSet version from previous image set version , but instead from previous image version

eg:
```go
currentImageSet.Version = previousImage.Version + 1
```

This way we will be always safe , no matter what can happen.

what do you think ?

@loadtheaccumulator your thoughts?

Saurabh-Thakre: makes sense to me. This will make sure we're not unnecessarily updating the ImageSet version in case of images are not created.

ldjebran (approved):
LGMT

PR #1080, opened 2022-05-24
updates: FIXES devices update, when device does not exit.

ldjebran:
Added devices' validation.
Moved base response function to main function.
Added routes > readAccount as used in too many places.
Added some tests for device updates.

Fixes # THEEDGE-2200

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


loadtheaccumulator (approved):
/lgtm

pkg/routes/updates.go
```diff
@@ -148,115 +149,125 @@ func GetUpdates(w http.ResponseWriter, r *http.Request) {
...
+	if result := db.DB.Debug().Model(&models.Device{}).Where("account = ? AND uuid IN (?)", account, devicesUUID).Count(&devicesCount); result.Error != nil {
+		ctxServices.Log.WithFields(log.Fields{
+			"error": result.Error.Error(),
+		}).Error("failed to get devices count")
+		apiError := errors.NewInternalServerError()
+		apiError.SetTitle("failed to get devices count")
+		respondWithAPIError(w, ctxServices.Log, apiError)
+		return nil
+	}
+	if int64(len(devicesUUID)) != devicesCount {
```
acosferreira, line 192: I think we should log the error for the device not found but continue the process to the existent devices. In case we fire the update for 1000 devices and 10 are wrong, we should update the 990.
What do you think?

ldjebran: Think we should not silently ignore the non existent one, only in case we received that option from the user, say there is an option in the request to Ignore if not  exist, this may be realized a standard feature in our app, on how to deal with this situations. 
say three options:
0 - See the operation as atomic one, all items must pass or no one, as a default option.
1 - give option to ignore the non existent items.
2 - silently ignore the non existent one , but return back a list of the non existent items.

But think  anyway this is not a blocker for this (option 0) bug fix, as this is a feature  that should be discussed to be followed in other areas.

thom-at-redhat:
@mergifyio unqueue

PR #1081, opened 2022-05-25
Added Unleash feature flag support, an example of usage in thridparty…

AaronH88:
… repo and a fake server for unit testing.

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -84,7 +84,7 @@ objects:
         - name: LOG_LEVEL
           value: ${LOG_LEVEL}
         - name: UNLEASH_URL
-            value: ${UNLEASH_URL}
+          value: ${UNLEASH_URL}
         - name: UNLEASH_TOKEN
           valueFrom:
             secretKeyRef:
@@ -136,7 +136,7 @@ objects:
         - name: LOG_LEVEL
           value: ${LOG_LEVEL}
         - name: UNLEASH_URL
-            value: ${UNLEASH_URL}
+          value: ${UNLEASH_URL}
         - name: UNLEASH_TOKEN
           valueFrom:
             secretKeyRef:
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -83,6 +83,13 @@ objects:
           value: ${UPLOADWORKERS}
         - name: LOG_LEVEL
           value: ${LOG_LEVEL}
+        - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+        - name: UNLEASH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: ${UNLEASH_SECRET_NAME}
+              key: CLIENT_ACCESS_TOKEN
         resources:
           limits:
             cpu: ${{CPU_LIMIT}}
@@ -128,6 +135,13 @@ objects:
           value: ${UPLOADWORKERS}
         - name: LOG_LEVEL
           value: ${LOG_LEVEL}
+        - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
+        - name: UNLEASH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: ${UNLEASH_SECRET_NAME}
+              key: CLIENT_ACCESS_TOKEN
         resources:
           limits:
             cpu: ${{CPU_LIMIT}}
@@ -230,3 +244,9 @@ parameters:
   name: LOG_LEVEL
   required: false
   value: "DEBUG"
+- description: Unleash API url
+  name: UNLEASH_URL
+  value: ''
+- description: Unleash secret name
+  name: UNLEASH_SECRET_NAME
+  value: unleash-ephemeral
--- a/pkg/routes/thirdpartyrepo.go
+++ b/pkg/routes/thirdpartyrepo.go
@@ -136,6 +136,7 @@ func GetAllThirdPartyRepo(w http.ResponseWriter, r *http.Request) {
 	enabled := feature.CheckFeatureWithAccount(account, feature.FeatureCustomRepos)
 	if !enabled {
 		respondWithAPIError(w, ctxServices.Log, errors.NewFeatureNotAvailable("Feature not available"))
+		return
 	}
 
 	pagination := common.GetPagination(r)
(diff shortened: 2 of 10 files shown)


pkg/routes/thirdpartyrepo.go
```diff
@@ -124,6 +126,26 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+		return
+	}
+
+	// Check to see if feature is enabled
+	unleashCtx := unleashCTX.Context{
+		UserId: account,
+	}
+	enabled := unleash.IsEnabled("fleet-management.custom-repos", unleash.WithContext(unleashCtx))
+	if !enabled {
+		respondWithJSONBody(w, services.Log, "Feature not available")
```
ldjebran, line 146: I think we are missing a return here ?

ldjebran: maybe we should return an error here instead of 200 , as the api contract say 200 has a special structure

pkg/routes/thirdpartyrepo.go
```diff
@@ -124,6 +126,26 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+			services.Log.WithField("error", err.Error()).Error("Error while trying to encode")
+		}
+		return
+	}
+
+	// Check to see if feature is enabled
+	unleashCtx := unleashCTX.Context{
+		UserId: account,
+	}
+	enabled := unleash.IsEnabled("fleet-management.custom-repos", unleash.WithContext(unleashCtx))
```
ldjebran, line 144: 1 - how to with , when we fail to connect to the unleash service?

2- may be we should encapsulate this code in some project function in "feature" module example

```go
const FeatureCustomRepos = "fleet-management.custom-repos"

func isFeatureEnabled(account, feature) bool {
     unleashCtx := unleashCTX.Context{
		UserId: account,
	}
       return unleash.IsEnabled("fleet-management.custom-repos", unleash.WithContext(unleashCtx))
}

```
in the previous function we can use cash ...


from code 
```go
import feature 

featureEnabled = features.isFeatureEnabled(account, features.FeatureCustomRepos)

```

AaronH88: Ill add nil checking code for if we fail to connect :smile:

AaronH88: As for encapsulating the client, I did some research and the client does cache the feature toggles, this is why when we initialize the client we set a refresh time. (Currently set to 5 seconds)

https://docs.getunleash.io/client-specification

"To be super fast, the client SDK caches all feature toggles and their current configuration in memory."

Because the client does this for us out of the box, I dont see the need to further encapsulate it, please let me know if Im wrong :smile:

pkg/routes/thirdpartyrepo.go
AaronH88: Cool, I can change it.

My thoughts here where that perhaps an error would be confusing. An error , to me, suggests that I have done something wrong, maybe a wrong payload or bad auth, when in-fact I have done everything correctly but I just dont have access to the feature.

Happy to switch it to an error if we think thats the more appropriate solution.

pkg/routes/thirdpartyrepo.go
ldjebran: so you propose to write 4 line of code each time we want to check instead one line?

pkg/routes/thirdpartyrepo.go
AaronH88: Added an error for feature not available and gave it a status 501 :smile:

deploy/clowdapp.yaml
```diff
@@ -83,6 +83,13 @@ objects:
           value: ${UPLOADWORKERS}
         - name: LOG_LEVEL
           value: ${LOG_LEVEL}
+        - name: UNLEASH_URL
+            value: ${UNLEASH_URL}
```
thom-at-redhat, line 87: $ yamllint /tmp/clowdapp.yaml 
/tmp/clowdapp.yaml
  24:20     warning  truthy value should be one of [false, true]  (truthy)
  87:18     error    syntax error: mapping values are not allowed here (syntax)

thom-at-redhat (dismissed):

deploy/clowdapp.yaml
AaronH88: The indentation was incorrect in this older version. Fixed now and the line 87 error does not appear.

pkg/routes/thirdpartyrepo.go
```diff
@@ -129,6 +131,13 @@ func GetAllThirdPartyRepo(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	ctx := thirdPartyRepoFilters(r, db.DB).Model(&models.ThirdPartyRepo{}).Where("account = ?", account)
+
+	// Check to see if feature is enabled
+	enabled := feature.CheckFeatureWithAccount(account, feature.FeatureCustomRepos)
+	if !enabled {
+		respondWithAPIError(w, ctxServices.Log, errors.NewFeatureNotAvailable("Feature not available"))
```
ldjebran, line 138: Missing return

ldjebran (approved):
LGMT

unleash/features/feature.go
```diff
@@ -0,0 +1,17 @@
+package feature
+
+import (
+	"github.com/Unleash/unleash-client-go/v3"
+	unleashCTX "github.com/Unleash/unleash-client-go/v3/context"
+)
+
+//FeatureCustomRepos is the const of the custom repo feature flag
+const FeatureCustomRepos = "fleet-management.custom-repos"
```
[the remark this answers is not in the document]
AaronH88: Add and remove them here, yes.

Feature flags are meant to be temporary :smile:

deploy/clowdapp.yaml
```diff
@@ -230,3 +244,9 @@ parameters:
   name: LOG_LEVEL
   required: false
   value: "DEBUG"
+- description: Unleash API url
+  name: UNLEASH_URL
+  value: ''
+- description: Unleash secret name
+  name: UNLEASH_SECRET_NAME
+  value: unleash-ephemeral
```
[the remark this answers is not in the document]
AaronH88: As far as I know, all

PR #1085, opened 2022-05-26
Theedge 2165 add rhel9 support

acosferreira:
# Description

Add support to RHELX image creation based on distribution value.

Fixes # (THEEDGE-2165)

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:


pkg/services/images.go
```diff
@@ -36,6 +37,11 @@ import (
 // FIXME: this no longer applies to images. move to devices
 var WaitGroup sync.WaitGroup
 
+var (
+	oSTreeRefDistribution = regexp.MustCompile(`[a-zA-Z]+`)
```
loadtheaccumulator, line 41: oSTree... is that naming intentional or a typo?

loadtheaccumulator (changes requested):
have some questions

pkg/models/required_image_packages.json
```diff
@@ -0,0 +1,10 @@
+{   "8":["ansible"],
```
ldjebran, line 1: why not save them as map and as go lang var we may prefer refering them as per distribution and not as per major version

eg:
  ```go 
var distributionsPackages = map[string][]string{"rhel-84": {"packege1", "package2"}}
```

Think we should keep it very simple and do not rely on calculations, as we have no standard to follow. and we do not know the major version is for for which  dist.

File: pkg/services/images.go [hunk 3]
```diff
@@ -242,12 +248,17 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
...
 
 		image.Commit.OSTreeParentCommit = repo.URL
+
+		osTreeDistribution := osTreeRefDistribution.FindStringSubmatch(image.Distribution)[0]
+		osTreeVersion := osTreeRefVersion.FindStringSubmatch(image.Distribution)[0]
+		if osTreeDistribution == "" || osTreeVersion == "" {
+			s.log.WithField("error", err.Error()).Error("No OSTREE found to this distribution")
+			return &OstreeNotFound{}
+		}
+		defaultOstree := fmt.Sprintf("%s/%s/x86_64/edge", osTreeDistribution, osTreeVersion)
```
ldjebran, line 258: Do not a reason we are calculating the ostree ref , as we we have no standard to follow, 
vote to keep it very simple and readable a map dist => ostree ref  will be very sufficient

```go
var distributionsRefs = map[string]string{
	// I am not afraid to repeat values
	"rhel-84": "rhel/8/x86_64/edge",
	"rhel-85": "rhel/8/x86_64/edge",
	"rhel-90": "rhel/9/x86_64/edge",
}
```

ldjebran (commented):
Think we should keep the initial data very simple , and avoid values calculation.

pkg/models/images.go
```diff
@@ -200,7 +196,20 @@ func (i *Image) HasOutputType(imageType string) bool {
...
+	distributionPackage := config.DistributionsPackages[i.Distribution]
+	requiredPackages := config.RequiredPackages
+
+	if i.Distribution == "" {
+		return nil
+	}
+
+	for _, x := range distributionPackage {
+		requiredPackages = append(requiredPackages, x)
+	}
```
ldjebran, line 209: doing so we are modifying the original config slice, what we should not do.

propose 

```go
        requiredPackages := make([]string, 0, len(distributionPackages)+len(config.RequiredPackages))
	requiredPackages = append(requiredPackages, config.RequiredPackages...)
	requiredPackages = append(requiredPackages, distributionPackages...)
```

pkg/services/images.go
```diff
@@ -242,12 +242,10 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
...
 		image.Commit.OSTreeParentCommit = repo.URL
+		refs := config.DistributionsRefs[image.Distribution]
+
 		if image.Commit.OSTreeRef == "" {
-			if previousImage.Commit.OSTreeRef != "" {
-				image.Commit.OSTreeRef = previousImage.Commit.OSTreeRef
-			} else {
-				image.Commit.OSTreeRef = config.Get().DefaultOSTreeRef
-			}
+			image.Commit.OSTreeRef = refs
```
ldjebran, line 248: if the image.Distribution does not exist in config.DistributionsRefs   refs will equal ""
think we still  need more cases.

config/repo_config.go
```diff
@@ -0,0 +1,32 @@
...
+
+// RHEL8 contains aditional list of packages to build an image to >= RHEL85
+var RHEL8 = []string{"ansible"}
+
+// RHEL86 contains aditional list of packages to build an image to = RHEL86
+var RHEL86 = []string{"ansible-core"}
+
+// RHEL90 contains aditional list of packages to build an image to = RHEL90
+var RHEL90 = []string{"ansible-core"}
+
```
ldjebran, line 18: Think we still need Default distribution 
```go
const RHEL84 = "rhel-84"
const RHEL85 = "rhel-85"
const RHEL86 = "rhel-86"
const RHEL90 = "rhel-90"

const DefaultDistribution = RHEL85 

```

pkg/services/images.go
acosferreira: I think in this case if the distribution doesn't exists we should fail the image creation instead set a default, what do you think?

ldjebran: Yes let's fail it

config/repo_config.go
```diff
@@ -0,0 +1,36 @@
+package config
+
+// DefaultDistribution set the default image distribution in case miss it
+const DefaultDistribution = "rhel-85"
```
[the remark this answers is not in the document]
acosferreira: Agreed, we should use the last stable version

loadtheaccumulator (approved):
lgtm

PR #1150, opened 2022-06-14
Removing unneeded secret

jhutar:
# Description

During investigation some issue with Patrick Martin in https://coreos.slack.com/archives/C02CMTM9GG1/p1655137077271099 we noticed that this secret is actually not needed because in app-interface saas file there is:

```
managedResourceTypes:
- ClowdApp
```

## Type of change

- [x] Refactor

# Checklist:

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -4,12 +4,6 @@ kind: Template
 metadata:
   name: edge-api
 objects:
-- apiVersion: v1
-  kind: Secret
-  metadata:
-    name: psk-playbook-dispatcher
-  stringData:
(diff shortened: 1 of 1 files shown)


thom-at-redhat (approved):

jhutar:
We also discussed with Patrick Martin this bit might be used in non-standard deployments, like ephemeral cluster?

thom-at-redhat:
@jhutar testing failed

jhutar:
Ah, so looks like it is needed by bonfire deployment. Sorry for the mess.

jhutar:
Hello @bennyturns . Looks like this `mergify` bot did not noticed this is closed. Can it be silenced somehow?

PR #1153, opened 2022-06-14
Add more use case to check validation of invalid url

mgold1234:
related to THEEDGE-1210

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:


pkg/services/thirdpartyrepo_test.go
```diff
@@ -71,13 +72,19 @@ var _ = Describe("ThirdPartyRepos basic functions", func() {
...
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(Equal("invalid URL"))
-		})
+		},
+			Entry("validate invalid URL with : missing", "http//google.com"),
+			Entry("validate invalid URL with https:// missing", "google.com"),
+			Entry("validate invalid URL without https:// and . missing", "foo/bar"),
+			Entry("validate invalid URL with number", "5432/bar"),
+			Entry("validate invalid URL with symbols", "http//:&%$#@bar"),
+		)
```
ldjebran, line 87: "http://invalid-internet-host"  will pass validation

mgold1234: On the UI  we check if the url starts with https:// and is a valid url and don't check the url response.
maybe do you have an idea to solve the response of url?

ldjebran: I am not here talking about response problem at all.

only the host name is invalid , it should be of the form of :  foo.bar

ldjebran: so http://invalid-internet-host/llll   should not pass validation
bout "http://valid-internet-host.com/llll  should pass validation

ldjebran: you can make some testing here  https://www.regextester.com/94502

PR #1179, opened 2022-06-21
added org_id to image-sets

tomershinhar:
# Description

added org_id support to the image-sets route/service.
also reworked the unit-tests for the image-sets service and fixed some errors found on images service

Fixes # (issue)

[THEEDGE-1983](https://issues.redhat.com/browse/THEEDGE-1983)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

# Checklist:

### DIFF
--- a/pkg/services/imagesets.go
+++ b/pkg/services/imagesets.go
@@ -4,8 +4,8 @@ import (
 	"context"
 
 	"github.com/redhatinsights/edge-api/pkg/db"
-	"github.com/redhatinsights/edge-api/pkg/errors"
 	"github.com/redhatinsights/edge-api/pkg/models"
+	"github.com/redhatinsights/edge-api/pkg/routes/common"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/services/imagesets.go
```diff
@@ -30,17 +30,20 @@ type ImageSetsService struct {
...
+	}
+	result := db.AccountOrOrg(account, orgID, "image_sets").Debug().First(&imageSet, imageSetID)
 	if result.Error != nil {
 		s.log.WithField("error", result.Error.Error()).Error("Error getting image set by id")
-		err := errors.NewInternalServerError()
-		return nil, err
+		return nil, new(ImageSetNotFoundError)
 	}
-	result = db.DB.Where("image_set_id = ?", imageSetID).Find(&imageSet.Images)
+	result = db.AccountOrOrg(account, orgID, "").Debug().Where("image_set_id = ?", imageSetID).Find(&imageSet.Images)
```
AaronH88, line 43: is .Debug() needed here?

pkg/services/imagesets_test.go
```diff
@@ -1,19 +1,43 @@
-package services
+package services_test
```
AaronH88, line 1: Any reason for the package change? :smile:

tomershinhar: well, i started with remaking the one unit-test here, but thinking about adding more, i think it might be better to use `services_test` and not `services`

pkg/services/imagesets.go
tomershinhar: i wanted the logger to log the query that is being executed, i can remove it if you think we don't need to

AaronH88: No its fine :smile:

PR #1208, opened 2022-06-24
FIX : "limit" returns the count as total number of image-sets 

Saurabh-Thakre:
# Description

"limit" list-all-image-sets endpoint returns the total number of image-sets in count and not the count of the returned payload

Fixes #THEEDGE-2223

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/imagesets.go
+++ b/pkg/routes/imagesets.go
@@ -137,7 +137,7 @@ type ImageSetInstallerURL struct {
 // ListAllImageSets return the list of image sets and images
 func ListAllImageSets(w http.ResponseWriter, r *http.Request) {
 	s := dependencies.ServicesFromContext(r.Context())
-
+	var count int64
 	var imageSet []models.ImageSet
 	var imageSetInfo []ImageSetInstallerURL
 	var result *gorm.DB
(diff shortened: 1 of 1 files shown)


acosferreira (changes requested):
when you look to this result, if there is no limit value, the result will be always 30, and this will breadk the ui pagination
you need to add a fix to the count, to when receive the parameter of limit, the count also consider this value and return the correct amount

AaronH88:
Waiting for Ana's comment to be addressed :smile:

ldjebran (commented):
I commented in the original issue , but also comment here.
This this is not a bug , this how pagination and filtering works, we return the count of all entities affected only and only by the the filter.    
Otherwise the pagination will not work. 

Think we should close this PR.

and the issue as not a bug.

Saurabh-Thakre:
@ldjebran @AaronH88 yes, I;m working on it to make sure we have a condition where if the user passes a limit then we will respect that limit and show it in count.

Saurabh-Thakre:
closing the PR, the bug has been declared as not a bug after discussion

PR #1219, opened 2022-06-28
Validate image update

acosferreira:
# Description

Add validation if we can generate the update image based on current distribution

Fixes # ([issue](https://issues.redhat.com/browse/THEEDGE-2178))

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/config/repo_config.go
+++ b/config/repo_config.go
@@ -34,3 +34,8 @@ var DistributionsRefs = map[string]string{
 	"rhel-86": "rhel/8/x86_64/edge",
 	"rhel-90": "rhel/9/x86_64/edge",
 }
+
+var UpgradedAvailable = map[string]string{
+	//to - from
+	"rhel-90": "rhel-86",
+}
--- a/config/repo_config.go
+++ b/config/repo_config.go
@@ -35,6 +35,7 @@ var DistributionsRefs = map[string]string{
 	"rhel-90": "rhel/9/x86_64/edge",
 }
 
+// UpgradedAvailable set from to upgrade allowed
 var UpgradedAvailable = map[string]string{
 	//to - from
 	"rhel-90": "rhel-86",
--- a/pkg/services/images_test.go
+++ b/pkg/services/images_test.go
@@ -873,7 +873,7 @@ var _ = Describe("Image Service Test", func() {
 	Context("update image to diff refs", func() {
 
 		Context("when previous is 8.5 to 9 should fail", func() {
-			It("should successfully generate the update", func() {
+			It("should not successfully generate the update", func() {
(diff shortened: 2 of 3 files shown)


InsightsDroid:
Test results are available in Ibutsu:
https://url.corp.redhat.com/ibutsu-runs-d0eefe78-1f0a-4209-b2ab-d2aff5e58e09

pkg/services/images_test.go
```diff
@@ -867,4 +869,192 @@ var _ = Describe("Image Service Test", func() {
...
+				account := faker.UUIDHyphenated()
+				imageSet := &models.ImageSet{Account: account}
+				result := db.DB.Save(imageSet)
+				Expect(result.Error).To(Not(HaveOccurred()))
+				previousImage := &models.Image{
+					Account:      account,
+					Status:       models.ImageStatusSuccess,
+					Commit:       &models.Commit{RepoID: &uid},
+					Version:      1,
+					Distribution: "rhel-85",
```
omaciel, line 888: Maybe use constants or variables to hold `rhel-85` and `rhel-90` as it would simplify updating the test, if needed, in the future?

config/repo_config.go
```diff
@@ -34,3 +34,9 @@ var DistributionsRefs = map[string]string{
 	"rhel-86": "rhel/8/x86_64/edge",
 	"rhel-90": "rhel/9/x86_64/edge",
 }
+
+// UpgradedAvailable set from to upgrade allowed
+var UpgradedAvailable = map[string]string{
```
AaronH88, line 39: Hi Ana, great start,

Lets have a chat today to go over this.

At a high level. 
This file is very important and we need to have a level of encapsulation. We should create some functions that return a `Copy()` of these objects, not pointers to the objects themselves, as this could cause a lot of errors in the future if this map was accidentally changed.

Also, we should try and get this data from a central source. Lets speak to the RHEL for Edge team and discover a way to load in these relationships.

Also, I highly recommend reversing the relationship in this map, and having the map be `string` and `[] string`, this way we save the possible upgrade paths. 

Also, we can change the map name to `UpgradePaths`

And finally, once we have this mappings all get up we can then `Discover_path_to_upgrade()` from one version to another, this becomes a search problem that we could solve with Breath First Search, Depth First Search, but I think we should use Dijkstra's shortest path algorithm, as we dont want to find the first path, but we want to find the shortest path from one to another.

Myself and Djebran will set up a call with you today to go over this :smile:

AaronH88 (changes requested):

PR #1235, opened 2022-07-01
add ostree parentRef to update image and be able to calculate the upg…

acosferreira:
…rade

# Description

Based on ImageBuilder info we need to inform the parentRef when we want to upgrade an image version.
I've added the new field to our commit structure and also to the payload that we send to them

Fixes # ([issue](https://issues.redhat.com/browse/THEEDGE-2178))

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -316,6 +316,10 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
 		if image.Commit.OSTreeRef == "" {
 			image.Commit.OSTreeRef = refs
 		}
+		if previousImage.Commit.OSTreeRef == "" {
+			image.Commit.OSTreeParentRef = config.DistributionsRefs[previousImage.Distribution]
+		}
+
 	} else {
 		// Previous image was not built successfully
 		s.log.WithField("previousImageID", previousImage.ID).Info("Creating an update based on a image with a status that is not success")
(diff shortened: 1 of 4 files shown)


pkg/services/images.go
```diff
@@ -316,6 +316,10 @@ func (s *ImageService) UpdateImage(image *models.Image, previousImage *models.Im
 		if image.Commit.OSTreeRef == "" {
 			image.Commit.OSTreeRef = refs
 		}
+		if previousImage.Commit.OSTreeRef == "" {
```
ldjebran, line 319: what will happen if the previousImage failed, may be we should find the latest successful one   as previous , if none previous successful found , we should make a create instead.
what do you think ?

acosferreira: we just consider previous image as success build
https://github.com/RedHatInsights/edge-api/blob/0b73548cd4a777689eded63c983c49b0c9ca651d/pkg/services/images.go#L298

AaronH88: I agree, it should be the previous successful one, 
Now, should we even save failed images :smile:

ldjebran: This is not a blocker as not related directly to this PR

AaronH88 (approved):
/lgtm

ldjebran (approved):
LGTM

PR #1254, opened 2022-07-07
Add -cover to test-clean-no-fdo and test-no-fdo

lahavyuv86:
# Description

Remove unneeded directories from our coverage and coverage-no-fdo commands.

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,10 +136,10 @@ test:
 	go test $(BUILD_TAGS) $$(go list $(BUILD_TAGS) ./... | grep -v /test/) $(TEST_OPTIONS)
 
 test-clean-no-fdo:
-	go test -count=1 $$(go list ./... | grep -v /test/) $(TEST_OPTIONS)
+	go test -count=1 $$(go list ./... | grep -v /test/) $(TEST_OPTIONS) -cover
 
 test-no-fdo:
-	go test $$(go list ./... | grep -v /test/) $(TEST_OPTIONS)
+	go test $$(go list ./... | grep -v /test/) $(TEST_OPTIONS) -cover
 
 vet:
 	go vet $(BUILD_TAGS) $$(go list $(BUILD_TAGS) ./... | grep -v /vendor/)
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ build-test-container:
 		.
 
 coverage:
-	go test $(BUILD_TAGS) $$(go list $(BUILD_TAGS) ./... | grep -v /test/) $(TEST_OPTIONS) -coverprofile=coverage.txt -covermode=atomic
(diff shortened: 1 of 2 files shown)


omaciel:
@lahavyuv86  take a list of the files/folders we won't be executing tests against and add them to the `grep` section of the command to exclude them like `$(go list ./… | grep -v -e /test/ -e /unleash/)`. Make sure to only exclude files/folders we know we don't want to run tests against.

Makefile
```diff
@@ -19,6 +19,7 @@ KUBECTL=kubectl
 NAMESPACE=default
 TEST_OPTIONS="-race"
 BUILD_TAGS=-tags=fdo
+EXCLUDE_DIRS=-e /cmd/db -e /cmd/kafka -e /cmd/migrate/orgmigration -e /config -e /pkg/clients/imagebuilder/mock_imagebuilder -e /pkg/imagebuilder/mock_imagebuilder -e /pkg/clients/inventory -e /pkg/clients/inventory/mock_inventory -e /pkg/clients/playbookdispatcher -e /pkg/errors -e /pkg/services/mock_services -e /unleash
```
omaciel, line 22: I think we do want to run tests against `/cmd/migrate/orgmigration`, `/pkg/clients/playbookdispatcher`, `/pkg/clients/inventory`, no?

Makefile
```diff
@@ -136,10 +137,10 @@ test:
 	go test $(BUILD_TAGS) $$(go list $(BUILD_TAGS) ./... | grep -v /test/) $(TEST_OPTIONS)
 
 test-clean-no-fdo:
-	go test -count=1 $$(go list ./... | grep -v /test/) $(TEST_OPTIONS)
+	go test -count=1 -cover $$(go list ./... | grep -v -e /test/ $(EXCLUDE_DIRS)) $(TEST_OPTIONS)
```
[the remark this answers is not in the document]
lahavyuv86: good idea :)

Makefile
lahavyuv86: @ldjebran can you verify which dirs we can exclude?

AaronH88:
We have specific `coverage` commands in this Makefile, I suggest editing them and not adding coverage to our normal test commands

omaciel:
Thanks @AaronH88  for pointing this out. Is there a reason why we don't run that then during PR checks? I'd love to see coverage information for every new PR as it gives us information about the health of our code changes

Makefile
ldjebran: we should exclude only those that contain no tests

ldjebran: /cmd/migrate/orgmigration  contain tests and should not be excluded

AaronH88 (changes requested):
Suggest moving these changes to the `coverage` command

AaronH88 (approved):
/lgtm

PR #1271, opened 2022-07-13
Theedge 2354 static delta

acosferreira:
# Description

revert previous change, and add logs

Fixes # (THEEDGE-2354)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/repobuilder.go
+++ b/pkg/services/repobuilder.go
@@ -124,7 +124,7 @@ func (rb *RepoBuilder) BuildUpdateRepo(id uint) (*models.UpdateTransaction, erro
 				rb.log.WithField("error", err.Error()).Error("Error downloading tar")
 				return nil, fmt.Errorf("error Upload repo repo :: %s", err.Error())
 			}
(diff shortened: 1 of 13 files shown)


pkg/services/repobuilder.go
```diff
@@ -124,7 +124,7 @@ func (rb *RepoBuilder) BuildUpdateRepo(id uint) (*models.UpdateTransaction, erro
 				rb.log.WithField("error", err.Error()).Error("Error downloading tar")
 				return nil, fmt.Errorf("error Upload repo repo :: %s", err.Error())
 			}
-			err = rb.ExtractVersionRepo(&commit, tarFileName, path)
+			err = rb.ExtractVersionRepo(update.Commit, tarFileName, path)
```
ldjebran, line 127: what is going to be changed here, as the first arg is used only for  logging in the target function ExtractVersionRepo

ldjebran: LGTM, 
but we need to revisit all this functionality , as the repo is updated for each device repo , and devices may use different update commits

PR #1280, opened 2022-07-15
images: FIXES image installer orgID.

ldjebran:
Notes: We may set installer to nil when no iso needed , but that may break sql queries, that rely on simple joins (as we will need left join), this will need more time to investigate. We only fixed without changing the behaviour e.g. as it is functioning now.

# Description

Fixes #THEEDGE-2402

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -3,6 +3,7 @@ package models
 import (
 	"errors"
 
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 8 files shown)


ldjebran:
Nice collaboration @AaronH88 :)

PR #1286, opened 2022-07-18
fix a problem when we try to create image with custom repo

mgold1234:
related to THEEDGE-2405
the problem was that we need to copy the orgID to customRepo table

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -82,29 +82,27 @@ type ImageService struct {
 	RepoService  RepoServiceInterface
 }
 
-// ValidateAllImageReposAreFromOrgID validates the account for Third Party Repositories
-func ValidateAllImageReposAreFromOrgID(orgID string, repos []models.ThirdPartyRepo) error {
+// ValidateAllImageReposAreFromOrgID validates the OrgID for Third Party Repositories
+func ValidateAllImageReposAreFromOrgID(orgID string, repos []models.ThirdPartyRepo) (*[]models.ThirdPartyRepo, error) {
 
 	if orgID == "" {
-		return new(OrgIDNotSet)
+		return nil, new(OrgIDNotSet)
 	}
-	if len(repos) == 0 {
-		return nil
+	var imagesRepos []models.ThirdPartyRepo
+	for _, custRepo := range repos {
+		var repo models.ThirdPartyRepo
+		if custRepo.ID == 0 {
+			return nil, new(ThirdPartyRepositoryNotFound)
+		}
+		if result := db.Org(orgID, "").First(&repo, custRepo.ID); result.Error != nil {
+			if result.Error == gorm.ErrRecordNotFound {
+				return nil, new(ThirdPartyRepositoryNotFound)
+			}
+			return nil, result.Error
+		}
+		imagesRepos = append(imagesRepos, repo)
 	}
-	var ids []uint
-	for _, repo := range repos {
-		ids = append(ids, repo.ID)
-	}
-
-	var existingRepos []models.ThirdPartyRepo
-	if result := db.Org(orgID, "").Select("id").Find(&existingRepos, ids); result.Error != nil {
-		return result.Error
-	}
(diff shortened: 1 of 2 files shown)


loadtheaccumulator (dismissed):
Have a question on how the repos are being reconstructed and what is passed back.

pkg/services/images.go
```diff
@@ -82,29 +82,27 @@ type ImageService struct {
...
+			if result.Error == gorm.ErrRecordNotFound {
+				return nil, new(ThirdPartyRepositoryNotFound)
+			}
+			return nil, result.Error
+		}
+		imagesRepos = append(imagesRepos, repo)
 	}
-
-	return nil
+	return &imagesRepos, nil
```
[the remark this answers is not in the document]
ldjebran: I think if any of the repos , is not from org_id we will return new(ThirdPartyRepositoryNotFound) as we are using filter db.Org(orgID, "")  
if all the repos are from org_id , we will reach line 105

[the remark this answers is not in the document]
loadtheaccumulator: That's the way I read it. But why recreate image.ThirdPartyRepositories if what goes in will equal what comes out?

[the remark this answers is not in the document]
ldjebran: we are getting the repos from the ui without orgIDS, if we do not get the original ones from db 
gorm will fail when saving , as we have a BefroreCreate that check's orgID if empty
gorm will try a create but fail on the BeforeCreate hook.

it appears , that gorm is always trying to create the extended model of a model. and that's what's  happening to repos,

loadtheaccumulator (commented):
Got the answer I needed. Thanks!

pkg/services/images.go
```diff
@@ -82,29 +82,27 @@ type ImageService struct {
 	RepoService  RepoServiceInterface
 }
 
-// ValidateAllImageReposAreFromOrgID validates the account for Third Party Repositories
-func ValidateAllImageReposAreFromOrgID(orgID string, repos []models.ThirdPartyRepo) error {
+// ValidateAllImageReposAreFromOrgID validates the OrgID for Third Party Repositories
+func ValidateAllImageReposAreFromOrgID(orgID string, repos []models.ThirdPartyRepo) (*[]models.ThirdPartyRepo, error) {
```
ldjebran, line 86: May be we should rename this function as it's not only validate, what do you think ?

ldjebran (commented):
Left one question

pkg/services/images_test.go
```diff
@@ -541,7 +541,7 @@ var _ = Describe("Image Service Test", func() {
 				repo2 := models.ThirdPartyRepo{OrgID: orgID, Name: faker.UUIDHyphenated(), URL: "https://repo2.simple.com"}
 				result = db.DB.Create(&repo2)
 				Expect(result.Error).ToNot(HaveOccurred())
-				err := services.ValidateAllImageReposAreFromOrgID(orgID, []models.ThirdPartyRepo{repo1, repo2})
+				_, err := services.ValidateAllImageReposAreFromOrgID(orgID, []models.ThirdPartyRepo{repo1, repo2})
```
ldjebran, line 544: maybe check that the len(returned repos) ==  2
and that the returned repos correspond to the supplied one

ldjebran, line 544: you may supply only repo ids here to simulate the worst case 
may be like 
```go
...  := services.ValidateAllImageReposAreFromOrgID(orgID, []models.ThirdPartyRepo{{ID: repo1.ID}, {ID: repo2.ID}})

```

pkg/services/images.go
mgold1234: WDYT about ths name - GetImageReposbyOrgID

ldjebran (approved):

PR #1302, opened 2022-07-20
Added a sync with inventory when devices view is called.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/pkg/services/devices.go
+++ b/pkg/services/devices.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"sort"
+	"time"
 
 	version "github.com/knqyf263/go-rpm-version"
 	"github.com/redhatinsights/edge-api/pkg/clients/inventory"
@@ -30,7 +31,7 @@ const (
 type DeviceServiceInterface interface {
 	GetDevices(params *inventory.Params) (*models.DeviceDetailsList, error)
 	GetDeviceByID(deviceID uint) (*models.Device, error)
-	GetDevicesView(limit int, offset int, tx *gorm.DB) (*models.DeviceViewList, error)
+	GetDevicesView(limit int, offset int, tx *gorm.DB, params *inventory.Params) (*models.DeviceViewList, error)
 	GetDevicesCount(tx *gorm.DB) (int64, error)
 	GetDeviceByUUID(deviceUUID string) (*models.Device, error)
 	// Device by UUID methods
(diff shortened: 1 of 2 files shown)


AaronH88:
Fixing tests next, want to get early feedback on this approach

pkg/services/devices.go
```diff
@@ -901,3 +919,148 @@ func (s *DeviceService) ProcessPlatformInventoryDeleteEvent(message []byte) erro
...
+
+					for _, dep := range inDevice.Ostree.RpmOstreeDeployments {
+						idep := RpmOSTreeDeployment{
+							Booted:   dep.Booted,
+							Checksum: dep.Checksum,
+						}
+						deps = append(deps, idep)
+					}
+					profile := systemProfile{
+						HostType:             "edge",
```
ldjebran, line 1038: You need to check the real device hostType, as device from inventory may be not edge

pkg/services/devices.go
```diff
@@ -901,3 +919,148 @@ func (s *DeviceService) ProcessPlatformInventoryDeleteEvent(message []byte) erro
 	}
 	return nil
 }
+
+func (s *DeviceService) syncDevicesWithInventory(orgID string) {
```
ldjebran, line 923: We may rename this function to syncDevicesAndInventory

And we may create two functions that this function will call , as they have different functionality 

syncDevicesWithInventory
syncInventoryWithDevices

what do you think ?

pkg/clients/inventory/client.go
```diff
@@ -203,6 +204,60 @@ func (c *Client) ReturnDevicesByID(deviceID string) (Response, error) {
...
 }
 
+// ReturnDeviceListByID will return the list of devices by uuid
+func (c *Client) ReturnDeviceListByID(deviceIDs []string) (Response, error) {
+	if _, err := uuid.Parse(deviceIDs[0]); err != nil {
+		c.log.WithFields(log.Fields{"error": err, "deviceID": deviceIDs[0]}).Error("invalid device ID")
+		return Response{}, err
+	}
+	url := fmt.Sprintf("%s/%s/%s", config.Get().InventoryConfig.URL, inventoryAPI, deviceIDs[0])
+	for _, deviceID := range deviceIDs[1:] {
```
ldjebran, line 214: what will happen if len(deviceIDs) == 1

propose to check if  len(deviceIDs) == 0  return direct with  Response{}, nil
and parse all the deviceIDS at onces if the first one is bad any way we going to break and return

what do you think ?

pkg/services/devices.go
```diff
@@ -659,14 +665,22 @@ func (s *DeviceService) GetDevicesView(limit int, offset int, tx *gorm.DB) (*mod
...
 
-	// create a map of device group info and map it to given devices
+	// Check inventory to see if you return the same number of devices, else, sync with inventory in a routine
+	var params *inventory.Params
+	inventoryDevices, err := s.Inventory.ReturnDevices(params)
+	if err != nil {
+		s.log.WithField("error", err.Error()).Error("Error retrieving devices from inventory")
+		return nil, err
+	}
+	if inventoryDevices.Total != len(storedDevices) {
```
ldjebran, line 681: The stored device is the result of returning using limit , offset and filtering , when inventory query is without nothing , or with default page and perPage, this will lead to always having  inventoryDevices.Total > len(storedDevices) 
mabe we should envisage an other strategy for this, what do you think ?

pkg/services/devices.go
AaronH88: Added it to the inventory device type and added it here :smile:

pkg/services/devices.go
AaronH88: Good idea, added

pkg/clients/inventory/client.go
AaronH88: Added two extra checks.
The first as you said for DeviceIDs == 0
And a second for DeviceIDs > 1

This is because I need to build the URL with a comma seprated list of device ids, and it can not end in a trailing comma e.g. `id, id, id` is valid but `id,id,id,` is not

pkg/services/devices.go
AaronH88: Good spot,
This should have been a `count` of the stored devices, changing it now

pkg/services/devices.go
```diff
@@ -58,18 +60,22 @@ type RpmOSTreeDeployment struct {
...
-		SystemProfile struct {
-			HostType             string                `json:"host_type"`
-			RpmOSTreeDeployments []RpmOSTreeDeployment `json:"rpm_ostree_deployments"`
-			RHCClientID          string                `json:"rhc_client_id,omitempty"`
-		} `json:"system_profile"`
-	} `json:"host"`
+	Host host   `json:"host"`
+}
+
+type systemProfile struct {
```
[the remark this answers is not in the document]
AaronH88: I am not exporting this type :smile: 
Do you have a reason for this to be exported?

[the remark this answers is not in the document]
mgold1234: I thought maybe in the future we want to use this struct outside :)

ldjebran (approved):
LGTM

PR #1313, opened 2022-07-25
image-sets: implement image-sets/view/{imageSetID} ... details endpoints.

ldjebran:
Fixes #THEEDGE-2317
## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/imagesets.go
+++ b/pkg/routes/imagesets.go
@@ -22,8 +22,10 @@ import (
 )
 
 type imageSetTypeKey int
+type imageSetImageTypeKey int
 
 const imageSetKey imageSetTypeKey = iota
+const imageSetImageKey imageSetImageTypeKey = iota
 
 var sortOption = []string{"created_at", "updated_at", "name"}
 var statusOption = []string{models.ImageStatusCreated, models.ImageStatusBuilding, models.ImageStatusError, models.ImageStatusSuccess}
@@ -35,6 +37,15 @@ func MakeImageSetsRouter(sub chi.Router) {
 	sub.Route("/{imageSetID}", func(r chi.Router) {
 		r.Use(ImageSetCtx)
 		r.With(validateFilterParams).With(common.Paginate).Get("/", GetImageSetsByID)
+	})
+	sub.Route("/view/{imageSetID}", func(r chi.Router) {
+		r.Use(ImageSetViewCtx)
+		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/", GetImageSetViewByID)
+		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/versions", GetImageSetImagesView)
+		r.Route("/versions/{imageID}", func(rVersion chi.Router) {
+			rVersion.Use(ImageSetImageViewCtx)
+			rVersion.Get("/", GetImageSetImageView)
+		})
 	})
 }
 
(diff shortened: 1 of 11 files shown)


pkg/routes/imagesets.go
```diff
@@ -402,3 +413,188 @@ func GetImageSetsView(w http.ResponseWriter, r *http.Request) {
...
+		imageID, err := strconv.Atoi(imageIDString)
+		if err != nil {
+			ctxServices.Log.WithField("error", err.Error()).Error("error while converting image id from string")
+			respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest("bad image id"))
+			return
+		}
+		var image models.Image
+		if result := db.Org(orgID, "").Where("image_set_id", imageSet.ID).First(&image, imageID); result.Error != nil {
+			if result.Error == gorm.ErrRecordNotFound {
+				respondWithAPIError(w, ctxServices.Log, errors.NewNotFound("image not found"))
```
AaronH88, line 545: Can we make this error a little more explicit so its easier to debug?

e.g. "image set for image set view not found"

pkg/routes/imagesets.go
```diff
@@ -402,3 +413,188 @@ func GetImageSetsView(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			apiError := errors.NewInternalServerError()
+			apiError.SetTitle("internal server error occurred while getting image")
```
AaronH88, line 549: "while getting image set"

pkg/services/images.go
```diff
@@ -1518,3 +1520,68 @@ func (s *ImageService) SetDevicesUpdateAvailabilityFromImageSet(orgID string, Im
...
+	if err != nil {
+		return 0, err
+	}
+
+	if tx == nil {
+		tx = db.DB
+	}
+
+	var count int64
+	result := db.OrgDB(orgID, tx, "").Debug().Model(&models.Image{}).Count(&count)
```
AaronH88, line 1536: why `.Debug()` ?

pkg/routes/imagesets.go
ldjebran: here we are getting image,
so maybe 
 "while getting image set image"

pkg/services/imagesets.go
```diff
@@ -146,32 +207,160 @@ func (s *ImageSetsService) GetImageSetsView(limit int, offset int, tx *gorm.DB)
...
-		installerID, ok := imageSetsInstallersMap[imageSetRow.ID]
-		if ok {
-			imageSetView.ImageBuildIsoURL = GetStorageInstallerIsoURL(installerID)
+		s.log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error("error getting image-set")
+		return nil, result.Error
+	}
+
+	var lastImage models.Image
+	// get the last image-set image
+	if result := db.Org(orgID, "").Debug().Order("created_at DESC").
```
AaronH88, line 265: Should `Debug()` be here?

I see why you ordered by `created_at` , should we be ordering by `version` instead? I dont think it makes a difference to the functionality so im ok to keep it this way, version just makes more sense in my head :smile:

pkg/services/imagesets.go
```diff
@@ -146,32 +207,160 @@ func (s *ImageSetsService) GetImageSetsView(limit int, offset int, tx *gorm.DB)
...
+		if result.Error == gorm.ErrRecordNotFound {
+			s.log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error("image-set last image not found")
+			return nil, new(ImageNotFoundError)
 		}
-		imageSetsView = append(imageSetsView, imageSetView)
+		s.log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error("error getting image-set last-image")
+		return nil, result.Error
+	}
+
+	imageInfo, err := imageService.AddPackageInfo(&lastImage)
```
AaronH88, line 281: Does the `Preload("Packages")` on line 266 not add this package info? 

Should we remove this preload from the DB query if we are going to call this function regardless?

pkg/services/imagesets.go
```diff
@@ -146,32 +207,160 @@ func (s *ImageSetsService) GetImageSetsView(limit int, offset int, tx *gorm.DB)
...
+		if result.Error == gorm.ErrRecordNotFound {
+			s.log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error("image-set not found")
+			return nil, new(ImageSetNotFoundError)
+		}
+		return nil, result.Error
+	}
+
+	var image models.Image
+	// get the image-set image
+	if result := db.Org(orgID, "").Debug().Order("created_at DESC").
```
AaronH88, line 331: Same comments as in the `GetImageSetViewByID` func :smile:

pkg/services/imagesets.go
ldjebran: sorting by created_at at this I am sure that I will have the latest first , but with version if value is missed or missed, this portion of code will have some troubles, that's why relaying on created_at is more robust here

pkg/services/imagesets.go
ldjebran: this function call is for one image, not like the old one that was calling this for all images.  AddPackageInfo  is different from Preload("Packages")  
AddPackageInfo will return the difference packages from last image 
```go
	imgDetail.UpdateAdded = len(upd[len(upd)-1].PackageDiff.Removed)
		imgDetail.UpdateRemoved = len(upd[len(upd)-1].PackageDiff.Added)
		imgDetail.UpdateUpdated = len(upd[len(upd)-1].PackageDiff.Upgraded)
```

pkg/services/imagesets.go
ldjebran: same response :smile:

pkg/services/imagesets.go
ldjebran: we are also able to sort by ID DESC , but do not like to sort by ids :smile:

pkg/services/images.go
ldjebran: removed all debug :smile:

pkg/services/imagesets.go
ldjebran: removed the debug

AaronH88: Sounds good to me, ty

AaronH88 (approved):
/lgtm

PR #1321, opened 2022-07-28
Dont Merge: POC Cypress

dmzoneill:
# Description

- Proof of concept to run cypress
- Need job in app-interface to build a frontend image for storage in quay.io

- https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44447
- https://github.com/RedHatInsights/edge-api/pull/1321
- https://github.com/RedHatInsights/edge-frontend/pull/516

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -19,8 +19,22 @@ source $CICD_ROOT/build.sh
 # Run the unit tests with an ephemeral db
 # source $APP_ROOT/unit_test.sh
 
+echo $(date -u) "*** To start deployment"
+source ${CICD_ROOT}/_common_deploy_logic.sh
+export NAMESPACE=$(bonfire namespace reserve)
+
+bonfire deploy \
+    ${APP_NAME} \
+    --source=appsre \
+    --ref-env insights-production \
+    --set-template-ref ${COMPONENT_NAME}=${GIT_COMMIT} \
+    --set-image-tag $IMAGE=$IMAGE_TAG \
+    --namespace ${NAMESPACE} \
+    --timeout 900 --optional-deps-method hybrid
+# END WORKAROUND
+
 # Deploy edge to an ephemeral namespace for testing
-source $CICD_ROOT/deploy_ephemeral_env.sh
+# source $CICD_ROOT/deploy_ephemeral_env.sh
 
 # This code is to create a 'dummy' result file so Jenkins will not fail when smoke tests are disabled
 #mkdir -p $ARTIFACTS_DIR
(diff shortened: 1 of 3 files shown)


pr_check.sh
```diff
@@ -19,8 +19,22 @@ source $CICD_ROOT/build.sh
...
+    --ref-env insights-production \
+    --set-template-ref ${COMPONENT_NAME}=${GIT_COMMIT} \
+    --set-image-tag $IMAGE=$IMAGE_TAG \
+    --namespace ${NAMESPACE} \
+    --timeout 900 --optional-deps-method hybrid
+# END WORKAROUND
+
 # Deploy edge to an ephemeral namespace for testing
-source $CICD_ROOT/deploy_ephemeral_env.sh
+# source $CICD_ROOT/deploy_ephemeral_env.sh
```
loadtheaccumulator, line 37: Does this break the existing checks, and if so, is this temporary to test Cypress?

loadtheaccumulator (changes requested):
I added the Do Not Merge label, so automation won't pick it up until ready on approval.

scripts/ephemeral/cypress.sh
```diff
@@ -0,0 +1,46 @@
...
+    sleep 10
+    RUNNING=$(oc get pod cypress | tail -n 1 | awk '{print $3}')
+done
+
+
+rm -rf /tmp/edge-frontend
+git clone --depth 1 --branch master https://github.com/RedHatInsights/edge-frontend.git /tmp/edge-frontend
+cd /tmp/edge-frontend
+# git fetch origin pull/$ghprbPullId/head:pr-$ghprbPullId
+# git checkout pr-$ghprbPullId
```
elyezer, line 28: Is this part needed to uncomment so it will run on PR checks with the right branch?

dmzoneill: There are X scenarios here.

Firstly, you need to recognize whether the PR is on "Frontend" or the "Backend".

If the PR is on the "Backend", then the basic scenario is to test using cypress from the HEAD of "frontend".
(this is why that's commented out)

If the PR is on the "frontend", then the basic scenario is to test using the PR commit of the "frontend", against of the HEAD of the "backend".

The third scenario, of course, is to use a PR from the FRONTEND and a PR from the BACKEND.
This is tricky.  As you need to provide the PR SHA1 of the opposite repo in order to have this.
We may look at this with the use of a comment or other in the commit message:

against: <SHA1>

pr_check.sh
loadtheaccumulator: Cool! Thanks!

PR #1335, opened 2022-08-01
devicesview filter-by-image-id fix and unit-tests

tomershinhar:
# Description

fixed the filter-by-image-id option for devicesview endpoint and added unit-tests to verify the query params available are accepted.

also added myself to CODEOWNERS

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/devices_test.go
+++ b/pkg/routes/devices_test.go
@@ -14,9 +14,11 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"github.com/redhatinsights/edge-api/pkg/db"
 	"github.com/redhatinsights/edge-api/pkg/dependencies"
 	"github.com/redhatinsights/edge-api/pkg/models"
 	"github.com/redhatinsights/edge-api/pkg/routes"
+	"github.com/redhatinsights/edge-api/pkg/routes/common"
 	"github.com/redhatinsights/edge-api/pkg/services"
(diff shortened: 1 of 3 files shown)


pkg/routes/devices_test.go
```diff
@@ -199,4 +200,76 @@ var _ = Describe("Devices View Router", func() {
...
+		BeforeEach(func() {
+			mockDeviceService.EXPECT().GetDevicesCount(gomock.Any()).Return(int64(0), nil)
+			mockDeviceService.EXPECT().GetDevicesView(gomock.Any(), gomock.Any(), gomock.Any()).Return(&models.DeviceViewList{}, nil)
+
+		})
+
+		When("filter by name", func() {
+			It("should return devices with given name", func() {
+				name := "device-1"
+				req, err := http.NewRequest("GET", fmt.Sprintf("/devices/devicesview?name=%s", name), nil)
```
ldjebran, line 213: how we can be sure that the filter was applied

ldjebran: I do not see that devices with name= "device-1" was returned, or filter was applied ,

same for other tests, may be an other strategy must applied  like  here https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/common/filters_test.go#L124-L142

ldjebran (approved):
LGTM

PR #1345, opened 2022-08-04
add account id to notification topic

acosferreira:
# Description

Notification process still need the account id to be processed, adding back and should be revisited when they complete the migration

Fixes # ([issue](https://issues.redhat.com/browse/THEEDGE-2462))

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -1459,6 +1459,7 @@ func (s *ImageService) SendImageNotification(i *models.Image) (ImageNotification
 		recipients = append(recipients, recipient)
 
 		notify.OrgID = i.OrgID
+		notify.Account = i.Account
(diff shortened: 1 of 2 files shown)


omaciel:
Tests?

mgold1234 (approved):
/lgtm

omaciel:
I really think we should have added tests before merging

acosferreira:
@omaciel we do have tests to the notification,we are validating if the event was sent.
Do you think we should also validate the object that was built to send the event?

omaciel:
> @omaciel we do have tests to the notification,we are validating if the event was sent.
> 
> Do you think we should also validate the object that was built to send the event?

The point of this PR was to add a missing field, so tests should be added to ensure that in the future we don't forget to send it.

PR #1379, opened 2022-08-16
update max resources

loadtheaccumulator:

# Description

lowering max resources for microservices

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -152,8 +152,8 @@ objects:
           value: ${TENANT_TRANSLATOR_PORT}
         resources:
           limits:
-            cpu: ${{CPU_LIMIT}}
-            memory: ${MEMORY_LIMIT}
+            cpu: 250m
+            memory: 2Gi
           requests:
             cpu: 125m
             memory: 128Mi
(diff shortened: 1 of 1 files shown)


acosferreira (approved):
lgtm

deploy/clowdapp.yaml
```diff
@@ -152,8 +152,8 @@ objects:
           value: ${TENANT_TRANSLATOR_PORT}
         resources:
           limits:
-            cpu: ${{CPU_LIMIT}}
-            memory: ${MEMORY_LIMIT}
+            cpu: 250m
+            memory: 2Gi
```
de1987, line 157: Would be better if we set up these values on app-interface?
I'm saying it because with this change we gonna lose the ability to set up prod with different values.
WDYT?

de1987 (changes requested):
Would be better if we set up these values on app-interface?
I'm saying it because with this change we gonna lose the ability to set up prod with different values.
WDYT?

de1987 (approved):

PR #1382, opened 2022-08-16
add retrycreateimage on case when worker stop responding

mgold1234:
related to- THEEDGE-2419

# Description
add retrycreateimage on case when worker stop responding in getComposeStatus under image builder side 

Fixes # (issue) THEEDGE-2419

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


loadtheaccumulator (changes requested):
RetryCreateImage starts at the beginning. Consider marking the image status as INTERRUPTED in the database and exit the Create or Update so the resume process can pick it up.

We use this line in the interrupt code to set it when the application receives and SIGTERM...
https://github.com/RedHatInsights/edge-api/blob/ca945f0cf65066123ddc5c44af521ba36cedcdd5/pkg/services/images.go#L596

pkg/services/images.go
```diff
@@ -871,16 +872,21 @@ func (s *ImageService) cleanFiles(kickstart string, isoName string, imageID uint
 func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, error) {
 	if image.Commit.Status == models.ImageStatusBuilding {
 		image, err := s.ImageBuilder.GetCommitStatus(image)
-		if err != nil {
+		if strings.Contains(err.Error() , "Worker running this job stopped responding") {
+				err := s.RetryCreateImage(image)
```
loadtheaccumulator, line 876: RetryCreateImage starts at the beginning. Consider marking the image status as INTERRUPTED in the database and exit the Create or Update so the resume process can pick it up.

We use this line in the interrupt code to set it when the application receives a SIGTERM...
https://github.com/RedHatInsights/edge-api/blob/ca945f0cf65066123ddc5c44af521ba36cedcdd5/pkg/services/images.go#L596

loadtheaccumulator (changes requested):
Meant to put my comment inline...

pkg/services/images.go
mgold1234: where I can add this row? before retry create image?

loadtheaccumulator: In place of retry create image.

loadtheaccumulator:
Changes look good so far. Just to need to catch anything else that might change the status to ERROR before it exits the build process.

loadtheaccumulator:
And it might be possible to refactor a bit and leverage the context to exit immediately.

pkg/services/images_test.go
```diff
@@ -687,6 +688,26 @@ var _ = Describe("Image Service Test", func() {
...
+		//requestID := faker.UUIDHyphenated()
+		imageName := faker.UUIDHyphenated()
+		image := models.Image{OrgID: orgID, Distribution: "rhel-85", Name: imageName, Commit: &models.Commit{
+			Status: models.ImageStatusBuilding}}
+		expectedErr := fmt.Errorf("running this job stopped responding")
+		When("When image-builder failed in GetComposeStatus when worker stopped responding", func() {
+			It("image is created with INTERRUPTED status", func() {
+				ctrl := gomock.NewController(GinkgoT())
+				defer ctrl.Finish()
+				mockImageService.EXPECT().UpdateImageStatus(&image).Return(&image, expectedErr)
```
ldjebran, line 702: we should not mock the image service

pkg/services/images_test.go
```diff
@@ -687,6 +688,26 @@ var _ = Describe("Image Service Test", func() {
...
+				mockImageBuilderClient.EXPECT().GetCommitStatus(&image).Return(&image, expectedErr)
+				_, err := service.UpdateImageStatus(&image)
+				Expect(err).To(HaveOccurred())
+				// image is not created
+				Expect(image.ID).To(Equal(uint(0)))
```
ldjebran, line 707: we should reload the image from the DB, and test it's current status

pkg/services/images.go
```diff
@@ -788,7 +795,14 @@ func (s *ImageService) cleanFiles(kickstart string, isoName string, imageID uint
 func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, error) {
 	if image.Commit.Status == models.ImageStatusBuilding {
 		image, err := s.ImageBuilder.GetCommitStatus(image)
-		if err != nil {
+		if strings.Contains(err.Error(), "running this job stopped responding") {
```
ldjebran, line 798: we are missing the logic that we should not put image in interrupted status if image life time exceed 3 hours.

pkg/services/images_test.go
```diff
@@ -687,6 +686,30 @@ var _ = Describe("Image Service Test", func() {
 			})
 		})
 	})
+	Describe("update image stop when worker running job stopped responding", func() {
+		orgID := faker.UUIDHyphenated()
+		//requestID := faker.UUIDHyphenated()
```
ldjebran, line 691: remove the comments

ldjebran (dismissed):
lgtm,  @loadtheaccumulator can you make a second pass.

pkg/services/images.go
```diff
@@ -789,6 +796,16 @@ func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, er
 	if image.Commit.Status == models.ImageStatusBuilding {
 		image, err := s.ImageBuilder.GetCommitStatus(image)
 		if err != nil {
+			// check that if error contain timeout and job stop responding and image's time creation is less than 3 hours
+			if strings.Contains(err.Error(), "running this job stopped responding") &&
```
ldjebran, line 800: are you sure we are receiving client error with this error text in it ?
as I saw the code we receive 200 even if there is a building error, the client error concern only client failure to connect to host, but in case of building failure we have to check the data received.

ldjebran: you have to begin here https://github.com/RedHatInsights/edge-api/blob/42659319ae79e0c0575cc99ba82f8b89f2c93dad/pkg/clients/imagebuilder/client.go#L311

analyse ComposeStatus  may be add reason field , you have to go to image builder and see the structure returned 
so that you can add the reason why it's failing

ldjebran: image builder ComposeStatus https://github.com/osbuild/image-builder/blob/main/internal/v1/api.yaml#L276
ComposeStatusError https://github.com/osbuild/image-builder/blob/main/internal/v1/api.yaml#L294

pkg/services/images.go
```diff
@@ -789,6 +796,16 @@ func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, er
...
+				time.Now().Sub(image.CreatedAt.Time).Hours() <= 3 {
```
AaronH88, line 801: @loadtheaccumulator does your interrupted code also check for a 3 hour timeout? Is this a duplication?

loadtheaccumulator: @AaronH88 yes. it handles stale interrupted builds.

AaronH88: OK, this is a duplication and is not needed here :smile:

AaronH88 (approved):
/lgtm

tomershinhar (approved):
LGTM

PR #1387, opened 2022-08-16
Fix slow query image set 2

acosferreira:
# Description

fix service layer

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/pkg/services/imagesets.go
+++ b/pkg/services/imagesets.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"fmt"
+
 	"gorm.io/gorm"
 
 	"github.com/redhatinsights/edge-api/pkg/db"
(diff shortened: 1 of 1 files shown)


de1987 (approved):

pkg/services/imagesets.go
```diff
@@ -98,8 +99,8 @@ func (s *ImageSetsService) GetImageSetsViewCount(tx *gorm.DB) (int64, error) {
 	var count int64
 
 	if result := db.OrgDB(orgID, tx, "image_sets").Debug().
-		Joins(`JOIN images ON image_sets.id = images.image_set_id AND Images.id = (Select Max(id) from Images where images.image_set_id = image_sets.id)`).
-		Model(&models.ImageSet{}).Count(&count); result.Error != nil {
+		Joins(`JOIN images ON image_sets.id = images.image_set_id`).
```
tpapaioa, line 102: If this query is just returning a count of image sets, then do we actually need the join on the images table here?

de1987: I made the same question @tpapaioa, the query is using JOIN to make sure we have at least one image created for that image set.
Some image sets don't have an image because the image set is created first and we might get an error from the image build

PR #1410, opened 2022-08-24
THEEDGE-2489:  implement update repo security.

ldjebran:
# Description

Fixes # THEEDGE-2489

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -1903,3 +1903,87 @@ paths:
               schema:
                 $ref: "#/components/schemas/v1.NotFound"
           description: installer not found.
+  /storage/update-repos/{updateTransactionID}/{repoFilePath}:
+    get:
+      operationId: GetDeviceRepoFile
+      parameters:
+        - in: cookie
+          name: device
+          required: true
+          description: a joined string combine a payload and signature of the payload
+          schema:
+            type: string
+            example: eyJvcmdfaWQiOiIxMTc4OTc3MiIsImRldmljZV91dWlkIjoiMjhhOWFjN2YtZjliYi00NTE4LTljMTMtNTNlNjMwZjAzYThmIiwidXBkYXRlX3RyYW5zYWN0aW9uX2lkIjoyfQ==::RRlcLYl_oeD6dyV0or59iud1150I227Q5u4s3eBrti8=
+        - name: updateTransactionID
+          in: path
+          required: true
+          description: Installer ID
+          schema:
+            type: integer
+            example: 120
+        - name: repoFilePath
+          in: path
+          required: true
+          description: the ostree repository file location
+          schema:
+            type: string
+            example: refs/heads/rhel/8/x86_64/edge
+      responses:
+        "200":
+          description: return the content of the repo file path
+          content:
+            application/octet-stream: {}
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/v1.BadRequest"
+          description: The request sent couldn't be processed.
+        "404":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/v1.NotFound"
+          description: repository file not found.
+  /storage/update-repos/{updateTransactionID}/content/{repoFilePath}:
+    get:
+      operationId: GetDeviceRepoFileContent
+      parameters:
+        - in: cookie
+          name: device
+          required: true
+          description: a joined string combine a payload and signature of the payload
+          schema:
+            type: string
+            example: eyJvcmdfaWQiOiIxMTc4OTc3MiIsImRldmljZV91dWlkIjoiMjhhOWFjN2YtZjliYi00NTE4LTljMTMtNTNlNjMwZjAzYThmIiwidXBkYXRlX3RyYW5zYWN0aW9uX2lkIjoyfQ==::RRlcLYl_oeD6dyV0or59iud1150I227Q5u4s3eBrti8=
+        - name: updateTransactionID
+          in: path
+          required: true
+          description: Installer ID
+          schema:
+            type: integer
+            example: 120
+        - name: repoFilePath
+          in: path
+          required: true
+          description: the ostree repository file location
+          schema:
+            type: string
+            example: refs/heads/rhel/8/x86_64/edge
+      responses:
+        "303":
+          description: Redirect to actual stored iso file
+          content:
+            application/octet-stream: {}
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/v1.BadRequest"
+          description: The request sent couldn't be processed.
+        "404":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/v1.NotFound"
+          description: repository file not found.
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -1905,6 +1905,7 @@ paths:
           description: installer not found.
   /storage/update-repos/{updateTransactionID}/{repoFilePath}:
     get:
+      summary: Return the update transaction repository content located at repoFilePath from root repository
       operationId: GetDeviceRepoFile
       parameters:
         - in: cookie
@@ -1917,7 +1918,7 @@ paths:
         - name: updateTransactionID
           in: path
           required: true
-          description: Installer ID
+          description: Update Transaction ID
           schema:
             type: integer
             example: 120
@@ -1947,6 +1948,7 @@ paths:
           description: repository file not found.
   /storage/update-repos/{updateTransactionID}/content/{repoFilePath}:
     get:
+      summary: Redirect to signed URL of the update transaction repository content located at repoFilePath from root repository
       operationId: GetDeviceRepoFileContent
       parameters:
         - in: cookie
@@ -1959,7 +1961,7 @@ paths:
         - name: updateTransactionID
           in: path
           required: true
-          description: Installer ID
+          description: Update Transaction ID
           schema:
             type: integer
             example: 120
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -1908,8 +1908,8 @@ paths:
       summary: Return the update transaction repository content located at repoFilePath from root repository
       operationId: GetDeviceRepoFile
       parameters:
-        - in: cookie
-          name: device
+        - name: device
+          in: cookie
           required: true
           description: a joined string combine a payload and signature of the payload
           schema:
@@ -1951,8 +1951,8 @@ paths:
       summary: Redirect to signed URL of the update transaction repository content located at repoFilePath from root repository
       operationId: GetDeviceRepoFileContent
       parameters:
-        - in: cookie
-          name: device
+        - name: device
+          in: cookie
           required: true
           description: a joined string combine a payload and signature of the payload
           schema:
--- a/main.go
+++ b/main.go
@@ -82,6 +82,8 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		http.ServeFile(w, r, cfg.OpenAPIFilePath)
 	})
 
+	route.Route("/api/edge/v1/storage/update-repos", routes.MakeStorageUpdateReposRouter)
+
 	// Authenticated routes
 	authRoute := route.Group(nil)
 	if cfg.Auth {
--- a/pkg/models/updates.go
+++ b/pkg/models/updates.go
@@ -8,6 +8,9 @@ import (
 	"gorm.io/gorm"
 )
 
+// ErrUpdateTransactionEmptyUUID returned when trying to use an update transaction with uuid not initiated
+var ErrUpdateTransactionEmptyUUID = errors.New("error update transaction uuid is empty")
+
 // UpdateTransaction represents the combination of an OSTree commit and a set of Inventory
 // hosts that need to have the commit deployed to them
 // This will ultimately kick off a transaction where the old version(s) of
--- a/pkg/models/updates.go
+++ b/pkg/models/updates.go
@@ -1,10 +1,9 @@
-// FIXME: golangci-lint
-// nolint:govet,revive,staticcheck
 package models
 
 import (
 	"errors"
 
+	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 )
@@ -30,6 +29,7 @@ type UpdateTransaction struct {
 	Repo            *Repo            `json:"Repo"`
 	ChangesRefs     bool             `gorm:"default:false" json:"ChangesRefs"`
 	DispatchRecords []DispatchRecord `gorm:"many2many:updatetransaction_dispatchrecords;save_association:false" json:"DispatchRecords"`
+	UUID            uuid.UUID        `gorm:"type:uuid" json:"UUID"`
 }
 
 // DispatchRecord represents the combination of a Playbook Dispatcher (https://github.com/RedHatInsights/playbook-dispatcher),
@@ -105,6 +105,9 @@ func (ur *UpdateTransaction) BeforeCreate(tx *gorm.DB) error {
 		log.Error("update-transaction do not have an org_id")
 		return ErrOrgIDIsMandatory
 	}
-
+	if ur.UUID == uuid.Nil {
+		// if not initiated generate a new uuid
+		ur.UUID = uuid.New()
+	}
 	return nil
 }
--- a/pkg/routes/signature/signature.go
+++ b/pkg/routes/signature/signature.go
@@ -7,9 +7,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
-	"github.com/redhatinsights/edge-api/pkg/models"
 	"strings"
+
+	"github.com/redhatinsights/edge-api/pkg/models"
+
+	"github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
 )
 
 const joinSeparator = "::"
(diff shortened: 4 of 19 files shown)


pkg/models/updates.go
```diff
@@ -29,6 +30,7 @@ type UpdateTransaction struct {
 	Repo            *Repo            `json:"Repo"`
 	ChangesRefs     bool             `gorm:"default:false" json:"ChangesRefs"`
 	DispatchRecords []DispatchRecord `gorm:"many2many:updatetransaction_dispatchrecords;" json:"DispatchRecords"`
+	UUID            uuid.UUID        `gorm:"type:uuid" json:"UUID"`
```
AaronH88, line 33: Why do we need a new id for the `UpdateTransaction`? Any reason why we can not reuse the `id` in `Model.ID` ?

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+var ErrInvalidDataAndSignatureString = errors.New("invalid data and signature string")
+
+// ErrSignatureValidationFailure returned when validation error occur
+var ErrSignatureValidationFailure = errors.New("signature validation failure")
+
+// ErrSignatureKeyCannotBeEmpty return when trying to create signature or validate with an empty key
+var ErrSignatureKeyCannotBeEmpty = errors.New("signature key cannot be empty")
+
+// ErrUpdateTransactionEmptyUUID returned when trying to use an update transaction with uuid not initiated
+var ErrUpdateTransactionEmptyUUID = errors.New("error update transaction uuid is empty")
```
AaronH88, line 27: If this is needed, this error should be moved to a common place, as update transaction is used in more places than just here.

The rest of the errors can remain in this file as they are file specific :smile:

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+	h.Write(data)
+	return h.Sum(nil)
+}
+
+func urlEncode(data []byte) string {
+	return base64.URLEncoding.EncodeToString(data)
+}
+
+func urlDecode(data string) ([]byte, error) {
+	return base64.URLEncoding.DecodeString(data)
```
AaronH88, line 51: `DecodeString()` returns `([]byte, error)` . Any reason we ignore a possible `error` ?

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// PackDataAndSignature return a joined string, formed from the data payload and signature
+func PackDataAndSignature(data []byte, signature string) string {
+	return fmt.Sprintf("%s%s%s", urlEncode(data), joinSeparator, signature)
+}
+
+// UnpackDataAndSignature split a value to data and signature, from a unified value string
+func UnpackDataAndSignature(value string) ([]byte, string, error) {
+	if len(value) == 0 ||
```
AaronH88, line 69: Can you add an example of a valid signature as a comment here please?

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+	values := strings.Split(value, joinSeparator)
+	if len(values) != 2 {
+		return nil, "", ErrInvalidDataAndSignatureString
+	}
+	data, err := urlDecode(values[0])
+	if err != nil {
+		return nil, "", err
+	}
+	// return data, signature
+	return data, values[1], nil
```
AaronH88, line 84: What is saved in `values[1]` ? Can you expand the function comment on line 67 to explain all three return vals, as I only see two explained :smile:

pkg/models/updates.go
ldjebran: we need to use a string that will be used as the last part of the signature key, a uuid is an adequate solution as has an adequate length

pkg/routes/signature/signature.go
ldjebran: signature is a module,  Really prefer to have this error declared here for this context, but if you insist :smile:  we can  moved some where.

pkg/routes/signature/signature.go
ldjebran: as in the comment upper line 83 the values[1]  is a signature string

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// EncodeSignedPayloadValue create a one string from a data payload and a signature of the data payload using the key
+func EncodeSignedPayloadValue(key []byte, data interface{}) (string, error) {
+	if len(key) == 0 {
+		return "", ErrSignatureKeyCannotBeEmpty
+	}
+	dataBytes, err := json.Marshal(data)
```
AaronH88, line 92: Can `data` be cast to a concrete type before `Marshal` ? Sometimes `json.Marshal` dose unexpected things, e.g. Angle Brackets and ampersands are escaped, and decoding a JSON number to an `interface` yields a `float64`

main.go
```diff
@@ -79,6 +79,8 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		http.ServeFile(w, r, cfg.OpenAPIFilePath)
 	})
 
+	route.Route("/api/edge/v1/storage/update-repos", routes.MakeStorageUpdateReposRouter)
```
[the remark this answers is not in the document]
ldjebran: storage/update-repos does not use the standard authentication, but handle a specific auth relying on a specific cookie for this end point .

pkg/routes/signature/signature.go
ldjebran: json.Marshal by it self take an interface as arg by it self,  what concrete type you have in mind when looking at the whole module.

ldjebran: take in consideration that we cannot create an encode function for each payload type, this will be a nightmare.
take into consideration I have not noticed any unmarshal failures when supplying the right dataReceiver or struct

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+		return ErrSignatureKeyCannotBeEmpty
+	}
+	data, signature, err := UnpackDataAndSignature(value)
+	if err != nil {
+		return err
+	}
+	if !ValidateSignature(key, data, signature) {
+		return ErrSignatureValidationFailure
+	}
+	if err := json.Unmarshal(data, &dataReceiver); err != nil {
```
AaronH88, line 112: As `dataRecieiver` is an interface, is this statment really validating the data?
Is it not just proving that `json.Unmarshal` can parse `data` to the default json data types?

https://pkg.go.dev/encoding/json#Unmarshal

Default json data types:

`bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null`

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+	if len(key) == 0 {
+		return "", ErrSignatureKeyCannotBeEmpty
+	}
+	if update.UUID == uuid.Nil {
+		return "", ErrUpdateTransactionEmptyUUID
+	}
+	// extend the key with update transaction uuid
+	key = append(key, []byte(update.UUID.String())...)
+
+	return EncodeSignedPayloadValue(key, data)
```
AaronH88, line 130: `EncodeSignedPayloadValue` will change `data.UpdateTransactionID` from type `uint` into type `float64` because json uses default data types for interfaces. `data` will be changed to an interface, and the default json type for numbers is `float64` 

Is this what you expect?

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+func DecodeUpdateTransactionCookieValue(key []byte, cookieValue string, update models.UpdateTransaction, dataReceiver *UpdateTransactionPayload) error {
+	if len(key) == 0 {
+		return ErrSignatureKeyCannotBeEmpty
+	}
+	if update.UUID == uuid.Nil {
+		return ErrUpdateTransactionEmptyUUID
+	}
+	key = append(key, []byte(update.UUID.String())...)
+
+	return DecodeSignedPayloadValue(key, cookieValue, dataReceiver)
```
AaronH88, line 144: Same comment as line 130 :smile:

pkg/routes/signature/signature.go
ldjebran: dataReceiver is a pointer to a concrete type

pkg/services/updates.go
```diff
@@ -293,14 +311,27 @@ func (s *UpdateService) WriteTemplate(templateInfo TemplateRemoteInfo, orgID str
 		envName = "dev"
 	}
 
+	url, err := url2.Parse(cfg.EdgeAPIBaseURL)
+	if err != nil {
+		s.log.WithField("error", err.Error()).Error("error while parsing config edge api url")
+		return "", err
+	}
+	remotePath := fmt.Sprintf("/api/edge/v1/storage/update-repos/%d", templateInfo.UpdateTransactionID)
+	repoURL := fmt.Sprintf("%s://%s%s", url.Scheme, url.Host, remotePath)
```
AaronH88, line 320: Should we version this url also?

ldjebran: no do not think so, response , next comment about versioning payload

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,145 @@
...
+var ErrSignatureValidationFailure = errors.New("signature validation failure")
+
+// ErrSignatureKeyCannotBeEmpty return when trying to create signature or validate with an empty key
+var ErrSignatureKeyCannotBeEmpty = errors.New("signature key cannot be empty")
+
+// ErrUpdateTransactionEmptyUUID returned when trying to use an update transaction with uuid not initiated
+var ErrUpdateTransactionEmptyUUID = errors.New("error update transaction uuid is empty")
+
+// UpdateTransactionPayload  The structure used to save device data in ostree remote cookie
+type UpdateTransactionPayload struct {
```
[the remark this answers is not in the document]
ldjebran: we decided offline that may a user story for future implementation. we consider the current implementation as version 0
a version in url storage/update-repos/v/{version}/[update-transaction-id]
the payload has to upgraded with version 
do a two steps unmarshaling first one reading version from the payload 
the second unmarshaling the versioned data
return version, data as interface
the caller of the function has to cast to the needed data type

```go
type PayloadVesrion struct {
    Version int `json:version`
}


type MyPayload struct {
     PayloadVesrion
     ....
}
```

pkg/routes/signature/signature.go
ldjebran: resolved in a discussion offline, as we have flat Playloads structure

pkg/routes/signature/signature.go
ldjebran: resolved in a discussion offline, as we have flat Playloads structure

pkg/routes/signature/signature.go
AaronH88: Im easy, lets see what the second reviewer thinks :smile:

pkg/routes/signature/signature.go
ldjebran: done

AaronH88 (approved):
/lgtm

+1

cmd/spec/path.yaml
```diff
@@ -1878,3 +1878,87 @@ paths:
...
+          name: device
+          required: true
+          description: a joined string combine a payload and signature of the payload
+          schema:
+            type: string
+            example: eyJvcmdfaWQiOiIxMTc4OTc3MiIsImRldmljZV91dWlkIjoiMjhhOWFjN2YtZjliYi00NTE4LTljMTMtNTNlNjMwZjAzYThmIiwidXBkYXRlX3RyYW5zYWN0aW9uX2lkIjoyfQ==::RRlcLYl_oeD6dyV0or59iud1150I227Q5u4s3eBrti8=
+        - name: updateTransactionID
+          in: path
+          required: true
+          description: Installer ID
```
de1987, line 1895: is it an update id or installer id?
it's kind confuse

cmd/spec/path.yaml
```diff
@@ -1878,3 +1878,87 @@ paths:
               schema:
                 $ref: "#/components/schemas/v1.NotFound"
           description: installer not found.
+  /storage/update-repos/{updateTransactionID}/{repoFilePath}:
```
de1987, line 1881: I think if you add a `summary` for the new endpoints it might be useful for other developers.
WDYT?

pkg/routes/signature/signature.go
de1987: I agree with @AaronH88, since those errors will be used in multiple places it would be better if moved to a common place.
It's a module but should be outside of the routes folder.

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, "", ErrInvalidDataAndSignatureString
+	}
+	values := strings.Split(value, joinSeparator)
+	if len(values) != 2 {
+		return nil, "", ErrInvalidDataAndSignatureString
+	}
+	data, err := urlDecode(values[0])
+	if err != nil {
+		return nil, "", err
+	}
```
de1987, line 85: would be great to see some logs in these errors. WDYT?

pkg/routes/signature/signature_test.go
```diff
@@ -0,0 +1,190 @@
...
+			err = signature.DecodeSignedPayloadValue(key, string(dataPayload)+"."+signatureString, &resultDeviceData)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(signature.ErrInvalidDataAndSignatureString))
+		})
+		Context("UpdateTransaction Cookie Value", func() {
+			var cookieValue string
+			updateTransaction := models.UpdateTransaction{
+				OrgID: orgID,
+				//Devices: []models.Device{},
+				//Repo:    &models.Repo{URL: "https://repo-storage.org/path/to/bucket", Status: models.ImageStatusSuccess},
```
de1987, line 144: are you hoping to use these?

pkg/routes/storage.go
```diff
@@ -116,3 +136,241 @@ func GetInstallerIsoStorageContent(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		updateTransactionID, err := strconv.Atoi(updateIDString)
+		if err != nil {
+			respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest("update transaction id must be an integer"))
+			return
+		}
+
+		var updateTransaction models.UpdateTransaction
+		if result := db.DB.Preload("Repo").First(&updateTransaction, updateTransactionID); result.Error != nil {
```
de1987, line 157: should we check if the update transaction belongs to the user's org id?

de1987:
On overall LGTM! 
I left some suggestions if you could take look before.

pkg/routes/signature/signature.go
ldjebran: done, moved to models module, as other locations will raise circular import error.

cmd/spec/path.yaml
ldjebran: updated

pkg/routes/signature/signature.go
ldjebran: resolved

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,148 @@
...
+func DecodeUpdateTransactionCookieValue(key []byte, cookieValue string, update models.UpdateTransaction, dataReceiver *UpdateTransactionPayload) error {
+	if len(key) == 0 {
+		return ErrSignatureKeyCannotBeEmpty
+	}
+	if update.UUID == uuid.Nil {
+		return ErrUpdateTransactionEmptyUUID
+	}
+	key = append(key, []byte(update.UUID.String())...)
+
+	return DecodeSignedPayloadValue(key, cookieValue, dataReceiver)
```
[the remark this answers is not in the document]
ldjebran: resolved

ldjebran:
@AaronH88 @de1987 all comments addressed.

pkg/routes/signature/signature_test.go
ldjebran: removed

cmd/spec/path.yaml
ldjebran: summary added to both end points

ldjebran:
signature updated, playbook tests updated with signature,  do not merge label removed

AaronH88:
Lets wait for Djebran to get back to safely merge.

cmd/spec/path.yaml
```diff
@@ -1903,3 +1903,89 @@ paths:
...
+    get:
+      summary: Return the update transaction repository content located at repoFilePath from root repository
+      operationId: GetDeviceRepoFile
+      parameters:
+        - in: cookie
```
tomershinhar, line 1911: maybe `name` should be the first parameter? for consistency

pkg/routes/signature/signature.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, "", ErrInvalidDataAndSignatureString
+	}
+	values := strings.Split(value, joinSeparator)
+	if len(values) != 2 {
+		log.WithField("error", ErrInvalidDataAndSignatureString.Error()).Error("unpack value must have 2 items, the payload and signature")
+		return nil, "", ErrInvalidDataAndSignatureString
+	}
+	data, err := urlDecode(values[0])
+	if err != nil {
+		log.WithField("error", err.Error()).Error("payload data is valid for url decode")
```
tomershinhar, line 86:
```suggestion
		log.WithField("error", err.Error()).Error("payload data is invalid for url decode")
```

ldjebran:
put do not merge as the update is not working and need m, to be sure update is working before merge

ldjebran:
@AaronH88 @tomershinhar @de1987 all comments addressed, do not merge removed

AaronH88 (approved):
/lgtm

de1987 (approved):

PR #1428, opened 2022-08-29
THEEDGE-2527 Added updates and dispatcher records for device endpoints

de1987:
# Description
Added updates and dispatcher records to be returned for device endpoints.
This is needed for epic THEEDGE-23

Fixes # ([THEEDGE-2527](https://issues.redhat.com/browse/THEEDGE-2527))

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [x] Refactor

# Checklist:


pkg/models/devices.go
```diff
@@ -44,16 +44,17 @@ type DeviceViewList struct {
...
+	DeviceName        string               `json:"DeviceName"`
+	DeviceUUID        string               `json:"DeviceUUID"`
+	ImageID           uint                 `json:"ImageID"`
+	ImageName         string               `json:"ImageName"`
+	LastSeen          EdgeAPITime          `json:"LastSeen"`
+	UpdateAvailable   bool                 `json:"UpdateAvailable"`
+	Status            string               `json:"Status"`
+	ImageSetID        uint                 `json:"ImageSetID"`
+	DeviceGroups      []DeviceDeviceGroup  `json:"DeviceGroups"`
+	UpdateTransaction *[]UpdateTransaction `json:"UpdateTransaction"`
```
ldjebran, line 57: do not think this is a good idea, as we have created views to not return all the data, for performance reasons,

pkg/services/devices.go
```diff
@@ -667,7 +667,10 @@ func (s *DeviceService) GetDevicesView(limit int, offset int, tx *gorm.DB) (*mod
 
 	var storedDevices []models.Device
 	// search for all stored devices that are also in inventory
-	if res := db.OrgDB(orgID, tx, "").Limit(limit).Offset(offset).Preload("UpdateTransaction").Preload("DevicesGroups").Find(&storedDevices); res.Error != nil {
+	if res := db.OrgDB(orgID, tx, "").Limit(limit).Offset(offset).
+		Preload("UpdateTransaction").
+		Preload("UpdateTransaction.DispatchRecords").
+		Preload("DevicesGroups").Find(&storedDevices); res.Error != nil {
```
ldjebran, line 673: we created views to avoid preloading everything, as with time this will be a mess , and we will be obliged to create a view2 for performance

AaronH88:
I agree with Djebran,

This idea will have negative performance implications. The big O notation of this becomes O(n^m) where N is the number of devices and M is the number of transactions per device.
It is exponential. Another solution is needed that is linear. 

Feel free to reach out to myself and Djebran with exactly what the front end requires and we can help get that data in a linear fashion

pkg/models/devices.go
de1987: I agree with you. I will let this as a list, but I will be returning just the latest `update_transaction`.

pkg/services/devices.go
de1987: @ldjebran @AaronH88 I agree with you. I made the changes required, now it's preloading only the latest `update_transaction` and the same for `dispatch_records`.
If frontend needs to load all, we can create a new endpoint for them to fetch.

de1987: if it still requires improvements, please let me know, then we can jump into a call

ldjebran (commented):
I think we can add dispatcher_status (this will be in the spirit of view) to device view (this what I understood is needed from the UI) and set it here in this if block like the normal status https://github.com/RedHatInsights/edge-api/blob/main/pkg/services/devices.go#L725, 

if we forced return updateTransaction let's name it LatestUpdateTransaction *Updatetransaction and also set it in the upper if block

in that block the update transaction are sorted ascendingly and the latest is the latest of that slice , so may no need to add order in the preload for update transaction or we should keep the order by and remove the sort.

for GetLatestUpdateTransactionForDevice  it's OK for my point of view

 what do you think ?

ldjebran (commented):
Thank you @de1987 
LGTM 
is it ready to merged or we have to wait for any dependencies ?

ldjebran (approved):

PR #1455, opened 2022-09-07
Added the produce and consumer frame work for building ISO's.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


loadtheaccumulator (approved):
/LGTM

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,127 @@
...
+	if err != nil {
+		mslog.Error("Subscribing to topics failed")
+		os.Exit(1)
+	}
+
+	mslog.Info("ISO Microservice ready")
+
+	run := true
+	for run {
+		ev := c.Poll(100)
```
thom-at-redhat, line 73: should this be a parameter/constant?  While testing golangci-lint, it warned against "magic numbers": 'What is a magic number?
A magic number is a numeric literal that is not defined as a constant, but which may change, and therefore can be hard to update. It's considered a bad programming practice to use numbers directly in any source code without an explanation. It makes programs harder to read, understand, and maintain.'

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,127 @@
...
+			switch key {
+			case models.EventTypeEdgeImageISORequested:
+				crcEvent := &image.EventImageISORequestedBuildHandler{}
+
+				err = json.Unmarshal(e.Value, crcEvent)
+				if err != nil {
+					mslog.Error("Failed to unmarshal CRC ISO event")
+				}
+
+				mslog = mslog.WithField("event_id", crcEvent.ID)
```
thom-at-redhat, line 103: Should there be an information message and log level?

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,127 @@
...
+				ctx = image.ContextWithLogger(ctx, mslog)
+
+				// call the event's Consume method
+				go crcEvent.Consume(ctx)
+			default:
+				mslog.Trace("Record key is not recognized by ISO consumer")
```
thom-at-redhat, line 109: Should we display the key in the log?

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,127 @@
...
+		"ImageBuilderURL":          cfg.ImageBuilderConfig.URL,
+		"InventoryURL":             cfg.InventoryConfig.URL,
+		"PlaybookDispatcherConfig": cfg.PlaybookDispatcherConfig.URL,
+		"TemplatesPath":            cfg.TemplatesPath,
+		"DatabaseType":             cfg.Database.Type,
+		"DatabaseName":             cfg.Database.Name,
+		"EdgeAPIURL":               cfg.EdgeAPIBaseURL,
+		"EdgeAPIServiceHost":       cfg.EdgeAPIServiceHost,
+		"EdgeAPIServicePort":       cfg.EdgeAPIServicePort,
+	}).Info("Configuration Values:")
```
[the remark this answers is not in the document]
AaronH88: Sorry Thom, I dont understand,

Nothing is being hard coded, this is a print out of the MicroServices config at boot, all the strings are just field names for the print.

Happy to remove any fields you think need to be secure :smile:

pkg/services/images_iso/main.go
AaronH88: ill add an int now

pkg/services/images_iso/main.go
AaronH88: As I am overloading the log to always include the request id, I am also setting the log level to info unless otherwise specified

pkg/services/images_iso/main.go
AaronH88: The key could be nil, but yeah, I can add it, and just check for nil first before printing.

AaronH88: ill add

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,111 @@
...
+	config.Init()
+	cfg := config.Get()
+	var configValues map[string]interface{}
+	cfgBytes, _ := json.Marshal(cfg)
+	_ = json.Unmarshal(cfgBytes, &configValues)
+	log.WithFields(configValues).Info("Configuration Values")
+	db.InitDB()
+
+	if cfg.KafkaConfig.Brokers == nil {
+		os.Exit(1)
```
ldjebran, line 32: put a log why we are exiting

pkg/services/images_iso/main.go
```diff
@@ -1,20 +1,111 @@
...
+				mslog.WithField("headers", e.Headers).Debug("Headers received with the event")
+			}
+
+			switch key {
+			case models.EventTypeEdgeImageISORequested:
+				crcEvent := &image.EventImageISORequestedBuildHandler{}
+
+				err = json.Unmarshal(e.Value, crcEvent)
+				if err != nil {
+					mslog.Error("Failed to unmarshal CRC ISO event")
```
ldjebran, line 84: if unmarshall failed nay be should continue in the loop  as crcEvent is not valid ?

ldjebran: what to do with consumer , should we commit any way ?

ldjebran: may be we should break the switch , to not consume this event , the next step will be c.Commit

AaronH88: Great catch, adding continue now, I might have to commit before continuing

PR #1465, opened 2022-09-11
Port segmentio to confluent Kafka libs with SASL

loadtheaccumulator:
# Description

Porting segmentio Kafka libs to confluent libs.
Inherits SASL capability with this change.
Fixed header reads from previously reverted PR
(currently includes dev config update from unmerged PR)

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

# Checklist:

### DIFF
--- a/pkg/services/consumers.go
+++ b/pkg/services/consumers.go
@@ -6,9 +6,11 @@ import (
 	"sync"
 	"time"
 
+	kafkacommon "github.com/redhatinsights/edge-api/pkg/common/kafka"
+
+	"github.com/confluentinc/confluent-kafka-go/kafka"
 	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 
-	"github.com/segmentio/kafka-go"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -20,7 +22,7 @@ type ConsumerService interface {
 
 // KafkaConsumerService is the implementation of a consumer service based on Kafka topics
 type KafkaConsumerService struct {
-	Reader        *kafka.Reader
+	Reader        *kafka.Consumer
 	UpdateService UpdateServiceInterface
 	DeviceService DeviceServiceInterface
 	ImageService  ImageServiceInterface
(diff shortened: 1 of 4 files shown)


AaronH88 (approved):
/lgtm

I think we are over logging but we can change that later

pkg/services/consumers.go
```diff
@@ -60,120 +63,164 @@ func NewKafkaConsumerService(config *clowder.KafkaConfig, topic string) Consumer
...
+			// commit the Kafka offset
+			_, err = s.Reader.Commit()
+			if err != nil {
+				log.WithField("error", err).Error("Error committing offset after message")
+			}
+		case kafka.Error:
+			log.WithFields(log.Fields{"code": e.Code(), "error": e}).Error("Exiting ConsumePlatformInventoryEvents loop due to Kafka broker issue")
+			if e.Code() == kafka.ErrAllBrokersDown {
+				run = false
+				return errors.New("uh oh, caught an error due to kafka broker issue") // create an error in errors.go
```
Saurabh-Thakre, line 212: We can remove the comment and create an error in errors.go

pkg/services/consumers.go
```diff
@@ -60,120 +62,164 @@ func NewKafkaConsumerService(config *clowder.KafkaConfig, topic string) Consumer
...
-		m, err := s.Reader.ReadMessage(context.Background())
-		if err != nil {
-			log.WithFields(log.Fields{
-				"error": err.Error(),
-			}).Error("Error reading message from Kafka topic")
-			return err
+
+	run := true
+	for run {
+		cs := s.Reader.Poll(100)
```
acosferreira, line 104: can we move this value to a const to be reusable and also know why we should 100?!

loadtheaccumulator: We can, but need to plan that out so it can be treated as variable per consumer across all of our consumers. Next PR?

acosferreira (approved):
lgtm 
(const change will be done later)

PR #1500, opened 2022-09-21
rearranged commit validation for devices update

tomershinhar:
# Description

the validation for commit provided by the user checked that the commit matches the image-set before it checked that it actually exists, this PR changes it so that the existence of the commit is validated first

also added unit-test for the case where the commit-id provided dose not exist  

Fixes # (issue)
[THEEDGE-2623](https://issues.redhat.com/browse/THEEDGE-2623)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/updates.go
+++ b/pkg/routes/updates.go
@@ -181,6 +181,7 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) *[]models.UpdateTran
 		respondWithAPIError(w, ctxServices.Log, errors.NewNotFound("some devices where not found"))
 		return nil
 	}
+	var commit *models.Commit
 	if devicesUpdate.CommitID == 0 {
 		commitID, err := ctxServices.DeviceService.GetLatestCommitFromDevices(orgID, devicesUUID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/routes/updates.go
```diff
@@ -199,7 +200,27 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) *[]models.UpdateTran
...
+			ctxServices.Log.WithFields(log.Fields{
+				"error":    err.Error(),
+				"commitID": devicesUpdate.CommitID,
+			}).Error("Can't find latest commit for the devices")
+			respondWithAPIError(w, ctxServices.Log, errors.NewNotFound(fmt.Sprintf("No commit found for CommitID %d", devicesUpdate.CommitID)))
+			return nil
+		}
 	} else {
+		// validate if commit is valid before continue process
+		commit_tmp, err := ctxServices.CommitService.GetCommitByID(devicesUpdate.CommitID, orgID)
```
AaronH88, line 214: I understand why you created `commit_tmp` however I suggest re-using the `commit` var you created earlier.

You can do this by declaring `err` within this `else` scope.

The syntax will look something like:

`var err error
commit, err = ctxServices.CommitService.GetCommitByID(devicesUpdate.CommitID, orgID)`


The reason I am suggesting doing it this way around is because we can get rid of line 223 `commit = commit_tmp ` 

It is more likely that the commit will be their than not :smile:

tomershinhar: done,
Thanks!

AaronH88 (approved):
/lgtm

mgold1234 (approved):
/lgtm

pkg/routes/updates.go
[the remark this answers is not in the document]
mgold1234: ignore it, it's related to code before the fix,
I put a comment and forgot to press on save

PR #1516, opened 2022-09-27
Removed unnneeded logs.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


unleash/features/feature.go
```diff
@@ -95,9 +95,5 @@ func (ff *Flag) IsEnabled() bool {
 			"status": "enabled"}).Debug("Feature flag status")
 		return true
 	}
-
-	log.WithFields(log.Fields{"feature": ff.Name,
```
loadtheaccumulator, line 99: Any thoughts about switching both of these to trace level instead?

loadtheaccumulator (changes requested):
quick question...

unleash/features/feature.go
AaronH88: I dont think so, nether log gives any useful information for debugging.
The log here is triggered every time a feature flag is checked, and we have many feature flags now, we are drowning in these logs.

The log in consumers.go has no use at all, its not attached to an event, its not even attached to an edge event, the only use it had was to make sure we are connected to kafka, but we are well past that stage now.

de1987 (approved):

loadtheaccumulator (approved):

PR #1530, opened 2022-10-05
Added producer interface to allow mocking. Turned producer, topic and…

AaronH88:
… kafkaconfig map into a service. Added mocks for the new services. Added new tests for the producer service.

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [X] Refactor

# Checklist:


ldjebran (approved):

pkg/common/kafka/producer.go
```diff
@@ -14,10 +15,48 @@ import (
 
 var lock = &sync.Mutex{}
 
-var singleInstance *kafka.Producer
+var singleInstance Producer
+
+// Producer the an interface for a kafka producer, it matches the confluent producer type defination
```
de1987, line 20: typo
```suggestion
// Producer an interface for a kafka producer, it matches the confluent producer type definition
```

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+		config.Get().KafkaBrokers = brokerSlice
+
+	})
+	AfterEach(func() {
+		ctrl.Finish()
+		service.UnsetProducer()
+	})
+	Describe("Get Producer instance", func() {
+		When("One dosent exist yet", func() {
+			Context("One dosent exist yet", func() {
```
de1987, line 62: typo
```suggestion
		When("One doesnt exist yet", func() {
			Context("One doesnt exist yet", func() {
```

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+					p1 := service.GetProducerInstance()
+					Expect(p1).To(BeNil())
+				})
+			})
+		})
+	})
+	Describe("Produce Event", func() {
+		When("OK", func() {
+			Context("ok", func() {
+				It("should be ok", func() {
```
de1987, line 109: Would make sense to change those for `Context -> When -> It`, I think that's a better approach as presented by Gomega.
WDYT?

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+						ID:          faker.HyphenatedID,
+						RedHatOrgID: common.DefaultOrgID,
+						Source:      models.SourceEdgeEventAPI,
+						SpecVersion: "v1",
+						Subject:     image.Name,
+						Time:        time.Now().Format(time.RFC3339),
+						Type:        models.EventTypeEdgeImageRequested,
+					}
+					err := service.ProduceEvent(gomock.Any().String(), gomock.Any().String(), cloudEvent)
+					Expect(err).To(BeNil())
```
de1987, line 152: I would recommend adding a check for the value produced.
My suggestion, add spy for `producer.Produce` and check if the event message if correct after `json.Marshal` and if the `topic` is the same that is suppose to be used.
WDYT?

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+	Describe("Get Producer instance", func() {
+		When("One dosent exist yet", func() {
+			Context("One dosent exist yet", func() {
+				It("should be ok", func() {
+					kafkaConfigMap := kafka.ConfigMap{
+						"bootstrap.servers": "192.168.1.3:80",
+					}
+					mockKafkaConfigService.EXPECT().GetKafkaProducerConfigMap().Return(kafkaConfigMap)
+					p := service.GetProducerInstance()
```
mgold1234, line 68: can you please change it to a more meaningful name? like 'producer'

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+	AfterEach(func() {
+		ctrl.Finish()
+		service.UnsetProducer()
+	})
+	Describe("Get Producer instance", func() {
+		When("One dosent exist yet", func() {
+			Context("One dosent exist yet", func() {
+				It("should be ok", func() {
+					kafkaConfigMap := kafka.ConfigMap{
+						"bootstrap.servers": "192.168.1.3:80",
```
mgold1234, line 65: maybe define the "bootstrap.servers" in the BeforEach section and reuse it?

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+					}
+					mockKafkaConfigService.EXPECT().GetKafkaProducerConfigMap().Return(kafkaConfigMap2)
+					p1 := service.GetProducerInstance()
+					Expect(p1).To(BeNil())
+				})
+			})
+		})
+	})
+	Describe("Produce Event", func() {
+		When("OK", func() {
```
mgold1234, line 107: insted of 'OK' maybe write 'success to produce event '

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+					cloudEvent := models.CRCCloudEvent{
+						Data:        edgePayload,
+						DataSchema:  "v1",
+						ID:          faker.HyphenatedID,
+						RedHatOrgID: common.DefaultOrgID,
+						Source:      models.SourceEdgeEventAPI,
+						SpecVersion: "v1",
+						Subject:     image.Name,
+						Time:        time.Now().Format(time.RFC3339),
```
mgold1234, line 148: what is RFC3339?

pkg/common/kafka/producer_test.go
de1987: Or by checking if the consumer is getting the event.

de1987: I was thinking about how to mock this in Go, and unfortunately it requires to expose the `SingleInstance`, like this:
```
type ProducerService struct {
	Topic          TopicServiceInterface
	KafkaConfigMap KafkaConfigMapServiceInterface
	SingleInstance Producer
}
```

pkg/common/kafka/producer_test.go
```diff
@@ -0,0 +1,157 @@
...
+				})
+			})
+		})
+	})
+	Describe("Produce Event", func() {
+		When("OK", func() {
+			Context("ok", func() {
+				It("should be ok", func() {
+					topic := "Test Topic"
+					mockTopicService.EXPECT().GetTopic(gomock.Any()).Return(topic, nil)
```
de1987, line 111: the `GetTopic` parameter is `string`, I think would make sense to use the `topic` as expect value

pkg/common/kafka/producer_test.go
loadtheaccumulator: Suggesting "One does not exist yet"

pkg/common/kafka/producer_test.go
loadtheaccumulator: IMHO, the description provides the context for the "OK" and the extra wording would be redundant.

pkg/common/kafka/producer.go
AaronH88: Ty, changed

pkg/common/kafka/producer_test.go
AaronH88: Ty, changed to Jonathans suggestion

pkg/common/kafka/producer_test.go
AaronH88: Feel free to re-factor all our existing tests to follow your suggested model. I followed how all the other tests are done.

pkg/common/kafka/kafkaconfigmap.go
```diff
@@ -9,8 +10,23 @@ import (
...
+}
+
+// NewKafkaConfigMapService returns a new service
+func NewKafkaConfigMapService() KafkaConfigMapServiceInterface {
+	return &KafkaConfigMapService{}
+}
+
 // GetKafkaProducerConfigMap returns the correct kafka auth based on the environment and given config
-func GetKafkaProducerConfigMap() kafka.ConfigMap {
+func (k *KafkaConfigMapService) GetKafkaProducerConfigMap() kafka.ConfigMap {
```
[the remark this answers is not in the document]
AaronH88: The google search term you are looking for is something like "Adding Methods to Structs", here is an example answer: http://golangtutorials.blogspot.com/2011/06/methods-on-structs.html

pkg/common/kafka/producer_test.go
AaronH88: What else could p be? P is being assigned directly to the result of `GetProducerInstance()` and as a result you knew exactly that for this test p was a `producer`. 

I can change it to producer, but it is overly verbose and unneeded. (You knew exactly what it was)

pkg/common/kafka/producer_test.go
AaronH88: According to google is it a Date and Time format

pkg/common/kafka/producer_test.go
AaronH88: I agree with Jonathan. Nothing to change here

pkg/common/kafka/producer_test.go
AaronH88: I dont understand what you mean by "expect value"

`topic` is the value I wish my mock to return, I dont care what values are passed to my mock, just return `topic` when called.

pkg/common/kafka/producer_test.go
AaronH88: good idea :smile:

pkg/common/kafka/producer_test.go
AaronH88: Exposing `singleInstance` in the example you showed breaks the producer singleton I created. This is not what I want to do. 

What exactly are you asking me to test here? 
We dont test `json.Marshal`, we dont test the concrete kafka producer. 

The producer has already been mocked out, so we can force an error and ensure our code catches these errors, that will look like this: 
```
mockProducerService = mock_kafkacommon.NewMockProducerServiceInterface(ctrl)
mockProducerService.EXPECT().ProduceEvent(gomock.Any(), gomock.Any(), gomock.Any()).Return(ERROR)
```  

The main goal for my task was to write the framework, not write all test cases, then can be added in time :smile:

pkg/common/kafka/producer_test.go
mgold1234: good point

pkg/common/kafka/producer_test.go
mgold1234: maybe you can put this variable upper and reuse it?

AaronH88: I dont understand, what would you like as a variable?
`time` is an imported `type`, `time.RFC3339` is a type within the `time` library, the function `time.Now()` gets the current time, and then `.Format()` tells me what format I would like this time in. :smile:

pkg/common/kafka/producer_test.go
de1987: my suggestion was just to use it like it below because if somehow you sent a different  value than the topic it will fail
so you are guaranteed that value will be the topic, and also mocking it
`
```suggestion
					mockTopicService.EXPECT().GetTopic(topic).Return(topic, nil)
```

PR #1532, opened 2022-10-05
Add memory size check method to validate log's object size

adeliaferreira:
Signed-off-by: Adelia Ferreira <tay.figueira@gmail.com>

# Description
- Add new method to check object memory size to be used on validation of Log's object size
- Add unit test for new method

Fixes #2583 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


pkg/services/files/utils.go
```diff
@@ -17,3 +20,23 @@ func sanitizePath(destination string, filePath string) (destpath string, err err
...
 	return
 }
+
+func objectIsInsideMemoryLimit(object interface{}) bool {
+	objReflection := reflect.ValueOf(object)
+	len, _ := Sizeof(objReflection)
+	return len < CloudwatchLimitLogSize
+}
+
+func Sizeof(rv reflect.Value) (int, int) {
```
AaronH88, line 30: As this function is public it will need a comment explaining what it dose, in the format `// Sizeof ....`

pkg/services/files/utils.go
```diff
@@ -17,3 +20,23 @@ func sanitizePath(destination string, filePath string) (destpath string, err err
...
+	rt := rv.Type()
+
+	switch rt.Kind() {
+	case reflect.Slice:
```
AaronH88, line 34: Are strings treated as `slice` or should we add a new `case` for `case reflect.String` ?

pkg/services/files/utils.go
```diff
@@ -17,3 +20,23 @@ func sanitizePath(destination string, filePath string) (destpath string, err err
...
+func objectIsInsideMemoryLimit(object interface{}) bool {
+	objReflection := reflect.ValueOf(object)
+	len, _ := Sizeof(objReflection)
+	return len < CloudwatchLimitLogSize
+}
+
+func Sizeof(rv reflect.Value) (int, int) {
+	rt := rv.Type()
+
+	switch rt.Kind() {
```
AaronH88, line 33: Any default case for a type that isnt in this switch case?

pkg/services/files/utils_test.go
```diff
@@ -22,3 +22,17 @@ func TestSanitizeWithFilePathWithFolder(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestObjectIsInsideMemoryLimitPositiveReturn(t *testing.T) {
+	result := objectIsInsideMemoryLimit(make([][256]byte, 8, 16))
+	if !result {
+		t.Error("object was inside limite")
```
AaronH88, line 29: typo, `limit` :smile:

pkg/services/files/utils_test.go
```diff
@@ -22,3 +22,17 @@ func TestSanitizeWithFilePathWithFolder(t *testing.T) {
...
+	}
+}
+
+func TestObjectIsInsideMemoryLimitNegativeReturn(t *testing.T) {
+	result := objectIsInsideMemoryLimit(make([][271890]byte, 8, 16))
+	if result {
+		t.Error("object is outside limite")
```
AaronH88, line 36: typo

AaronH88:
Overall, very impressive first PR, well done! 

A little rework is needed for the switch case and it will be all good! :smile:

pkg/services/files/utils_test.go
de1987: I would suggest changing it to `object is within the limit`
WDYT?

pkg/services/files/utils_test.go
de1987: I would suggest changing it to `object is exceeding the limit`
WDYT?

de1987:
Congratulations on your first API PR!

I left a couple of suggestions, overall after Aaron's suggestion, and I think we are good.

pkg/services/files/utils.go
adeliaferreira: reflect.ValueOf inside objectIsWithinMemoryLimit function will made strings be handle on slice. Added a unitTest with string as test case.

pkg/services/files/utils.go
adeliaferreira: The method is expected the reflect.Value to be handle by reflect.slice on the case of map is supported and I added a unit Test for map. Anyway, I added a default case to throw a panic.

de1987 (approved):

AaronH88 (approved):
/lgtm

omaciel:
@AaronH88 @de1987 @adeliaferreira  I know I'm late to the party, but I was wondering why was `Sizeof` written to return `capacity of an object`--the second argument being returned--if we're not using it anywhere else?

PR #1535, opened 2022-10-06
update: [secure repo] Add update transaction repository storage end-points.

ldjebran:
Fixes # THEEDGE-2489

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/storage.go
+++ b/pkg/routes/storage.go
@@ -49,7 +49,7 @@ func MakeStorageRouter(sub chi.Router) {
 }
 
 // redirectToStorageSignedURL redirect request to real content storage url using a signed url
-func redirectToStorageSignedURL(w http.ResponseWriter, r *http.Request, path string) error {
+func redirectToStorageSignedURL(w http.ResponseWriter, r *http.Request, path string) {
 	ctxServices := dependencies.ServicesFromContext(r.Context())
 	logContext := ctxServices.Log.WithField("service", "device-repository-storage")
 	signedURL, err := ctxServices.FilesService.GetSignedURL(path)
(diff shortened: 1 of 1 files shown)


pkg/routes/storage_test.go
```diff
@@ -128,4 +128,319 @@ var _ = Describe("Storage Router", func() {
...
+
+				url, err := url2.Parse(updateTransaction.Repo.URL)
+				Expect(err).ToNot(HaveOccurred())
+				targetPath := fmt.Sprintf("%s/%s", url.Path, targetRepoFile)
+
+				fileContentReader := strings.NewReader(fileContent)
+				fileContentReadCloser := io.NopCloser(fileContentReader)
+				mockFilesService.EXPECT().GetFile(targetPath).Return(fileContentReadCloser, nil)
+
+				rr := httptest.NewRecorder()
```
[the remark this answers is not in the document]
ldjebran: renamed to httpTestRecorder

mgold1234 (approved):
/lgtm

pkg/routes/storage.go
```diff
@@ -118,3 +186,144 @@ func GetInstallerIsoStorageContent(w http.ResponseWriter, r *http.Request) {
...
+	updateTransaction := getContextStorageUpdateTransaction(w, r)
+	if updateTransaction == nil {
+		return
+	}
+
+	requestPath := ValidateStorageUpdateTransaction(w, r)
+	if requestPath == "" {
+		// ValidateStorageUpdateTransaction will handle errors
+		return
+	}
```
de1987, line 294: for trace tracking purposes, could we add logs for those? I think if some day we had an issue related to that piece of code, it will be easier to identify.
wdut?

pkg/routes/storage.go
```diff
@@ -118,3 +186,144 @@ func GetInstallerIsoStorageContent(w http.ResponseWriter, r *http.Request) {
...
+
+	logContext = logContext.WithFields(log.Fields{
+		"orgID":               updateTransaction.OrgID,
+		"updateTransactionID": updateTransaction.ID,
+		"path":                requestPath,
+	})
+	logContext.Info("return storage update transaction repo resource content")
+	_ = serveStorageContent(w, r, requestPath)
+	// serveStorageContent will respond with proper response on failure and handle logging
+}
```
de1987, line 329: if the return is not important, could we change those functions to not return the `error` or `nil`.
Then, we can drop the comments and `_ =`

pkg/routes/storage.go
ldjebran: But that function is already logging on failures

pkg/routes/storage.go
ldjebran: ok

pkg/routes/storage.go
de1987: awesome, if there is stack trace linking to that function call, I think we are good

PR #1559, opened 2022-10-12
update: FIX update change refs variable.

ldjebran:
The current problem is that when a dist 8.5 and we update to 8.6 the change refs is True and the playbook will fail as the ref has not changed.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/updates.go
+++ b/pkg/services/updates.go
@@ -785,7 +785,7 @@ func (s *UpdateService) BuildUpdateTransactions(devicesUpdate *models.DevicesUpd
 					if uError != nil {
 						s.log.WithField("error", err.Error()).Error("Error returning current image ostree checksum")
 					}
-					if currentImage.Distribution != updatedImage.Distribution {
+					if config.DistributionsRefs[currentImage.Distribution] != config.DistributionsRefs[updatedImage.Distribution] {
 						update.ChangesRefs = true
 					}
 				}
--- a/pkg/services/files/utils.go
+++ b/pkg/services/files/utils.go
@@ -21,14 +21,14 @@ func sanitizePath(destination string, filePath string) (destpath string, err err
 	return
 }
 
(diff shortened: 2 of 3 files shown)


AaronH88 (approved):
/lgtm

de1987 (approved):

omaciel:
Please add unittests for this before merging

omaciel:
Added donotmerge label

omaciel:
Removed donotmerge label

pkg/services/updates.go
```diff
@@ -785,7 +785,7 @@ func (s *UpdateService) BuildUpdateTransactions(devicesUpdate *models.DevicesUpd
 					if uError != nil {
 						s.log.WithField("error", err.Error()).Error("Error returning current image ostree checksum")
 					}
-					if currentImage.Distribution != updatedImage.Distribution {
+					if config.DistributionsRefs[currentImage.Distribution] != config.DistributionsRefs[updatedImage.Distribution] {
```
loadtheaccumulator, line 788: In looking at the tests post-merge, I think this should be a func changesRefs(current string, updated string) bool {} and then simple functional in/out tests for each scenario
- minor to minor (same version e.g., 8.4 to 8.4)
- minor to minor (different version e.g., 8.4 to 8.5)
- minor to major (e.g., 8.5 to 9.0)

Mocking the service(s) becomes unnecessary for this simple test.

loadtheaccumulator (commented):
Leaving a note here for future reference and will bring it up separately.

pkg/services/updates.go
loadtheaccumulator: @omaciel @ldjebran @AaronH88 catching this post merge, so maybe can be handled in a refactor or just for future reference.

PR #1582, opened 2022-10-18
Refactored SendNotification function to use the new kafka producer wi…

AaronH88:
…th auth. Refactored the Update Service to use the kafka Produce Service. Modified the unit tests to allow for kafka mocking.

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


pkg/services/updates.go
```diff
@@ -480,81 +482,41 @@ func (s *UpdateService) SetUpdateStatus(update *models.UpdateTransaction) error
...
+	recordKey := "ImageCreationStarts"
+	recordValue, _ := json.Marshal(notify)
+
+	// send the message
+	p := s.ProducerService.GetProducerInstance()
+	perr := p.Produce(&kafka.Message{
+		TopicPartition: kafka.TopicPartition{Topic: &topic, Partition: kafka.PartitionAny},
+		Key:            []byte(recordKey),
+		Value:          []byte(recordValue),
+	}, nil)
```
de1987, line 512: since this code might be repeated too much because of EDA and to avoid the comments
it could be refactored into a function, for example
`p.ProducerService.SendKakfaMessage`

WDYT?

loadtheaccumulator: This producer is existing code caught in the diff, so we should leave this as-is and design notifications app-wide in a separate epic/story.

loadtheaccumulator (approved):

pkg/services/updates.go
AaronH88: @de1987 The function your talking about already exists, it is called `ProduceEvent()` and you will find it in `pkg/common/kafka/producer.go` 

`ProduceEvent()` is not appropriate for this solution as here we are producing an external event. `ProduceEvent()` uses our internal event schema, and as I am calling an external service, it dosent fit.

Hence, I called the normal `Produce()`

de1987 (approved):

PR #1606, opened 2022-10-25
fix validation against commit and images, add scenario to prevent rol…

acosferreira:
# Description

Fix the validation under commitId and images
Prevent rollback on update when pass commitId
Add tests 

Fixes # (THEEDGE-2654)

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/updates.go
+++ b/pkg/routes/updates.go
@@ -226,7 +226,7 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) *[]models.UpdateTran
 			ctxServices.Log.WithFields(log.Fields{
 				"error":    err.Error(),
 				"commitID": devicesUpdate.CommitID,
-			}).Error("Commit does not belong to the same image-set as devices")
+			}).Error(err.Error())
 			respondWithAPIError(w, ctxServices.Log, errors.NewBadRequest(fmt.Sprintf("Commit %d does not belong to the same image-set as devices", devicesUpdate.CommitID)))
 			return nil
 		}
--- a/pkg/routes/updates_test.go
+++ b/pkg/routes/updates_test.go
@@ -503,6 +503,135 @@ var _ = Describe("Update routes", func() {
 		})
 	})
 
+	Context("Devices update by diff commits", func() {
+		var edgeAPIServices *dependencies.EdgeAPIServices
+		var mockUpdateService *mock_services.MockUpdateServiceInterface
+		var ctrl *gomock.Controller
+
+		orgID := common.DefaultOrgID
+
+		imageSet := models.ImageSet{
+			OrgID: orgID,
+		}
+		db.DB.Create(&imageSet)
+
+		commits := []models.Commit{{OrgID: orgID, Name: "1"},
+			{OrgID: orgID, Name: "2"},
+			{OrgID: orgID, Name: "3"},
+			{OrgID: orgID, Name: "4"},
+			{OrgID: orgID, Name: "5"}}
+		db.DB.Debug().Create(&commits)
+
+		images := [5]models.Image{
+			{OrgID: orgID, CommitID: commits[0].ID, Status: models.ImageStatusSuccess, Version: 1, ImageSetID: &imageSet.ID},
+			{OrgID: orgID, CommitID: commits[1].ID, Status: models.ImageStatusSuccess, Version: 2, ImageSetID: &imageSet.ID},
+			{OrgID: orgID, CommitID: commits[2].ID, Status: models.ImageStatusSuccess, Version: 3, ImageSetID: &imageSet.ID},
+			{OrgID: orgID, CommitID: commits[3].ID, Status: models.ImageStatusSuccess, Version: 4, ImageSetID: &imageSet.ID},
+			{OrgID: orgID, CommitID: commits[4].ID, Status: models.ImageStatusSuccess, Version: 5, ImageSetID: &imageSet.ID},
+		}
+		db.DB.Debug().Create(&images)
+
+		device := models.Device{
+			OrgID:   orgID,
+			UUID:    faker.UUIDHyphenated(),
+			ImageID: images[1].ID,
+		}
+		db.DB.Create(&device)
+		BeforeEach(func() {
+			ctrl = gomock.NewController(GinkgoT())
+			mockUpdateService = mock_services.NewMockUpdateServiceInterface(ctrl)
+			ctx := context.Background()
+			logger := log.NewEntry(log.StandardLogger())
+
+			edgeAPIServices = &dependencies.EdgeAPIServices{
+				DeviceService: services.NewDeviceService(ctx, logger),
+				CommitService: services.NewCommitService(ctx, logger),
+				UpdateService: mockUpdateService,
+				Log:           logger,
+			}
+		})
+		AfterEach(func() {
+			ctrl.Finish()
+		})
+
+		When("when try to update a device to specific image", func() {
+
+			It("should not allow to update", func() {
+				updateData, err := json.Marshal(models.DevicesUpdate{DevicesUUID: []string{device.UUID}, CommitID: commits[0].ID})
+				Expect(err).To(BeNil())
+				req, err := http.NewRequest(http.MethodPost, "/", bytes.NewBuffer(updateData))
+				Expect(err).To(BeNil())
+
+				ctx := dependencies.ContextWithServices(req.Context(), edgeAPIServices)
+				req = req.WithContext(ctx)
+
+				rr := httptest.NewRecorder()
+				handler := http.HandlerFunc(AddUpdate)
+				handler.ServeHTTP(rr, req)
+
+				var response common.APIResponse
+				respBody, err := ioutil.ReadAll(rr.Body)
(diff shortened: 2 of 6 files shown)


pkg/services/commits.go
```diff
@@ -105,6 +105,9 @@ func (s *CommitService) ValidateDevicesImageSetWithCommit(devicesUUID []string,
 		return result.Error
 	}
 
+	if imageSetDevices.Commit_Id > commitID {
```
omaciel, line 108: Unittest please

omaciel (changes requested):
Add unittest for conditional statement please

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,135 @@ var _ = Describe("Update routes", func() {
...
+				Log:           logger,
+			}
+		})
+		AfterEach(func() {
+			ctrl.Finish()
+		})
+
+		When("when try to update a device to specific image", func() {
+
+			It("should not allow to update", func() {
```
acosferreira, line 559: @omaciel follow the test for the scenario with minor commit, I should not be able to update

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,135 @@ var _ = Describe("Update routes", func() {
...
+				handler := http.HandlerFunc(AddUpdate)
+				handler.ServeHTTP(rr, req)
+
+				var response common.APIResponse
+				respBody, err := ioutil.ReadAll(rr.Body)
+				err = json.Unmarshal(respBody, &response)
+
+				Expect(rr.Code).To(Equal(http.StatusOK))
+				Expect(err).Should(BeNil())
+				// Expect(response.Message).To(Equal("There are no updates to perform"))
```
AaronH88, line 628: Is checking the response message needed?

pkg/services/commits.go
```diff
@@ -72,10 +73,10 @@ func (s *CommitService) ValidateDevicesImageSetWithCommit(devicesUUID []string,
 		return err
 	}
 
-	if result := db.Org(orgID, "devices").Table("devices").
-		Select(`images.image_set_id as "image_set_id", count(devices.id) as devices_count`).
-		Joins("JOIN images ON devices.image_id = images.id").
-		Where("devices.uuid in (?)", devicesUUID).
+	if result := db.Org(orgID, "devices").Table("devices").Debug().
```
AaronH88, line 76: Is `Debug()` needed? Are we looking for this gorm log?

pkg/services/commits.go
```diff
@@ -72,10 +73,10 @@ func (s *CommitService) ValidateDevicesImageSetWithCommit(devicesUUID []string,
...
+		Select(`images.image_set_id as "image_set_id", images.commit_Id , count(devices.uuid) as devices_count`).
+		Joins("JOIN images ON devices.image_id = images.id join COMMITS on commits.ID = images.commit_id").
```
AaronH88, line 78: We seem to be mixing gorm syntax and SQL syntax here, It looks like the gorm way to do this is something like : 
`Joins("JOIN images ON devices.image_id = images.id).Joins(JOIN COMMITS on commits.ID = images.commit_id").`

I suggest keeping one syntax for readability :smile:

pkg/services/commits.go
```diff
@@ -72,10 +73,10 @@ func (s *CommitService) ValidateDevicesImageSetWithCommit(devicesUUID []string,
...
+		Where("devices.uuid in (?) AND devices.DELETED_AT is null", devicesUUID).
 		Group("images.image_set_id").
 		Find(&imageSetsDevices); result.Error != nil {
```
AaronH88, line 81: So in this query we have joined the `image set` , `images` and `devices` tables, this is a lot of data. Is all of this data needed?

If not everything is needed then please consider using the gorm `Scan()` function to only take needed data from a table.

AaronH88: https://gorm.io/docs/query.html#Scan

pkg/routes/updates.go
```diff
@@ -226,7 +226,7 @@ func updateFromHTTP(w http.ResponseWriter, r *http.Request) *[]models.UpdateTran
 			ctxServices.Log.WithFields(log.Fields{
 				"error":    err.Error(),
 				"commitID": devicesUpdate.CommitID,
-			}).Error("Commit does not belong to the same image-set as devices")
+			}).Error(err.Error())
```
ldjebran, line 229: we need to have a message what failed ValidateDevicesImageSetWithCommit failure  
or any phrase that indicate that, anyway the error message is already there.

pkg/routes/updates_test.go
omaciel: @acosferreira even if we're indirectly testing the conditional statement for the `services/commits.go` package, I feel that having a `services/commits_test.go` test suite would be well worth the investment for us. So if not for this PR, then we should aim to add tests for `services/commits.go` or we run the risk of never handling our own technical debt AND decreasing our level of confidence for future pushes to PROD.

pkg/services/commits_test.go
```diff
@@ -0,0 +1,82 @@
...
+			OrgID:   orgID,
+			UUID:    faker.UUIDHyphenated(),
+			ImageID: images[1].ID,
+		}
+		db.DB.Create(&device)
+		ctx = context.Background()
+		commitService = services.NewCommitService(ctx, log.NewEntry(log.StandardLogger()))
+	})
+
+	Context("ValidateDevicesImageSetWithCommit", func() {
```
omaciel, line 62: Great! How about some positive tests too, please?

omaciel (approved):
My concern/request about adding unit tests have been addressed.

pkg/routes/updates.go
de1987: yeah, the `err.Error()` is attached as a field, so we should get the info about it
should we not?

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,134 @@ var _ = Describe("Update routes", func() {
...
+				respBody, err := ioutil.ReadAll(rr.Body)
+				err = json.Unmarshal(respBody, &response)
+
+				responseRecorder := httptest.NewRecorder()
+				handler.ServeHTTP(responseRecorder, req)
+				Expect(responseRecorder.Code).To(Equal(http.StatusBadRequest))
+
+			})
+
+			It("should update to version 3 and see 4,5 availble", func() {
```
de1987, line 582:
```suggestion
			It("should update to version 3 and see 4,5 available", func() {
```
typo

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,134 @@ var _ = Describe("Update routes", func() {
...
+				updateData, err := json.Marshal(models.DevicesUpdate{DevicesUUID: []string{device.UUID}, CommitID: commits[3].ID})
+				Expect(err).To(BeNil())
+				req, err := http.NewRequest(http.MethodPost, "/", bytes.NewBuffer(updateData))
+				Expect(err).To(BeNil())
+
+				ctx := req.Context()
+				ctx = dependencies.ContextWithServices(ctx, edgeAPIServices)
+				req = req.WithContext(ctx)
+
+				mockUpdateService.EXPECT().BuildUpdateTransactions(gomock.Any(), orgID, gomock.Any()).Return(&[]models.UpdateTransaction{}, nil)
```
de1987, line 616: instead of using `gomock.Any()`
I would suggest checking for the expected content which will increase our confidence in the code.

de1987 (changes requested):

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,134 @@ var _ = Describe("Update routes", func() {
...
+
+				var response common.APIResponse
+				respBody, err := ioutil.ReadAll(rr.Body)
+				err = json.Unmarshal(respBody, &response)
+
+				Expect(rr.Code).To(Equal(http.StatusOK))
+				Expect(err).Should(BeNil())
+			})
+
+			It("should update to version 4 and see only 5 availble", func() {
```
de1987, line 606:
```suggestion
			It("should update to version 4 and see only 5 available", func() {
```
typo

pkg/routes/updates_test.go
```diff
@@ -503,6 +503,139 @@ var _ = Describe("Update routes", func() {
...
+			It("should update to version 3 and see 4,5 available", func() {
+				updateData, err := json.Marshal(models.DevicesUpdate{DevicesUUID: []string{device.UUID}, CommitID: commits[2].ID})
+				Expect(err).To(BeNil())
+				req, err := http.NewRequest(http.MethodPost, "/", bytes.NewBuffer(updateData))
+				Expect(err).To(BeNil())
+
+				ctx := req.Context()
+				ctx = dependencies.ContextWithServices(ctx, edgeAPIServices)
+				req = req.WithContext(ctx)
+				var desiredCommit models.Commit
```
de1987, line 591: nice name, it will help the team understand more about mocks

de1987 (approved):

omaciel (approved):
/lgtm

PR #1616, opened 2022-10-31
Add Consumer interface to allow mocking.

mgold1234:
realted to THEEDGE-2678

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


pkg/services/consumers.go
```diff
@@ -23,6 +25,51 @@ type ConsumerService interface {
 	Close()
 }
 
+// Consumer is an interface for a kafka consumer, it matches the confluent consumer type definition
+type Consumer interface {
```
AaronH88, line 29: I think this interface should live in /pkg/comon/kafka

AaronH88:
I am failing to see how mocks can be injected into this solution. Can you write some example tests please and show how your solution works

AaronH88:
Looks like we are getting closer, but again, can you please use these mocks in tests and prove that your proposal works.

Here is a PR where I used the producer mocks: https://github.com/RedHatInsights/edge-api/pull/1582/files

pkg/services/image/event_image_iso_requested_test.go
```diff
@@ -63,8 +69,12 @@ var _ = Describe("Event Image ISO Event Test", func() {
 					mockImageService.EXPECT().AddUserInfo(gomock.Any()).Return(nil)
 					event := &eventImageReq.EventImageISORequestedBuildHandler{}
 					event.Data = *edgePayload
+					timeout := 100
+					groupIDConsumer := "imagebuilder"
+					mockConsumer.EXPECT().Poll(timeout)
+					consumer := mockConsumerService.EXPECT().GetConsumer(groupIDConsumer).Return(mockConsumer, nil)
```
AaronH88, line 75: This is just being initialised, not being used. 
The syntax looks ok, you just need to test a consumer, 
In the file `event_image_iso_requested.go` we do not use a kafka consumer

pkg/common/kafka/consumer_test.go
```diff
@@ -0,0 +1,3 @@
+// FIXME: golangci-lint
+// nolint:revive
+package kafkacommon_test
```
AaronH88, line 3: Lets go ahead and write this test file

pkg/services/images_iso/main_test.go
```diff
@@ -3,7 +3,126 @@
...
+						DataSchema:  "v1",
+						ID:          faker.HyphenatedID,
+						RedHatOrgID: common.DefaultOrgID,
+						Source:      models.SourceEdgeEventAPI,
+						SpecVersion: "v1",
+						Subject:     image.Name,
+						Time:        time.Now().Format(time.RFC3339),
+						Type:        models.EventTypeEdgeImageRequested,
+					}
+					err := mockConsumer.EXPECT().Poll(timeout).Return(cloudEvent)
```
AaronH88, line 119: Great!

Now time for the rest of the tests, 
Returning an error, no consumer etc.

pkg/common/kafka/consumer_test.go
```diff
@@ -0,0 +1,74 @@
...
+	. "github.com/onsi/gomega"
+	kafkacommon "github.com/redhatinsights/edge-api/pkg/common/kafka"
+
+	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
+	"github.com/redhatinsights/edge-api/config"
+	mock_kafkacommon "github.com/redhatinsights/edge-api/pkg/common/kafka/mock_kafka"
+	"github.com/redhatinsights/edge-api/pkg/db"
+)
+
+var _ = Describe("Image Iso Kafka Consumer Test", func() {
```
AaronH88, line 18: This is not an ISO test, this describe is misleading

AaronH88:
Please dont upload the binary file [pkg/services/1667593625653020087-services.db-journal](https://github.com/RedHatInsights/edge-api/pull/1616/files#diff-2574cadc9233e01e4ad5c075d116233c0d806081e97b91d099bbc6e24beddde0)

pkg/services/image/event_image_iso_requested_test.go
AaronH88: This code is still not being used. Im surprised this is not causing an error. Is this test correctly being run?

pkg/services/images_iso/main_test.go
AaronH88: Why did you remove this?
This was correct and is the actual test of the consumer

pkg/services/image/event_image_iso_requested_test.go
mgold1234: @AaronH88 it's look like it running
github.com/redhatinsights/edge-api/pkg/services/image   0.082s

I will remove it for now

pkg/common/kafka/consumer.go
```diff
@@ -0,0 +1,96 @@
...
+		log.WithField("error", err.Error()).Error("Failed to create consumer")
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (s *ConsumerService) GetInstanceConsumer(topics []string, groupID string) (Consumer, error) {
+	kafkaConfigMap := NewKafkaConfigMapService().GetKafkaConsumerConfigMap(groupID)
+	c, err := kafka.NewConsumer(&kafkaConfigMap)
```
ldjebran, line 81: mabe instead of calling 

```go
kafkaConfigMap := NewKafkaConfigMapService().GetKafkaConsumerConfigMap(groupID)
	c, err := kafka.NewConsumer(&kafkaConfigMap)
```
we should call 

```go 
c, err := GetConsumer(groupID)
```
what do you think ?

pkg/common/kafka/consumer_test.go
```diff
@@ -0,0 +1,73 @@
...
+		When("One does not exist yet", func() {
+			Context("One does not exist yet", func() {
+				It("should be ok", func() {
+					consumerGroupID := "imagesisobuild"
+					mockKafkaConfigService.EXPECT().GetKafkaConsumerConfigMap(consumerGroupID).Return(*kafkaConfigMap)
+					c, err := service.GetConsumer(consumerGroupID)
+					Expect(err).To(BeNil())
+					Expect(c).ToNot(BeNil())
+				})
+			})
```
ldjebran, line 70: we can test GetInstanceConsumer and verify that GetConsumer get called

pkg/common/kafka/consumer.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if err != nil {
+		log.WithField("error", err.Error()).Error("Failed to create consumer")
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (s *ConsumerService) GetInstanceConsumer(topics []string, groupID string) (Consumer, error) {
```
ldjebran, line 79: as this function is not used (maybe this is for future refactoring) mabe should be removed.
what do you think ?

ldjebran (commented):
Looks good left some comments

pkg/common/kafka/consumer.go
mgold1234: I will delete this function

pkg/common/kafka/consumer_test.go
mgold1234: I will delete this function

pkg/common/kafka/consumer.go
mgold1234: delete this function

pkg/common/kafka/consumer.go
```diff
@@ -0,0 +1,76 @@
...
+
+// NewConsumerService returns a new service
+func NewConsumerService() ConsumerServiceInterface {
+	return &ConsumerService{
+		Topic:          NewTopicService(),
+		KafkaConfigMap: NewKafkaConfigMapService(),
+	}
+}
+
+func (s *ConsumerService) GetConsumer(groupID string) (Consumer, error) {
```
AaronH88, line 66: This exposed function needs a comment in the format `// GetConsumer ...`

pkg/services/images_iso/main.go
```diff
@@ -104,14 +100,25 @@ func main() {
...
 				mslog.WithField("error", err).Error("Error storing offset after ISO message")
 			}
-		case kafka.Error:
+			continue
+			// run = false
+		case *kafka.Error:
 			// terminate the application if all brokers are down.
+			mslog.Info("ISO Microservice Error")
 			log.WithFields(log.Fields{"code": e.Code(), "error": e}).Error("Exiting ISO loop due to Kafka broker issue")
-			if e.Code() == kafka.ErrAllBrokersDown {
```
AaronH88, line 110: Lets go back to using this if statement

pkg/services/images_iso/main.go
```diff
@@ -104,14 +100,25 @@ func main() {
...
 		}
 	}
 }
+
+func main() {
+	ctx := context.Background()
+	// Init edge api services and attach them to the context
+	edgeAPIServices := dependencies.Init(ctx)
+	ctx = dependencies.ContextWithServices(ctx, edgeAPIServices)
+	// create a base logger with fields to pass through the entire flow
```
AaronH88, line 121: remove this comment please

pkg/services/images_iso/main.go
```diff
@@ -17,16 +17,11 @@ import (
...
 )
 
-func main() {
-	ctx := context.Background()
-	// Init edge api services and attach them to the context
-	edgeAPIServices := dependencies.Init(ctx)
-	ctx = dependencies.ContextWithServices(ctx, edgeAPIServices)
-	// create a base logger with fields to pass through the entire flow
+func initConsumer(ctx context.Context) {
+	edgeAPIServices := dependencies.ServicesFromContext(ctx)
```
ldjebran, line 21: I think we have a problem with this context in the current implementation of the consumer, this is not the problem with PR.
so we are passing the context that we globally create and then pass it the Consume function 
think we should create a new context for each event , and after that pass it to Consume function that can add identity to it (new context).

@AaronH88 @loadtheaccumulator what do you think ?
Note This problem exist in images_build also and other microservice

AaronH88: Where do you suggest this context gets created?

mgold1234: is this pr can stay as  is, and if we want to refactor it will be in other pr?

ldjebran: yes , the current can stay as is, as the problem is affecting all the consumers .

I will resolve this comment, waiting for a resolution in an other PR for all the consumers services.

ldjebran (dismissed):
LGTM

pkg/services/images_iso/main.go
ldjebran: when investigating offline we came to the conclusion that line with code 
ctx = image.ContextWithLogger(ctx, mslog)

is creating a copy of our context that is then passed to Consume function

mgold1234: @ldjebran change it as you suggest, thank you for that

ldjebran (dismissed):
LGTM

pkg/common/kafka/consumer.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"time"
+
+	"github.com/confluentinc/confluent-kafka-go/kafka"
+	log "github.com/sirupsen/logrus"
+)
+
+// Consumer is an interface for a kafka consumer, it matches the confluent consumer type definition
+type Consumer interface {
```
loadtheaccumulator, line 11: Should this only be the methods we use?
What is the process to maintain this if the definitions change upstream?

loadtheaccumulator (changes requested):
requesting some changes and info

pkg/dependencies/main.go
```diff
@@ -26,6 +27,7 @@ type EdgeAPIServices struct {
 	OwnershipVoucherService services.OwnershipVoucherServiceInterface
 	DeviceGroupsService     services.DeviceGroupsServiceInterface
 	FilesService            services.FilesService
+	ConsumerService         kafkacommon.ConsumerServiceInterface
```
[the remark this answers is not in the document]
mgold1234: we cannot just call it, because in tests we may need to use the mock service, to use the mock we have to pass the service via context 
in the code we pass the real service, in the tests we pass the mock service

loadtheaccumulator (approved):

AaronH88 (approved):
/lgtm

PR #1626, opened 2022-11-07
Created event device sync.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/devices/event_device_sync_test.go
+++ b/pkg/services/devices/event_device_sync_test.go
@@ -1,0 +1,54 @@
+package devices_test
+
+import (
+	"context"
+	"time"
+
+	"github.com/bxcodec/faker/v3"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/redhatinsights/edge-api/pkg/dependencies"
+	"github.com/redhatinsights/edge-api/pkg/models"
+	"github.com/redhatinsights/edge-api/pkg/routes/common"
+	eventReq "github.com/redhatinsights/edge-api/pkg/services/devices"
+	"github.com/redhatinsights/edge-api/pkg/services/mock_services"
+	"github.com/redhatinsights/edge-api/pkg/services/utility"
+	log "github.com/sirupsen/logrus"
+)
+
+var _ = Describe("Event Device sync Event Test", func() {
+	var ctx context.Context
+	var mockDeviceService *mock_services.MockDeviceServiceInterface
+	BeforeEach(func() {
+		ctrl := gomock.NewController(GinkgoT())
+		defer ctrl.Finish()
+		mockDeviceService = mock_services.NewMockDeviceServiceInterface(ctrl)
+
+		ctx = context.Background()
+		ctx = dependencies.ContextWithServices(ctx, &dependencies.EdgeAPIServices{
+			DeviceService: mockDeviceService,
+		})
+		ctx = utility.ContextWithLogger(ctx, log.NewEntry(log.StandardLogger()))
+	})
+	Describe("consume device sync event", func() {
+		When("device sync is requested", func() {
+			Context("DB and inventory mismatch", func() {
+				It("should be ok", func() {
+					ident, err := common.GetIdentityFromContext(ctx)
+					Expect(err).To(BeNil())
+					edgePayload := &models.EdgeBasePayload{
+						Identity:       ident,
+						LastHandleTime: time.Now().Format(time.RFC3339),
+						RequestID:      faker.UUIDHyphenated(),
+					}
+					event := &eventReq.EventDeviceSyncHandler{}
+					event.RedHatOrgID = ident.Identity.OrgID
+					event.Data = *edgePayload
+					mockDeviceService.EXPECT().SyncDevicesWithInventory(event.RedHatOrgID).Return()
+					event.Consume(ctx)
+				})
+			})
+		})
+	})
+})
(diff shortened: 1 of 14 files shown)


pkg/services/events/event_image_requested.go
```diff
@@ -1,6 +1,6 @@
 // FIXME: golangci-lint
 // nolint:gosimple,revive
-package image
+package events
```
loadtheaccumulator, line 3: I'm thinking if the app is truly EDA, image should be a self-contained package that includes its event handlers with functional code vs. going down the path we went with all services (images, devices, etc.) being in a single services package. Event would be a package specific to event framework logic leveraged by the other packages.

Is there a functional reason to move it or just thinking organizational?

loadtheaccumulator (dismissed):
Just a fundamental question on event_image_requested.go moving to events.

pkg/services/events/event_device_sync_test.go
```diff
@@ -0,0 +1,55 @@
...
+			DeviceService: mockDeviceService,
+		})
+		ctx = eventReq.ContextWithLogger(ctx, log.NewEntry(log.StandardLogger()))
+	})
+	Describe("consume device sync event", func() {
+		When("device sync is requested", func() {
+			Context("DB and inventory mis match", func() {
+				It("should be ok", func() {
+					orgID := faker.UUIDHyphenated()
+					ident, err := common.GetIdentityFromContext(ctx)
```
ldjebran, line 38: I suppose that identity has orgID = to DefaultOrgID and event orgID = faker.UUIDHyphenated()
strange that there is no error raised somewhere,
are we missing something ?

pkg/services/events/event_image_requested.go
AaronH88: Yes,

context.go and error.go are both in the image package, but will need to be reused for the devices events.

I have many ways of grouping these files together, some examples.
As you suggest, group image events in an image package, device events in a device package, and common utility files in come sort of common or utility package.
The alternative I was looking at. Group events together along with their helper files (context, error etc). Group the services by their domain, image service, device service etc, 

I think you have incorrectly said "truly EDA, image should be a self-contained package". 
Image is a domain, device is a domain, events can easily be a domain if we choose.

I will change this PR to reflect your idea outlined

loadtheaccumulator: Would image then strictly be a library and events would be the glue that makes it an application? I think I'd still be looking to separate image events from device events since they could exist as two separate applications.

loadtheaccumulator: Events as a domain gels in my brain more if we're looking at Edge Mgmt as a single application with image and device domains separated instead of being a part of a large monolithic service as we have now.

pkg/services/events/event_device_sync_test.go
AaronH88: I added a check in the consume to ensure these org ids match. 
I edited the tests :smile:

pkg/services/devices/event_device_sync_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+		ctx = context.Background()
+		ctx = dependencies.ContextWithServices(ctx, &dependencies.EdgeAPIServices{
+			DeviceService: mockDeviceService,
+		})
+		ctx = utility.ContextWithLogger(ctx, log.NewEntry(log.StandardLogger()))
+	})
+	Describe("consume device sync event", func() {
+		When("device sync is requested", func() {
+			Context("DB and inventory mis match", func() {
```
de1987, line 36:
```suggestion
			Context("DB and inventory miss match", func() {
```
typo

de1987 (dismissed):

pkg/services/devices/event_device_sync_test.go
ldjebran: maybe simply the known "mismatch" :)

AaronH88: ty! I accidentally hit the space bar, fixed now

pkg/services/devices/event_device_sync_test.go
```diff
@@ -0,0 +1,55 @@
...
+					edgePayload := &models.EdgeBasePayload{
+						Identity:       ident,
+						LastHandleTime: time.Now().Format(time.RFC3339),
+						RequestID:      faker.UUIDHyphenated(),
+					}
+					Expect(edgePayload).ToNot(BeNil())
+					event := &eventReq.EventDeviceSyncHandler{}
+					event.RedHatOrgID = ident.Identity.OrgID
+					event.Data = *edgePayload
+					mockDeviceService.EXPECT().SyncDevicesWithInventory(gomock.Any()).Return()
```
ldjebran, line 49:
```suggestion
					mockDeviceService.EXPECT().SyncDevicesWithInventory(event.RedHatOrgID).Return()
```

ldjebran (commented):
LGTM, left one comment

pkg/services/devices/event_device_sync_test.go
AaronH88: Tyvm! added

ldjebran (approved):
LGTM

pkg/services/devices/event_device_sync_test.go
```diff
@@ -0,0 +1,54 @@
...
+					event := &eventReq.EventDeviceSyncHandler{}
+					event.RedHatOrgID = ident.Identity.OrgID
+					event.Data = *edgePayload
+					mockDeviceService.EXPECT().SyncDevicesWithInventory(event.RedHatOrgID).Return()
+					event.Consume(ctx)
+				})
+			})
+		})
+	})
+})
```
mgold1234, line 54: what about negative test?

mgold1234 (commented):
I think the name of the folder should be 'device' instead of 'devices' like 'image'
left one comment,
look very good

loadtheaccumulator (approved):

PR #1633, opened 2022-11-08
fixes /devices filters

tomershinhar:
# Description

the /devices endpoint filters was not matching the devices filters listed in query_params.go which caused the endpoint to not except valid query params and also passing invalid one (which is not actually doing anything).
this PR updates the query_params.go file to match the query params accepted by the /devices endpoint and also updates the spec file to better document how to use those query params 

Fixes # (issue)
[THEEDGE-2717](https://issues.redhat.com/browse/THEEDGE-2717)

## Type of change

What is it?

- [x] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


AaronH88 (approved):
/lgtm

pkg/routes/query_params.go
```diff
@@ -18,7 +18,7 @@ func initalizeQueryParamsArray() map[string][]string {
 	if len(m) == 0 {
 		m = make(map[string][]string)
 		m["device-groups"] = []string{"limit", "offset", "name", "created_at", "updated_at", "sort_by"}
-		m["devices"] = []string{"per_page", "page", "sort_by", "order_how", "hostname_or_id"}
+		m["devices"] = []string{"per_page", "page", "order_how", "hostname_or_id", "order_by"}
```
ldjebran, line 21: I see sort_by and not order_by  
https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/common/filters.go#L120

cmd/spec/path.yaml
```diff
@@ -629,12 +629,12 @@ paths:
             type: integer
         - name: order_by
```
ldjebran, line 630: may be this line should be replaced by sort_by as per https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/common/filters.go#L120

ldjebran (changes requested):
I see sort_by and not order_by here
https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/common/filters.go#L120

pkg/routes/query_params.go
tomershinhar: for the /devices endpoint the filters are taken from host-inventory https://github.com/RedHatInsights/edge-api/blob/main/pkg/routes/devices.go#L252
so while sort_by was allowed it did not do anything

ldjebran (approved):

PR #1670, opened 2022-11-20
[THEEDGE-2844] fixed image-set view sort_by params

tomershinhar:
# Description

the image-sets routes were not validating the query params correctly, for some of the endpoints it was not validated at all and for others it did not do it correctly.

to resolve this i had to split `sortOption` to two different variables and use the one corresponding to endpoint being tested in `validateFilterParams`

also added unit-tests for the query params in general and sort_by specifically 

Fixes # (issue)
[THEEDGE-2844](https://issues.redhat.com/browse/THEEDGE-2844)
## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/imagesets.go
+++ b/pkg/routes/imagesets.go
@@ -29,6 +29,7 @@ type imageSetImageTypeKey int
 const imageSetKey imageSetTypeKey = iota
 const imageSetImageKey imageSetImageTypeKey = iota
 
+var sortImageSetImageOption = []string{"created_at", "name", "version"}
 var sortOption = []string{"created_at", "updated_at", "name"}
 var statusOption = []string{models.ImageStatusCreated, models.ImageStatusBuilding, models.ImageStatusError, models.ImageStatusSuccess}
 
@@ -40,7 +41,7 @@ func MakeImageSetsRouter(sub chi.Router) {
 		r.Use(ImageSetCtx)
 		r.With(validateFilterParams).With(common.Paginate).Get("/", GetImageSetsByID)
 	})
-	sub.With(validateFilterParams).Route("/view/{imageSetID}", func(r chi.Router) {
+	sub.With(ValidateQueryParams("imagesetimageview")).With(validateFilterParams).Route("/view/{imageSetID}", func(r chi.Router) {
 		r.Use(ImageSetViewCtx)
 		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/", GetImageSetViewByID)
 		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/versions", GetAllImageSetImagesView)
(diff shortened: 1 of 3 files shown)


mgold1234 (dismissed):
/lgtm

pkg/routes/imagesets.go
```diff
@@ -40,13 +41,13 @@ func MakeImageSetsRouter(sub chi.Router) {
 		r.Use(ImageSetCtx)
 		r.With(validateFilterParams).With(common.Paginate).Get("/", GetImageSetsByID)
 	})
-	sub.With(validateFilterParams).Route("/view/{imageSetID}", func(r chi.Router) {
+	sub.Route("/view/{imageSetID}", func(r chi.Router) {
 		r.Use(ImageSetViewCtx)
-		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/", GetImageSetViewByID)
-		r.With(ValidateGetAllImagesSearchParams).With(common.Paginate).Get("/versions", GetAllImageSetImagesView)
+		r.With(ValidateQueryParams("imagesetimageview")).With(validateFilterParams).With(common.Paginate).Get("/", GetImageSetViewByID)
+		r.With(ValidateQueryParams("imagesetimageview")).With(validateFilterParams).With(common.Paginate).Get("/versions", GetAllImageSetImagesView)
```
ldjebran, line 47: I think that ValidateGetAllImagesSearchParams was OK as we are using that params internally 
ValidateQueryParams("imagesetimageview") is a good addition here
we should not use validateFilterParams for imageSet and images (eg in a mix)
let validateFilterParams for imageSets
and ValidateGetAllImagesSearchParams for images

ldjebran (commented):
Thanks @tomershinhar, left a comment, the unit-tests are a very good additions here to prevent regressions

pkg/routes/imagesets.go
tomershinhar: the issue is that `ValidateGetAllImagesSearchParams` does not match this endpoints very well.
for example, it validates that `sort_by` gets "status" ,"name" ,"distribution" , "created_at" or"version" 
while this is close, "distribution" is not a valid option for `sort_by` on those endpoints.

it is also missing "id" and "version" query params, so those will not get validated if we use `ValidateGetAllImagesSearchParams`

`validateFilterParams` (with the fix in this PR) matches the valid query params for those endpoints

ldjebran (approved):
LGTM

pkg/routes/imagesets.go
ldjebran: we discussed offline about creating a new function to remove any confusion between imageSets , imageset versions  (images)

mgold1234 (approved):
/lgtm

PR #1674, opened 2022-11-21
Adding DELETE for image sets that have no devices using them.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/imagesets.go
+++ b/pkg/services/imagesets.go
@@ -24,6 +24,8 @@ type ImageSetsServiceInterface interface {
 	GetImageSetsBuildIsoURL(orgID string, imageSetIDS []uint) (map[uint]uint, error)
 	GetImagesViewData(imageSetID uint, imagesLimit int, imagesOffSet int, tx *gorm.DB) (*ImagesViewData, error)
 	GetImageSetImageViewByID(imageSetID uint, imageID uint) (*ImageSetImageIDView, error)
+	GetDeviceIdsByImageSetID(imageSetId uint) (int, []string, error)
+	DeleteImageSet(imageSetID uint) error
 }
 
 // ImagesViewData is the images view data return for images view with filters , limit, offSet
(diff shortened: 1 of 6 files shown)


AaronH88:
Writing tests now. 
Looking for early feedback on this draft. 

The plan for future PR's is to create `routes` for getting all device ids for an Image and an Image Set.

pkg/services/imagesets.go
```diff
@@ -372,3 +374,49 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
 
 	return &imageSetImageIDView, nil
 }
+
+func (s *ImageSetsService) GetDeviceIdsByImageSetID(imageSetID uint) (*DevicesByImage, error) {
```
ldjebran, line 378: I think we can do it with one SQL query without the loop over images

pkg/services/imagesets.go
```diff
@@ -372,3 +374,49 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
...
+		count += temp.count
+		deviceIds = append(deviceIds, temp.deviceIDs...)
+	}
+
+	ret := DevicesByImage{count: count, deviceIDs: deviceIds}
+
+	return &ret, nil
+}
+
+func (s *ImageSetsService) DeleteImageSet(imageSetID uint) error {
```
ldjebran, line 403: what will happen, when a device is being (in the process of being) deployed , using one  of the imageSet images ?
should we envisage restoring the related image and the corresponding imageSet ?

pkg/models/images.go
```diff
@@ -20,7 +20,7 @@ type ImageSet struct {
 	Version int     `json:"Version" gorm:"default:1"`
 	Account string  `json:"Account"`
 	OrgID   string  `json:"org_id" gorm:"index;<-:create"`
-	Images  []Image `json:"Images"`
+	Images  []Image `json:"Images" gorm:"constraint:OnDelete:CASCADE;"`
```
ldjebran, line 23: this will need some testing, as we are soft deleting our records, and this may permanently delete the records
(is this working on sqlite for the unit-tests?)

pkg/services/imagesets.go
AaronH88: We can make this happen, but in a different PR. By default gorm does a "soft" delete unless "Unscoped()" is used

pkg/models/images.go
AaronH88: Tested and it does not work. I removed it

AaronH88:
Tests have been added, the code is ready for review

pkg/services/imagesets.go
```diff
@@ -407,3 +409,70 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
...
+		s.log.WithFields(log.Fields{"error": "Image Set in use"}).Error("Error, unable to delete image set in use")
+		err := new(ImageSetInUse)
+		return err
+	}
+
+	var imageSet models.ImageSet
+	result := db.DB.First(&imageSet, imageSetID)
+	if result.Error != nil {
+		s.log.WithField("error", result.Error.Error()).Error("Error getting image set by id")
+		return new(ImageSetNotFoundError)
```
ldjebran, line 454: ImageSetNotFoundError if result.Error == gorm.ErrRecordNotFound for other errors we return result.Error

pkg/services/imagesets.go
```diff
@@ -407,3 +409,70 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
...
+	}
+	result = db.DB.Where("image_set_id = ?", imageSetID).Find(&imageSet.Images)
+	if result.Error != nil {
+		s.log.WithField("error", result.Error.Error()).Error("Error getting image set's images")
+		return new(ImageNotFoundError)
```
ldjebran, line 459: here the error cannot be record not found as we are using Find, the return should be result.Error

pkg/services/imagesets.go
```diff
@@ -407,3 +409,70 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
...
+	}
+
+	result = db.DB.Delete(&imageSet.Images)
```
ldjebran, line 462: let's check len(imageSet.Images) > 0 before deleting what do you think ?

ldjebran (commented):
looks good , left some comments

pkg/services/imagesets.go
AaronH88: ill do this for both :smile:

pkg/services/imagesets.go
AaronH88: cool, ill add a test for this also

ldjebran (approved):
LGTM

pkg/services/imagesets.go
```diff
@@ -407,3 +409,71 @@ func (s *ImageSetsService) GetImageSetImageViewByID(imageSetID uint, imageID uin
...
+
+func (s *ImageSetsService) GetDeviceIdsByImageSetID(imageSetID uint) (int, []string, error) {
+	type DeviceInfo struct {
+		DeviceID   uint   `json:"device_id"`
+		DeviceUUID string `json:"device_uuid"`
+	}
+	var deviceInfos []DeviceInfo
+
+	if result := db.DB.Table("image_sets").
+		Where("image_sets.id = ?", imageSetID).
```
acosferreira, line 421: on images you have the imageset info, do you realy need to load 3 tables here?
is it possible  start the join between devices and image, getting the distinct image.image_set_id ?

tomershinhar:
@AaronH88 please update path.yaml with the new endpoint

acosferreira (approved):
lgtm

loadtheaccumulator (approved):

PR #1682, opened 2022-11-23
apiResponse: FIXES routes image-set and image api responses.

ldjebran:
FIXES routes image-set and image api responses, using appropriate functions RespondWithAPIError and respondWithJSONBody. some code fixes.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:


pkg/routes/imagesets.go
```diff
@@ -116,35 +115,36 @@ func ImageSetCtx(next http.Handler) http.Handler {
 			s.Log = s.Log.WithField("imageSetID", imageSetID)
 			_, err := strconv.Atoi(imageSetID)
 			if err != nil {
-				err := errors.NewBadRequest(err.Error())
-				w.WriteHeader(err.GetStatus())
-				if err := json.NewEncoder(w).Encode(&err); err != nil {
-					s.Log.WithField("error", err.Error()).Error("Error while trying to encode")
-				}
+				s.Log.WithField("error", err.Error()).Error("imageSetID must be of type integer")
```
omaciel, line 118: nitpick: the error message could be a constant

pkg/routes/imagesets.go
```diff
@@ -116,35 +115,36 @@ func ImageSetCtx(next http.Handler) http.Handler {
...
 				return
 			}
 			result := db.Org(orgID, "").Where("Image_sets.id = ?", imageSetID).First(&imageSet)
 
 			if result.Error != nil {
-				err := errors.NewNotFound(result.Error.Error())
-				w.WriteHeader(err.GetStatus())
-				if err := json.NewEncoder(w).Encode(&err); err != nil {
-					s.Log.WithField("error", err.Error()).Error("Error while trying to encode")
+				s.Log.WithField("error", result.Error.Error()).Error("error getting image-set")
```
omaciel, line 125: wish: include the imageset id in the message

pkg/routes/imagesets.go
```diff
@@ -245,24 +237,21 @@ func ListAllImageSets(w http.ResponseWriter, r *http.Request) {
 		imageSetInfo = append(imageSetInfo, imgSet)
 	}
 	if result.Error != nil {
-		s.Log.WithField("error", result.Error.Error()).Error("Image sets not found")
-		err := errors.NewBadRequest("Not Found")
-		w.WriteHeader(err.GetStatus())
-		if err := json.NewEncoder(w).Encode(&err); err != nil {
-			s.Log.WithField("error", err.Error()).Error("Error while trying to encode")
+		s.Log.WithField("error", result.Error.Error()).Error("error occurred while getting Image sets image installer")
```
omaciel, line 240: wish: create a new, separate function to fetch the image installer for an image set. I think it would be easier to maintain AND test the new function

pkg/routes/imagesets.go
```diff
@@ -287,11 +276,9 @@ func GetImageSetsByID(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	imageSet, ok := ctx.Value(imageSetKey).(*models.ImageSet)
 	if !ok {
-		err := errors.NewBadRequest("Must pass image set id")
-		w.WriteHeader(err.GetStatus())
-		if err := json.NewEncoder(w).Encode(&err); err != nil {
-			s.Log.WithField("error", err.Error()).Error("Error while trying to encode")
-		}
+		s.Log.Error("image-set not found in the context")
```
omaciel, line 279: wish: add the imageset id to the log message.

omaciel (changes requested):
I think all existing tests should also check/assert that the error raised was of type `respondWithAPIError`

pkg/routes/imagesets.go
ldjebran: There is https://github.com/RedHatInsights/edge-api/blob/main/pkg/services/imagesets.go#L200
and it return a map of all passed image-set

The purpose of this PR is not to change logic but to unify the routes base code

omaciel: > There is https://github.com/RedHatInsights/edge-api/blob/main/pkg/services/imagesets.go#L200 and it return a map of all passed image-set
> 
> The purpose of this PR is not to change logic but to unify the routes base code

My point being that there's a significant chunk of code here that might benefit from a future refactoring exercise, as we're currently looping through the imagesets returned to find their individual installers, unless I'm mistaken?

pkg/routes/imagesets.go
ldjebran: we have no accessible imageset id here to add, and we assume that the imageSet value have to be in the context , if not we raise an error

omaciel: Can we use `imageSetKey` then?

pkg/routes/imagesets.go
omaciel: @ldjebran and I discussed over Slack and agreed that what I suggested requires a much more involved code refactoring, so I'll mark this as resolved.

pkg/routes/imagesets.go
ldjebran: unfortunately imageSetKey will not give us much information, as this is a project private constant iota , we have to look at this function like (the object must be in context before calling it at key imageSetKey and must be of type *models.ImageSet)

omaciel: No problem. This was more of a wish anyhow :)

pkg/routes/imagesets.go
ldjebran: fixed

pkg/routes/imagesets.go
ldjebran: fixed

ldjebran:
increased routes/images from 32.7% up to 71.7 %

 work in progress for routes/imagesets

omaciel:
This is really impressive and deeply appreciated!

ldjebran:
updated image sets tests from 67.7 to 72.2
The uncovered are due to using db queries directly from routes, this needs a deep code refactoring that will be done in a separate PR  
image

image

ldjebran:
@omaciel please revisit this PR

omaciel (approved):
/lgtm

PR #1715, opened 2022-12-06
Added route for returning device ids for a given image set.

AaronH88:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


ldjebran (dismissed):
LGTM

pkg/routes/imagesets.go
```diff
@@ -686,3 +687,24 @@ func DeleteImageSet(w http.ResponseWriter, r *http.Request) {
...
+
+func GetImageSetsDevicesByID(w http.ResponseWriter, r *http.Request) {
+	ctxServices := dependencies.ServicesFromContext(r.Context())
+	imageSet := getContextImageSet(w, r)
+	if imageSet == nil {
+		return
+	}
+	count, devices, err := ctxServices.ImageSetService.GetDeviceIdsByImageSetID(imageSet.ID)
+	if err != nil {
+		respondWithAPIError(w, ctxServices.Log, errors.NewInternalServerError())
```
omaciel, line 704: Should we explicitly log something specific here?

pkg/routes/imagesets_test.go
```diff
@@ -153,6 +153,100 @@ func TestDeleteImageSetError(t *testing.T) {
...
+			status, http.StatusOK)
+	}
+
+	respBody, err := ioutil.ReadAll(rr.Body)
+	jsonBody := ImageSetDevices{}
+	err = json.Unmarshal(respBody, &jsonBody)
+	if err != nil {
+		t.Errorf("failed decoding response body: %s", err.Error())
+	}
+	if jsonBody.Count != len(deviceIDs) || !compareStringArr(jsonBody.Data, deviceIDs) {
```
omaciel, line 227: Could you have used `!reflect.DeepEqual(jsonBody.Data, deviceIDs)` instead and saved yourself from writing `func compareStringArr`, which has no tests for btw :) ?

omaciel (commented):
Please review my comments

de1987 (dismissed):

pkg/routes/imagesets.go
AaronH88: Logs are thrown in the called service and can be found in `pkg/services/imageset.go` function `GetDeviceIdsByImageSetID` around line 426.

I can log again but it will be a duplication.

I can add a log here regardless if you feel its useful :smile: 
Let me know

pkg/routes/imagesets_test.go
AaronH88: tyvm, added!

pkg/routes/imagesets.go
omaciel: I'll answer your question with a question ;) If you were debugging an issue related to this new endpoint, would a specific log message for the route be helpful? If not, discard my suggestion.

AaronH88: Ahh the old question with a question trick! Love it :laughing: 

In this instance, no, but only because the service throws only **one** error. 
If the service threw more than one error then adding explicit logs is absolutely useful. 
but its great to double check depending on the route!

If its ok with you, ill discard this one :smile:

omaciel (approved):
/lgtm

PR #1741, opened 2022-12-13
move it from DeviceGroup end point to imporve performance

mgold1234:
pacakgeDiff appear every time we get deviceGroup and is affect on th…e performance,

so I move it from DeviceGroup end point to imporve performance related to issue THEEDGE-2914

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


pkg/services/devicegroups.go
```diff
@@ -183,7 +183,6 @@ func (s *DeviceGroupsService) GetDeviceImageInfo(images map[int]models.DeviceIma
 			var updAvailable bool
 			var deviceImage models.Image
 			var deviceImageSet models.ImageSet
-			var imagePackageDiff models.PackageDiff
```
de1987, line 186: did you check if it will affect the frontend?

ldjebran: I agree, this a service function that may be used in many different scenarios, may be we should create a device-groups-view  like we did for devices and for image-sets , that view will be available and serve only the UI needs.

ldjebran: but if that functions was created for only this UI purpose , we can go with this fix.

mgold1234: I checked and this function GetDeviceImageInfo used only once on getDeviceGroups function

de1987 (approved):

omaciel:
So no tests were affected... because we didn't have any???

mgold1234:
@omaciel we didn't check this specific field in the unit test,
and I checked the frontend side with Demetris and we see there is no affect on UI side

omaciel:
@mgold1234 great opportunity to add a test then, don't you think? Think about it: you have just deleted a few lines of code and we have no way to tell if this will break our code...

mgold1234:
@omaciel good point, will add it

omaciel (approved):
/lgtm

ldjebran (approved):

PR #1742, opened 2022-12-13
Added unittests for pkg/routes/common/org_id.go.

omaciel:
# Description

Added unittests for `pkg/routes/common/org_id.go`.

Screenshot 2022-12-13 at 12 14 25

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


omaciel:
By the way, it doesn't look like the function `GetAccountOrOrgIDFromContext` is used anywhere. How about deleting it?

omaciel:
From 0% to 86.96%

pkg/routes/common/org_id_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+)
+
+func TestGetOrgID(t *testing.T) {
+	req, err := http.NewRequest("GET", "/", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	gotOrgID, gotError := GetOrgID(req)
```
mgold1234, line 20: better to write getOrgID, err IMHO

pkg/routes/common/org_id_test.go
```diff
@@ -0,0 +1,132 @@
...
+			ExpectedError: errors.New("cannot find org-id"),
+		},
+	}
+
+	for _, test := range cases {
+		t.Run(test.Name, func(t *testing.T) {
+			config.Init()
+			cfg := config.Get()
+			cfg.Auth = test.Auth
+			gotOrgID, gotError := GetOrgIDFromContext(test.Context)
```
mgold1234, line 77: better to write getOrgID, err IMHO

pkg/routes/common/org_id_test.go
```diff
@@ -0,0 +1,132 @@
...
+			ExpectedError:   nil,
+		},
+	}
+
+	for _, test := range cases {
+		t.Run(test.Name, func(t *testing.T) {
+			config.Init()
+			cfg := config.Get()
+			cfg.Auth = test.Auth
+			gotAccount, gotOrgID, gotError := GetAccountOrOrgIDFromContext(test.Context)
```
mgold1234, line 126: same

pkg/routes/common/org_id_test.go
omaciel: I think it is more intuitive to use `got` and not `get` because it is supposed to be read as part of the assertions: I want Y and I got Y. This applies to the orgID I got and the error I got.

pkg/routes/common/org_id_test.go
omaciel: Same explanation as before.

pkg/routes/common/org_id_test.go
omaciel: Same explanation as before.

pkg/routes/common/org_id_test.go
```diff
@@ -0,0 +1,132 @@
...
+			ExpectedOrgID: "",
+			ExpectedError: errors.New("cannot find org-id"),
+		},
+	}
+
+	for _, test := range cases {
+		t.Run(test.Name, func(t *testing.T) {
+			config.Init()
+			cfg := config.Get()
+			cfg.Auth = test.Auth
```
ldjebran, line 76: we need to restore the config auth value after this tests , as this may affect other tests, only if you are sure that no new tests will be created  in this test suite.

ldjebran (commented):
overral looks good, left only one comment about config auth restore.

pkg/routes/common/org_id_test.go
omaciel: Done

mgold1234 (dismissed):
/lgtm

omaciel:
@ldjebran yoohoo :)

pkg/routes/common/org_id_test.go
```diff
@@ -0,0 +1,87 @@
...
+			cfg := config.Get()
+			// Save current config.Auth
+			auth := cfg.Auth
+			cfg.Auth = test.Auth
+			gotOrgID, gotError := GetOrgIDFromContext(test.Context)
+			assert.Equal(t, gotOrgID, test.ExpectedOrgID)
+			assert.Equal(t, gotError, test.ExpectedError)
+
+			// Reset config.Auth
+			cfg.Auth = auth
```
ldjebran, line 84: I would prefer if it was at the beginning of the test with defer

```go
func TestGetDefaultAccount(t *testing.T) {
    cfg := config.Get()   // note we do not need to call Init() before
    auth := cfg.Auth
    defer func(auth bool) {
        config.Get().Auth = auth
    }(auth)
```

PS: if you do this, you do not need to init and get a new config on each loop, as once initiated it will return always the same instance value

config.Get() will init if not initiated and return the conf

omaciel: Done

ldjebran (approved):
LGTM

ldjebran (approved):

omaciel:
@mgold1234  yoohoo

mgold1234 (approved):
/lgtm

PR #1772, opened 2022-12-21
image-sets-view: fixes image-sets-view count sql.

ldjebran:
align the count sql with the one for data.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/imagesets.go
+++ b/pkg/services/imagesets.go
@@ -102,10 +102,16 @@ func (s *ImageSetsService) GetImageSetsViewCount(tx *gorm.DB) (int64, error) {
 
 	var count int64
 
-	if result := db.OrgDB(orgID, tx, "image_sets").Debug().
-		Joins(`JOIN images ON image_sets.id = images.image_set_id`).
-		Model(&models.ImageSet{}).Distinct("image_sets.id").Count(&count); result.Error != nil {
-		s.log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error("Error getting image sets count")
+	// create a sub query of the latest images and their corresponding image sets
+	latestImagesSubQuery := db.Org(orgID, "").Model(&models.Image{}).Select("image_set_id", "max(id) as image_id").Group("image_set_id")
+	if result := db.OrgDB(orgID, tx, "images").Table("(?) as latest_images", latestImagesSubQuery).
+		Joins("JOIN images on images.id = latest_images.image_id").
+		Joins("JOIN image_sets on image_sets.id = latest_images.image_set_id").
+		Where("image_sets.deleted_at IS NULL").
+		Count(&count); result.Error != nil {
+		log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error(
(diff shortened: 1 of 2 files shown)


pkg/services/imagesets.go
```diff
@@ -102,10 +102,16 @@ func (s *ImageSetsService) GetImageSetsViewCount(tx *gorm.DB) (int64, error) {
...
+	// create a sub query of the latest images and their corresponding image sets
+	latestImagesSubQuery := db.Org(orgID, "").Model(&models.Image{}).Select("image_set_id", "max(id) as image_id").Group("image_set_id")
+	if result := db.OrgDB(orgID, tx, "images").Table("(?) as latest_images", latestImagesSubQuery).
+		Joins("JOIN images on images.id = latest_images.image_id").
+		Joins("JOIN image_sets on image_sets.id = latest_images.image_set_id").
+		Where("image_sets.deleted_at IS NULL").
+		Count(&count); result.Error != nil {
+		log.WithFields(log.Fields{"error": result.Error.Error(), "OrgID": orgID}).Error(
+			"error when getting image sets view data",
+		)
```
de1987, line 114: Would be great if we have a test to cover this change.
WDYT?

tomershinhar (dismissed):
LGTM

pkg/services/imagesets.go
ldjebran: Ok will create a test , note that we are only aligning with the data one so nothing new done here.

ldjebran: there was one unit-tests for count added 3 case for status , that reflect the functionality.

pkg/services/imagesets_test.go
```diff
@@ -90,6 +92,33 @@ var _ = Describe("ImageSets Service Test", func() {
...
+			Expect(count).To(Equal(int64(1)))
+		})
+
+		It("should return The right image view count  when filtering by status building", func() {
+
+			dbFilter := db.DB.Where("image_sets.name LIKE ? AND images.status =  ?", CommonName+"%", models.ImageStatusBuilding)
+
+			count, err := service.GetImageSetsViewCount(dbFilter)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(count).To(Equal(int64(0)))
```
tomershinhar, line 119: shouldn't count be 1 here? as otherImage0 is with status BUILDING
i might be missing something though

ldjebran: yes it's not the last one :)

ldjebran: because otherImage0 is not the last one in image-set , it's not and should not be affected by the filler.

tomershinhar (approved):
LGTM

mgold1234 (approved):
/lgtm

pkg/services/imagesets.go
ldjebran: resolve this as implemented the unittests

ldjebran:
remove do not merge as seems everything ok now , with automation

PR #1804, opened 2023-01-10
image-sets: Fixes image-sets linter problems and remove any linter skipping.

ldjebran:
Fixing code to satisfy linters, remove skipping from image-sets files 

 
## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/imagesets.go
+++ b/pkg/routes/imagesets.go
@@ -1,5 +1,3 @@
-// FIXME: golangci-lint
-// nolint:gocritic,govet,revive
 package routes
 
 import (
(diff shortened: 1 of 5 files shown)


mgold1234 (approved):
/lgtm

pkg/routes/imagesets.go
```diff
@@ -576,9 +574,8 @@ func GetImageSetViewByID(w http.ResponseWriter, r *http.Request) {
 		// log and response handled by getContextImageSet
 		return
 	}
-	imagesDBFilters := imageDetailFilters(r, db.DB)
-	pagination := common.GetPagination(r)
-	imageSetIDView, err := ctxServices.ImageSetService.GetImageSetViewByID(imageSet.ID, pagination.Limit, pagination.Offset, imagesDBFilters)
+
+	imageSetIDView, err := ctxServices.ImageSetService.GetImageSetViewByID(imageSet.ID)
```
de1987, line 578: I'm concerned that you removed that `pagination`, I know this function is loading the image set by id but it might be paginating the images of the image sets

Did I miss anything?

de1987: So even if it's not being used, I think we should paginate it
WDYT?

ldjebran: That args are not used , there is no where in that code to use them, this args are here from a version where we had images returned with an image set and we had pagination , but we changed all that code but we forget to remove the not used args.

de1987: LGTM

de1987 (approved):

PR #1805, opened 2023-01-10
Change readiness probe to use metrics

thom-at-redhat:
# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [X] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/imagesets.go
+++ b/pkg/routes/imagesets.go
@@ -1,3 +1,4 @@
+// nolint:gocritic,govet,revive
 package routes
 
 import (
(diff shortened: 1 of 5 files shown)


pkg/routes/imagesets.go
```diff
@@ -1,3 +1,4 @@
+// nolint:gocritic,govet,revive
```
ldjebran, line 1: Do not think this is a good idea putting them again , if they are failing on 1.19 and we are running 1.18 we have to find a solution for this, do not think this is how it should be, having this skips, on each file of the project.

thom-at-redhat: we are running against Go 1.18 starting with next patch

ldjebran: as per our offline discussion, this is an intermediate stage , resolve comment

ldjebran (approved):
lgtm

PR #1820, opened 2023-01-16
Create a index to Name field on thirdpartyrepo model

adeliaferreira:
We have the option to query by name in the thirdpartyrepo end point. We had some issue related to latency on this endpoint while query by name, as an improvement to avoid this happen again in the future, this PR will create an index to name field as a way to improve our retrieve capacity. 

More discussion could be found on the Jira ticket THEEDGE-2913 

Signed-off-by: Adelia Ferreira <tay.figueira@gmail.com>

# Description

Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


omaciel:
Do you have any data you can share about the optimization before and after the indexing? I think it would really help with the review of this PR :)

adeliaferreira:
> Do you have any data you can share about the optimization before and after the indexing? I think it would really help with the review of this PR :)

Actually not, this suggestion to add the field as a index surface from the ticket discussion, once the issue is intermittent, it's hard to collect data about it. But on this case, due to fact we use this field as a parameter on sortBy, it's a good practice to have it as an index to improve this query performance on large databases  for high demanded fields and it was suggested by  @ldjebran on the ticket comment section.

omaciel:
Perhaps you could use this as an opportunity to figure out how we can collect performance data and use it to make a stronger case? I'm certain that indexing would help, but I also believe that we should be able to say "Adding the new index to <field> has improved the performance by 2500x." There's also the potential of having [too many indexes](https://sqlstudies.com/2016/10/13/too-many-indexes/).

omaciel:
@adeliaferreira my comment is not a blocker; I'm only asking you to look for ways to measure your fix. Definitely get someone else to review it

adeliaferreira:
> Perhaps you could use this as an opportunity to figure out how we can collect performance data and use it to make a stronger case? I'm certain that indexing would help, but I also believe that we should be able to say "Adding the new index to has improved the performance by 2500x." There's also the potential of having [too many indexes](https://sqlstudies.com/2016/10/13/too-many-indexes/).


After sync with the team, we define a strategy to measure it based on the time elapsed in the query, because it will directly been impact by the index creation for this case. As the worst case scenario, I create a database with 16984 random registers and try to look for a register that has a unique name, without the index, the same sql statement used in our system get 6ms to return the unique value after the index creating and clean any cache remain, the same query is returned in 2ms, so we're reducing 2/3 of the time on perform the query, when we use a orderby to get a alphabetic order, it reduce from 5 ms to 3 ms in the same query. An interesting thing here, the time without the index will be a exponential grow for the worst case scenario in bigger databases and this query will be heavily required once it's used in filter options on the frontend, so either if we add some latency on the add operation (not notice in the tests, still less than 1 ms to insert a new register after the index creation), the mostly use operation to this table (list, order and search) will have a reduce execution time, that will impact positively in the user experience.

omaciel (approved):
/lgtm

Thank you for doing the extra work and gather metrics for us. I hope it was a helpful experience.

PR #1824, opened 2023-01-18
dd unit test to event in common/kafka

mgold1234:

# Description

the coverage unit test for event.go is 0%.
this pr add unit test and increase the coverage to 100%
Fixes # (issue)

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:


pkg/common/kafka/events_test.go
```diff
@@ -0,0 +1,81 @@
...
+						RedHatOrgID: common.DefaultOrgID,
+						Source:      models.SourceEdgeEventAPI,
+						SpecVersion: "v1",
+						Subject:     image.Name,
+						Time:        time.Now().Format(time.RFC3339),
+						Type:        models.EventTypeEdgeImageRequested,
+					}
+					event := kafkacommon.CreateEdgeEvent(common.DefaultOrgID, models.SourceEdgeEventAPI, image.RequestID, models.EventTypeEdgeImageRequested, image.Name, edgePayload)
+					Expect(event.Data).To(Equal(cloudEvent.Data))
+					Expect(event.RedHatOrgID).To(Equal(cloudEvent.RedHatOrgID))
```
ldjebran, line 75: maybe like this: 
```suggestion
					Expect(event.Data).To(Equal(edgePayload))
					Expect(event.RedHatOrgID).To(Equal(common.DefaultOrgID))
```

and remove cloudEvent variable, 

what do you think ?

ldjebran (approved):
lgtm

pkg/common/kafka/events_test.go
```diff
@@ -0,0 +1,73 @@
...
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		config.Init()
+
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+	Describe("Get Consumer", func() {
```
tomershinhar, line 30: i am not sure i understand the naming here, this test is about creating an event, not getting a consumer so "Get consumer" is not correct.

also "One does not exist yet" seems to be unrelated to the test.
as a final note, i believe there is no need fo both `When` and `Context` here, you can drop `Context`

de1987: adding more
if your goal is to test just function I would suggest using something like this
`#CreateEdgeEvent`

tomershinhar (approved):
LGTM

ldjebran (approved):

pkg/common/kafka/events_test.go
ldjebran: I believe that this is resolved

PR #1838, opened 2023-01-24
routes-services: Fixes gomock ctrl.Finish()

ldjebran:
The ctrl.Finish() in the current implementation happen just after executing the BeforeEach function,
 but should happen after the test has been executed eg: in the AfterEach function.
 This PR fixes the mock calls counting, and the wrong's tests behavior caused.

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/devices_test.go
+++ b/pkg/routes/devices_test.go
@@ -35,12 +35,12 @@ var _ = Describe("Devices Router", func() {
 	var deviceUUID string
 	var mockDeviceService *mock_services.MockDeviceServiceInterface
 	var router chi.Router
+	var ctrl *gomock.Controller
 
 	BeforeEach(func() {
 		// Given
 		deviceUUID = faker.UUIDHyphenated()
-		ctrl := gomock.NewController(GinkgoT())
-		defer ctrl.Finish()
+		ctrl = gomock.NewController(GinkgoT())
 
 		mockDeviceService = mock_services.NewMockDeviceServiceInterface(ctrl)
 		mockServices := &dependencies.EdgeAPIServices{
@@ -57,6 +57,11 @@ var _ = Describe("Devices Router", func() {
 		})
 		router.Route("/devices", MakeDevicesRouter)
 	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
 	Context("get available updates", func() {
 		var req *http.Request
 
@@ -67,10 +72,12 @@ var _ = Describe("Devices Router", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			It("should give an error", func() {
-				mockDeviceService.EXPECT().GetUpdateAvailableForDeviceByUUID(gomock.Eq(deviceUUID), false, 30, 0).Return(nil, int64(0), new(services.DeviceNotFoundError))
 				recorder := httptest.NewRecorder()
 				router.ServeHTTP(recorder, req)
 				Expect(recorder.Code).To(Equal(http.StatusBadRequest))
+				respBody, err := io.ReadAll(recorder.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(respBody)).To(ContainSubstring("DeviceUUID must be sent"))
 			})
 		})
 		When("device UUID is passed", func() {
@@ -110,10 +117,11 @@ var _ = Describe("Devices Router", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			It("should give an error", func() {
-				mockDeviceService.EXPECT().GetUpdateAvailableForDeviceByUUID(gomock.Eq(deviceUUID), true, 30, 0).Return(nil, int64(0), new(services.DeviceNotFoundError))
 				recorder := httptest.NewRecorder()
 				router.ServeHTTP(recorder, req)
-				Expect(recorder.Code).To(Equal(http.StatusBadRequest))
+				respBody, err := io.ReadAll(recorder.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(respBody)).To(ContainSubstring("DeviceUUID must be sent"))
 			})
 		})
 		When("device UUID is passed", func() {
@@ -169,11 +177,11 @@ var _ = Describe("Devices View Router", func() {
 	var mockDeviceService *mock_services.MockDeviceServiceInterface
 	var router chi.Router
 	var mockServices *dependencies.EdgeAPIServices
+	var ctrl *gomock.Controller
 
 	BeforeEach(func() {
 		// Given
-		ctrl := gomock.NewController(GinkgoT())
-		defer ctrl.Finish()
+		ctrl = gomock.NewController(GinkgoT())
 
 		mockDeviceService = mock_services.NewMockDeviceServiceInterface(ctrl)
 		mockServices = &dependencies.EdgeAPIServices{
@@ -190,6 +198,11 @@ var _ = Describe("Devices View Router", func() {
 		})
 		router.Route("/devicesview", MakeDevicesRouter)
 	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
 	Context("get devicesview", func() {
 
 		When("when devices are not found", func() {
--- a/pkg/routes/devices_test.go
+++ b/pkg/routes/devices_test.go
@@ -119,6 +119,7 @@ var _ = Describe("Devices Router", func() {
 			It("should give an error", func() {
 				recorder := httptest.NewRecorder()
 				router.ServeHTTP(recorder, req)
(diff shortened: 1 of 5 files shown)


pkg/routes/devices_test.go
```diff
@@ -67,10 +72,12 @@ var _ = Describe("Devices Router", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			It("should give an error", func() {
-				mockDeviceService.EXPECT().GetUpdateAvailableForDeviceByUUID(gomock.Eq(deviceUUID), false, 30, 0).Return(nil, int64(0), new(services.DeviceNotFoundError))
```
omaciel, line 70: I'm sure there is a valid reason, but I don't have any context to determine why this was removed since the PR is about ensuring that `ctrl.Finish()` is executed `AfterEach` test :/ We just don't want to mock this anymore, is that it?

pkg/routes/devices_test.go
```diff
@@ -110,10 +117,11 @@ var _ = Describe("Devices Router", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			It("should give an error", func() {
-				mockDeviceService.EXPECT().GetUpdateAvailableForDeviceByUUID(gomock.Eq(deviceUUID), true, 30, 0).Return(nil, int64(0), new(services.DeviceNotFoundError))
 				recorder := httptest.NewRecorder()
 				router.ServeHTTP(recorder, req)
-				Expect(recorder.Code).To(Equal(http.StatusBadRequest))
```
omaciel, line 116: We're no longer expecting an error? The test claims `It("should give an error"`

pkg/services/images_test.go
```diff
@@ -1413,7 +1418,7 @@ var _ = Describe("Image Service Test", func() {
 				Expect(image.ImageSetID).To(BeNil())
 			})
 		})
-		When("When image-builder SearchPackage fail", func() {
+		When("When image-builder Validation fail", func() {
```
omaciel, line 1421: As this block of tests was updated: I'm not sure I have the context to determine why the tests were changed from `SearchPackage` to `Validation.`

pkg/services/images_test.go
```diff
@@ -1423,19 +1428,18 @@ var _ = Describe("Image Service Test", func() {
...
-				Expect(err.Error()).To(Equal(expectedErr.Error()))
+				Expect(err.Error()).To(Equal("value is not one of the allowed values"))
 				// image is not created
 				Expect(image.ID).To(Equal(uint(0)))
 				// But imageSet is not Created
 				Expect(image.ImageSetID).To(BeNil())
 			})
 		})
-		When("When image-builder SearchPackage fail", func() {
+		When("When image-builder Validation fail", func() {
```
omaciel, line 1442: As this block of tests was updated: I'm not sure I have the context to determine why the tests were changed from `SearchPackage` to `Validation`.

pkg/services/updates_test.go
```diff
@@ -956,13 +970,15 @@ var _ = Describe("UpdateService Basic functions", func() {
...
 
 				mockImageService.EXPECT().GetImageByOSTreeCommitHash(currentCommit.OSTreeCommit).Return(&currentImage, nil)
-				mockImageService.EXPECT().GetImageByOSTreeCommitHash(commit.OSTreeCommit).Return(&image, nil)
 				mockImageService.EXPECT().GetImageByOSTreeCommitHash(latestCommit.OSTreeCommit).Return(&latestImage, nil)
+
 				mockProducer.EXPECT().Produce(gomock.Any(), gomock.Any()).Return(nil)
 				mockProducerService.EXPECT().GetProducerInstance().Return(mockProducer)
 
-				upd, err := updateService.BuildUpdateTransactions(&devicesUpdate, orgID, &commit)
+				upd, err := updateService.BuildUpdateTransactions(&devicesUpdate, orgID, &latestCommit)
```
omaciel, line 978: What is the context for this specific change?

pkg/services/updates_test.go
```diff
@@ -1013,7 +1033,7 @@ var _ = Describe("UpdateService Basic functions", func() {
 
 		latestCommit := models.Commit{OrgID: orgID, OSTreeCommit: faker.UUIDHyphenated(), ChangesRefs: false}
 		db.DB.Create(&latestCommit)
-		latestImage := models.Image{OrgID: orgID, CommitID: latestCommit.ID, ImageSetID: &imageSet.ID, Status: models.ImageStatusSuccess, Distribution: "rhel-90"}
+		latestImage := models.Image{OrgID: orgID, CommitID: latestCommit.ID, ImageSetID: &imageSet.ID, Status: models.ImageStatusSuccess, Distribution: "rhel-91"}
```
omaciel, line 1036: What is the context for this change?

pkg/services/updates_test.go
```diff
@@ -1054,13 +1074,14 @@ var _ = Describe("UpdateService Basic functions", func() {
 				mockInventoryClient.EXPECT().ReturnDevicesByID(device.UUID).Return(resp, nil)
 
 				mockImageService.EXPECT().GetImageByOSTreeCommitHash(currentCommit.OSTreeCommit).Return(&currentImage, nil)
-				mockImageService.EXPECT().GetImageByOSTreeCommitHash(commit.OSTreeCommit).Return(&image, nil)
 				mockImageService.EXPECT().GetImageByOSTreeCommitHash(latestCommit.OSTreeCommit).Return(&latestImage, nil)
 				mockProducer.EXPECT().Produce(gomock.Any(), gomock.Any()).Return(nil)
 				mockProducerService.EXPECT().GetProducerInstance().Return(mockProducer)
 
-				upd, err := updateService.BuildUpdateTransactions(&devicesUpdate, orgID, &commit)
+				upd, err := updateService.BuildUpdateTransactions(&devicesUpdate, orgID, &latestCommit)
```
omaciel, line 1081: What is the context for this specific change?

omaciel (commented):
All changes related to `ctrl.Finish()` are pretty straight forward to follow and understand, but some of the refactoring outside of the scope of the PR was hard for me to follow :/

pkg/routes/devices_test.go
ldjebran: now gomock is calculating how much time we calling this function and validate accordingly,

this end point in the current scenario does reach the place where we are calling  GetUpdateAvailableForDeviceByUUID, and this function is never called as we return early see function DeviceCtx at routes/devices.go.

pkg/routes/devices_test.go
ldjebran: it was removed by accident, thanks to have noticed this.

pkg/services/images_test.go
ldjebran: the search can not fail if it has never called, because the validation fails before reaching search function.

pkg/services/images_test.go
ldjebran: the validation fails before reaching search function.

ldjebran: search function is never called,

pkg/services/updates_test.go
ldjebran: initially we have three images
1- currentImage  with currentCommit
2 - image with commit
3 - latestImage with lastestCommit

we have a device with currentImage  that we want to update,

GetImageByOSTreeCommitHash is called only for the initial currentCommit and the target image commit 
so one of the mockImageService.EXPECT().GetImageByOSTreeCommitHash has to be removed 

I found the test more adequate if we update to latestCommit  instead of the intermediate image commit.


I can revert to update to image commit, in this case , we need to remove 
mockImageService.EXPECT().GetImageByOSTreeCommitHash(latestCommit.OSTreeCommit).Return(&latestImage, nil)
 
This is simply a matter of choice.

pkg/services/updates_test.go
ldjebran: you are right, as the test say "Update Devices to same distribution" will revert back to 90

pkg/services/updates_test.go
ldjebran: same think here, one of the mockImageService.EXPECT().GetImageByOSTreeCommitHash need to be removed 

I chosen to remove the intermediate image commit

ldjebran:
Thanks @omaciel for the review , responded to your comments, and reverted changes for 2 of them.
Thanks to revisit.

pkg/services/updates_test.go
omaciel: I trust your judgment here :)

pkg/services/updates_test.go
omaciel: Thank you.

omaciel (approved):
/lgtm

de1987 (approved):

PR #1857, opened 2023-02-01
Fix Inconsistent count returned from /image-sets 

Saurabh-Thakre:
# Description
Fixing  Inconsistent count returned from /image-sets and /image-sets/view

This PR fixes two issues, 

1. First 

The GET /image-sets and GET /image-sets/view APIs return different count values. For example:
```
$ $EDGE/image-sets?status=error | jq '.Count'
6629

$ $EDGE/image-sets/view?status=error | jq '.count'
5857
```

2. Second

And the Count of images-sets endpoint does not respect the Data from the same.
As an example,  if we apply the filter on image-sets api like `/image-sets?status=error`, the data shows output including the images which has status success. 
This also fixes the above issue. 

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


pkg/routes/imagesets.go
```diff
@@ -206,8 +206,12 @@ func ListAllImageSets(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	countResult := imageSetFilters(r, db.OrgDB(orgID, db.DB, "Image_Sets").Debug().Model(&models.ImageSet{})).
-		Joins(`JOIN Images ON Image_Sets.id = Images.image_set_id AND Images.deleted_at is NULL`).Distinct("image_sets.id").Count(&count)
+	latestImagesSubQuery := db.Org(orgID, "").Model(&models.Image{}).Select("image_set_id", "max(id) as image_id").Group("image_set_id")
+	countResult := imageSetFilters(r, db.OrgDB(orgID, db.DB, "image_sets")).Table("(?) as latest_images", latestImagesSubQuery).
+		Joins("JOIN images on images.id = latest_images.image_id").
+		Joins("JOIN image_sets on image_sets.id = latest_images.image_set_id").
+		Where("image_sets.deleted_at IS NULL").
+		Count(&count)
```
ldjebran, line 214: This count has to be the same as for the records count in query at line https://github.com/RedHatInsights/edge-api/pull/1857/files#diff-7d067fdfe152f3f147f9e7772b4eab0b02785685caf2348b0beef7fd46be1e25R222

so you have to fund a solution that satisfy the two.

pkg/routes/imagesets.go
```diff
@@ -206,42 +194,36 @@ func ListAllImageSets(w http.ResponseWriter, r *http.Request) {
...
+	result = imageSetFilters(r, db.OrgDB(orgID, db.DB, "Image_Sets").Debug().Model(&models.ImageSet{})).Table("(?) as latest_images", latestImagesSubQuery).
+		Distinct("image_sets.*").
+		Limit(pagination.Limit).Offset(pagination.Offset).
+		Preload("Images").
+		Preload("Images.Commit").
+		Preload("Images.Installer").
+		Preload("Images.Commit.Repo").
+		Joins("JOIN images on images.id = latest_images.image_id").
+		Joins("JOIN image_sets on image_sets.id = latest_images.image_set_id").
+		Find(&imageSet)
```
de1987, line 219: is there a test covering these changes?

de1987 (changes requested):

pkg/routes/imagesets.go
ldjebran: @Saurabh-Thakre you should write a unittest that should demonstrate that your fix is working (of course it will fail without your changes) 

The test for example can have this scenario

Create 3 ImageSets 
1 imageSet has one image with Error status
2 imageset has one image without success status
3 imageset has 2 images  the latest one has success status , the first one has error status

filter by status=error => result should be only one imageset `the first one`

This scenario will fail for the current implementation , and should pass with your fix.


you can create a second test that filter by status=success that should return 2 image-sets (number 2 and number 3).

Saurabh-Thakre: gotcha, I wrote it but then I saw we already have some tests around filter by status so I removed. I will put it back

pkg/routes/imagesets_test.go
```diff
@@ -562,6 +584,23 @@ var _ = Describe("ImageSets Route Test", func() {
...
+				req, err := http.NewRequest("GET", fmt.Sprintf("/image-sets?status=%s", status), nil)
+				Expect(err).ToNot(HaveOccurred())
+				w := httptest.NewRecorder()
+				req = req.WithContext(dependencies.ContextWithServices(req.Context(), &dependencies.EdgeAPIServices{}))
+				handler := http.HandlerFunc(ListAllImageSets)
+				handler.ServeHTTP(w, req)
+				Expect(w.Code).To(Equal(http.StatusOK), fmt.Sprintf("expected status %d, but got %d", w.Code, http.StatusOK))
+				respBody, err := io.ReadAll(w.Body)
+				Expect(err).To(BeNil())
+				Expect(string(respBody)).To(ContainSubstring("image-set-1"))
```
ldjebran, line 599: I think we need to unmarshal this respBody and analyse how much image-sets we received and go through all image-sets returned to find image-set-1 and image-set-2 , in error messages we can do this , but here I think we need to be more explicit, as we are building a functionality contract .

ldjebran (commented):
loog good, left one comment about the unittest.

pkg/routes/imagesets_test.go
ldjebran: hint:
 you can create the image names having all the same prefix , so that you can find them
```go
prefix := faker.UUIDHyphenated() 
image1 := models.Image{
    ...
    Name: prefix + "image-set-1" 
    ...
}

// and then filter by name and status 
req, err := http.NewRequest("GET", fmt.Sprintf("/image-sets?"?name=?&status=?",prefix, status), nil)
// this should return all imagesets that name contain prefix value and the status 
```

pkg/routes/imagesets_test.go
```diff
@@ -562,6 +584,85 @@ var _ = Describe("ImageSets Route Test", func() {
 				Expect(string(respBody)).ToNot(ContainSubstring("image-set-1"))
 			})
 		})
+
+	})
+	Context("Image-sets filter", func() {
+		BeforeEach(func() {
+			imageSet1 := &models.ImageSet{
+				Name:  "imageSet-1",
```
ldjebran, line 592: think a more random prefix must be used here (as if any one will use this name in tests , this test will fail):
```go
var namePrefix string 

BeforeEach(func() {
     namePrefix := faker.UUIDHyphenated()
     imageSet1 := &models.ImageSet{
				Name:  namePrefix + "imageSet-1",
				....
// in request filter by prefix
req, err := http.NewRequest("GET", fmt.Sprintf("/image-sets?status=%s&name=namePrefixt", status), nil)

```

pkg/routes/imagesets_test.go
```diff
@@ -529,6 +529,28 @@ var _ = Describe("ImageSets Route Test", func() {
 			}
 			db.DB.Create(&imageSuccess)
 			db.DB.Create(&imageError)
+
+			imageSet3 := &models.ImageSet{
```
ldjebran, line 533: maybe these data should be removed ?

adeliaferreira (approved):
LGTM

PR #1877, opened 2023-02-09
add unit test to repo.

mgold1234:
repo has 11% of covarage, so this pr will increase the covarage

# Description

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor


pkg/services/repo_test.go
```diff
@@ -0,0 +1,54 @@
...
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+	Describe("test repo id", func() {
+		When("is valid", func() {
+			It("repo id create and tested", func() {
+				id, _ := faker.RandomInt(1)
+				uid := uint(id[0])
+				imageSet := &models.ImageSet{OrgID: orgID}
+				image := &models.Image{
```
ldjebran, line 36: why we need image to retieve a repo ?

pkg/services/repo_test.go
```diff
@@ -0,0 +1,41 @@
...
+		repoService = services.NewRepoService(ctx, log.NewEntry(log.StandardLogger()))
+
+	})
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+	Describe("test repo id", func() {
+		When("is valid", func() {
+			It("repo id create and tested", func() {
+				id, _ := faker.RandomInt(1)
```
ldjebran, line 32: as understand this, pick a random number , and that id may exist in database.
for testing and for happy , we should be sure that  the record exists.
Imagine the database is empty, so this test should fail,  so it can an repo id that have status  Error

ldjebran (changes requested):
1. Need to create a real repo. 
2. Check the error first before any other check.

pkg/services/repo_test.go
```diff
@@ -0,0 +1,52 @@
...
+				testRepo = models.Repo{
+					URL:    "www.test.com",
+					Status: models.RepoStatusSuccess,
+				}
+				result := db.DB.Create(&testRepo)
+				Expect(result.Error).ToNot(HaveOccurred())
+				repo, err := repoService.GetRepoByID(&testRepo.ID)
+				Expect(repo.Status).To(Equal(testRepo.Status))
+				Expect(repo.Model.ID).To(Equal(testRepo.Model.ID))
+				Expect(err).To(BeNil())
```
ldjebran, line 40: the error assert must just after the the function call
```go
repo, err := repoService.GetRepoByID(&testRepo.ID)
Expect(err).To(BeNil())
```

pkg/services/repo_test.go
```diff
@@ -0,0 +1,52 @@
...
+			})
+		})
+		When("is not valid", func() {
+			It("repo id doesnt create", func() {
+				var id1 *uint
+				repo, err := repoService.GetRepoByID(id1)
+				Expect(repo).To(BeNil())
+				Expect(err).ToNot(BeNil())
```
ldjebran, line 48: the error assert must just after the the function call
```go
repo, err := repoService.GetRepoByID(id1)
Expect(err).To(BeNil())
```

pkg/services/repo_test.go
```diff
@@ -0,0 +1,52 @@
...
+				result := db.DB.Create(&testRepo)
+				Expect(result.Error).ToNot(HaveOccurred())
+				repo, err := repoService.GetRepoByID(&testRepo.ID)
+				Expect(repo.Status).To(Equal(testRepo.Status))
+				Expect(repo.Model.ID).To(Equal(testRepo.Model.ID))
+				Expect(err).To(BeNil())
+			})
+		})
+		When("is not valid", func() {
+			It("repo id doesnt create", func() {
```
ldjebran, line 44: what is the test is testing, when passing a nil id?

mgold1234: repo id doesnt create and verify that repo id doesnt exist in case of sending empty id, WDYT?

ldjebran: I think that , this a case of two negative unitests, one with id nil , and one with id not nil but does not exist, as they return different errors. 

I think also that function GetRepoByID should validate id to be not nil before passing it to gorm select function.
eg:

```go

// GetRepoByID receives RepoID uint and get a *models.Repo back
func (s *RepoService) GetRepoByID(repoID *uint) (*models.Repo, error) {
	if repoID == nil {
		return nil, new(ImageRepositoryIsUndefined)
	}
	s.log.Debug("Retrieving repo by ID")
        ...
}
```

ldjebran: what dou think ?

pkg/services/repo_test.go
```diff
@@ -0,0 +1,61 @@
...
+				Expect(err).To(BeNil())
+				Expect(repo.Status).To(Equal(testRepo.Status))
+				Expect(repo.Model.ID).To(Equal(testRepo.Model.ID))
+			})
+		})
+		When("is not valid with nil value", func() {
+			It("repo id doesnt create and verify that we dont get repo id with nil vale", func() {
+				var id1 *uint
+				repo, err := repoService.GetRepoByID(id1)
+				Expect(err.Error()).To(Equal("unsupported data"))
```
ldjebran, line 47:
```suggestion
				Expect(result.Error).To(HaveOccurred())
				Expect(err.Error()).To(Equal("unsupported data"))
```

pkg/services/repo_test.go
```diff
@@ -0,0 +1,61 @@
...
+				Expect(repo).To(BeNil())
+			})
+		})
+		When("is not valid with repo id that doesnt exist", func() {
+			It("repo id doesnt create and verify that repo id doesnt exist in db", func() {
+				id := "999999"
+				uid := uint(id[0])
+				repo, err := repoService.GetRepoByID(&uid)
+				Expect(err.Error()).To(Equal("record not found"))
```
ldjebran, line 56:
```suggestion
				Expect(err).To(HaveOccurred())
				Expect(err.Error()).To(Equal("record not found"))
```

pkg/services/repo_test.go
```diff
@@ -0,0 +1,63 @@
...
+				Expect(repo.Model.ID).To(Equal(testRepo.Model.ID))
+			})
+		})
+		When("is not valid with nil value", func() {
+			It("repo id doesnt create and verify that we dont get repo id with nil vale", func() {
+				var id1 *uint
+				repo, err := repoService.GetRepoByID(id1)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("unsupported data"))
+				Expect(repo).To(BeNil())
```
ldjebran, line 49: remove this line

pkg/services/repo_test.go
```diff
@@ -0,0 +1,63 @@
...
+			})
+		})
+		When("is not valid with repo id that doesnt exist", func() {
+			It("repo id doesnt create and verify that repo id doesnt exist in db", func() {
+				id := "999999"
+				uid := uint(id[0])
+				repo, err := repoService.GetRepoByID(&uid)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("record not found"))
+				Expect(repo).To(BeNil())
```
ldjebran, line 59: remove this line

mgold1234:
@djebran for some reason the test with non existed id always fail when I run it in githlab, because somehow this id is found in the db.
so I remove this test, or maybe you have any idea for an id that does not exist?

ldjebran (changes requested):
add Expect(err).To(HaveOccurred())
as per comments

pkg/services/repo_test.go
mgold1234: do you think its ok to change the existing code now? or its ok to refactor some places?

ldjebran (approved):
lgtm

acosferreira (approved):
lgtm

PR #1885, opened 2023-02-14
Test scan

acosferreira:
# Description
Add feature flag to validate package creation and avoid installedPackage growht.
Add omit to avoid dup 
Create commit_installed_package to be able to reuse installed_package_id from previous package created
Added test scenario
The goal is reduce the exponential data creation, we should do one more refactor to fix m2m relationship and clean up the 26gb of data

FIXES: THEEDGE-2648/2469

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


pkg/clients/imagebuilder/client.go
```diff
@@ -137,22 +146,6 @@ type S3UploadStatus struct {
 	URL string `json:"url"`
 }
 
-// MetaCount contains Count of a SearchPackageResult
```
ldjebran, line 140: why we need to move this structs from this module , their declaration in this module is very natural
as they are created in this module, and their existence is related to this module  ?

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +575,25 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
...
 		c.log.WithField("error", err.Error()).Error(new(PackageRequestError))
 		return nil, err
 	}
 	return &searchResult, nil
 }
+
+func validatePackages(pkg InstalledPackage) uint {
+	// think to use (Where("name ,release" in ((?,?)) )
+	var result *models.InstalledPackage
+	db.DB.Table("Installed_Packages").Select("ID, name,release, arch, version, epoch").
```
ldjebran, line 590: missing error handling

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,50 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	var dupPackages []uint
 	for n := range metadata.InstalledPackages {
-		pkg := models.InstalledPackage{
-			Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
-			Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
-			Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
-			Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+
+		if feature.DedupPackage.IsEnabled() {
+			pkg := validatePackages(metadata.InstalledPackages[n])
```
ldjebran, line 465: this may be very inefficient, as if we have 1000 packages , we will do 1000 sql queries. 
I think about a function that we pass all the packages and return a map[int64]*models.InstalledPackage
where the key is the index in metadata.InstalledPackages

what do you think ?

acosferreira: yeah, i left a comment on validatePackage function about it
I'll see if I can make it using in condition like (name, release) in ('a', 'b')

ldjebran: I think we go with something like this: 

```go
func GetInstalledPackagesMap(installedPackages []models.InstalledPackage) (map[int64]models.InstalledPackage, error) {
	installedPackagesMap := make(map[int64]models.InstalledPackage, len(installedPackages))
	if len(installedPackages) == 0 {
		return installedPackagesMap, nil
	}
	// build md5 slice
	packagesMD5 := make([]string, 0, len(installedPackages))
	for _, installPackage := range installedPackages {
		packagesMD5 = append(packagesMD5, installPackage.Sigmd5)
	}
	// prepare 
	var md5Packages models.InstalledPackage
	if err := db.DB.Where("sigmd5", packagesMD5).Find(&md5Packages).Error; err != nil {
		return installedPackagesMap, err
	}

	... 
	
	return installedPackagesMap, nil
}

```

we may need two maps to resolve this

pkg/clients/imagebuilder/client.go
```diff
@@ -438,13 +432,15 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
-	res, err := client.Do(req)
+
+	res, err := ImageBuilderHTTPClient.Do(req)
+
 	if err != nil {
 		c.log.WithField("error", err.Error()).Error("Image Builder GetMetadata Request Error")
 		return nil, err
 	}
 	respBody, err := io.ReadAll(res.Body)
+	fmt.Printf("\n body %v", respBody)
```
omaciel, line 443: Do you want this printed or logged?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +459,69 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+			} else {
+				pkg := models.InstalledPackage{
+					Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
+					Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
+					Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
+					Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+				}
+				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
+
+				db.DB.Debug().Omit("Image.InstalledPackages.*").Save(image.Commit)
```
omaciel, line 497: Do we need `Debug()` here? It'll probably speed things up without it.

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +459,69 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
 	image.Commit.OSTreeCommit = metadata.OstreeCommit
+
+	if feature.DedupPackage.IsEnabled() &&
+		len(dupPackages) > 0 {
+		for i := range dupPackages {
+			// build batch
+			cip = append(cip, models.CommitInstalledPackages{InstalledPackageId: dupPackages[i], CommitId: image.Commit.ID})
+		}
+
+		if len(cip) > 0 {
```
omaciel, line 520: Since you checked `len(dupPackages) > 0` above, is it necessary to check that `len(cip) > 0`?

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +595,34 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
...
 		return nil, err
 	}
 	return &searchResult, nil
 }
+
+func (c *Client) ValidatePackages(pkgs []string) (map[uint]*models.InstalledPackage, error) {
+	var result []models.InstalledPackage
+	setOfPackages := make(map[uint]*models.InstalledPackage)
+
+	err := db.DB.Debug().Table("Installed_Packages").Select("ID, name,release, arch, version, epoch").
```
omaciel, line 611: Remove `Debug()`?

pkg/clients/imagebuilder/client_test.go
```diff
@@ -3,24 +3,28 @@
...
 	"testing"
 	"time"
 
 	"github.com/bxcodec/faker/v3"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	log "github.com/sirupsen/logrus"
+	// "github.com/redhatinsights/edge-api/pkg/clients/imagebuilder/mock_imagebuilder"
```
omaciel, line 23: Remove comment?

pkg/clients/imagebuilder/client_test.go
```diff
@@ -517,4 +534,125 @@ var _ = Describe("Image Builder Client Test", func() {
...
+					},
+				}
+
+				img, a := client.GetMetadata(img)
+
+				Expect(a).ToNot(HaveOccurred())
+				Expect(img).ToNot(BeNil())
+				Expect(len(img.Commit.InstalledPackages)).To(Equal(2))
+				var s int64
+				db.DB.Model(models.CommitInstalledPackages{}).Where("Commit_Id=?", img.Commit.ID).Debug().Count(&s)
```
omaciel, line 607: Debug?

pkg/models/commits.go
```diff
@@ -42,7 +42,7 @@ type Commit struct {
 	BuildNumber          uint               `json:"BuildNumber"`
 	BlueprintToml        string             `json:"BlueprintToml"`
 	Arch                 string             `json:"Arch"`
-	InstalledPackages    []InstalledPackage `json:"InstalledPackages,omitempty" gorm:"many2many:commit_installed_packages;"`
+	InstalledPackages    []InstalledPackage `faker:"-" json:"InstalledPackages,omitempty" gorm:"many2many:commit_installed_packages;"`
```
omaciel, line 45: What does `faker:"-"` do here?

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +595,34 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
...
+	var searchResult models.SearchPackageResult
 	err = json.Unmarshal(respBody, &searchResult)
 	if err != nil {
 		c.log.WithField("error", err.Error()).Error(new(PackageRequestError))
 		return nil, err
 	}
 	return &searchResult, nil
 }
+
+func (c *Client) ValidatePackages(pkgs []string) (map[uint]*models.InstalledPackage, error) {
```
omaciel, line 607: Needs unittest

pkg/clients/imagebuilder/client.go
```diff
@@ -438,13 +432,15 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
 		req.Header.Add(key, value)
 	}
 	req.Header.Add("Content-Type", "application/json")
-	client := &http.Client{}
-	res, err := client.Do(req)
+
+	res, err := ImageBuilderHTTPClient.Do(req)
+
 	if err != nil {
```
omaciel, line 438: Now that you have mocked `http.Client.Do`, how about adding a test where `Do` returns an error?

omaciel (changes requested):
Left you some comments and suggestions

pkg/clients/imagebuilder/client.go
acosferreira: explaint in a call ;)

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,72 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+			} else {
+				pkg := models.InstalledPackage{
+					Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
+					Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
+					Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
+					Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+				}
+				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
+
+				db.DB.Omit("Image.InstalledPackages.*").Save(image.Commit)
```
ldjebran, line 497: if we are saving image.commit n times ?

ldjebran: may be this should be like this:
```go
if feature.DedupPackage.IsEnabled()  && concatPkg[metadata.InstalledPackages[n].Name] != "" {
    continue
} else {
   ....
}
```

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,72 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
-			Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
-			Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
-			Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+		metadataPackages = append(metadataPackages, fmt.Sprintf("%s-%s-%s", metadata.InstalledPackages[n].Name, metadata.InstalledPackages[n].Release, metadata.InstalledPackages[n].Version))
+	}
+
+	var pkg map[uint]*models.InstalledPackage
+	var cip []models.CommitInstalledPackages
+
+	concatPkg := make(map[string]string)
```
ldjebran, line 471: we are using concatPkg as to check if package already in dataBase 
maybe  it should be 
```go
packagesAlreadyExistsMap :=   make(map[string]bool)
```
when setting 
```go
oncatPkg[pkg[key].Name] = true
```

and when checking 

```go
if packagesAlreadyExistsMap[metadata.InstalledPackages[n].Name] == true {
...

```
this is more readable , what do you think?

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +597,32 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
...
 	return &searchResult, nil
 }
+
+func (c *Client) ValidatePackages(pkgs []string) (map[uint]*models.InstalledPackage, error) {
+	var result []models.InstalledPackage
+	setOfPackages := make(map[uint]*models.InstalledPackage)
+
+	err := db.DB.Table("Installed_Packages").Select("ID, name,release, arch, version, epoch").
+		Where("( (name || '-' || release || '-' ||  version)) in (?)", pkgs).
+		Find(&result)
```
ldjebran, line 615: maybe like this will more naturel 
```go
Find(&result).Error
if err != nil {
....

```

omaciel (changes requested):

pkg/clients/imagebuilder/client_test.go
```diff
@@ -517,4 +532,213 @@ var _ = Describe("Image Builder Client Test", func() {
...
+					fmt.Sprintf("%s-%s-%s", metadata.InstalledPackages[n].Name, metadata.InstalledPackages[n].Release, metadata.InstalledPackages[n].Version))
+			}
+
+			a, err := client.ValidatePackages(metadataPackages)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(a).NotTo(BeNil())
+			Expect(a[1].Name).To(Equal("rhc"))
+		})
+
+		It("should return nil ", func() {
```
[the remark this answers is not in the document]
acosferreira: Added scenario

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	var cip []models.CommitInstalledPackages
+
+	packagesExistsMap := make(map[string]bool)
+
+	if feature.DedupPackage.IsEnabled() {
+		pkg, err = c.ValidatePackages(metadataPackages)
+		if len(pkg) > 0 && err == nil {
+			for key := range pkg {
+				dupPackages = append(dupPackages, key)
+				packagesExistsMap[pkg[key].Name] = true
```
omaciel, line 478: It looks to me that `packagesExistsMap` will always and only filled with `true` value, so why not consider the value returned from `c.ValidatePackages(metadataPackages)` as the source of truth?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+
 	for n := range metadata.InstalledPackages {
-		pkg := models.InstalledPackage{
-			Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
-			Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
-			Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
-			Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+
+		if feature.DedupPackage.IsEnabled() {
+			if packagesExistsMap[metadata.InstalledPackages[n].Name] {
```
omaciel, line 486: Instead of checking if the package exists, perhaps you could check the opposite?

```go
if found, ok := packagesExistsMap[metadata.InstalledPackages[n].Name]; !ok {
    pkg := models.InstalledPackage{...}
}
```

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +596,31 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
...
+	var result []*models.InstalledPackage
+	setOfPackages := make(map[uint]*models.InstalledPackage)
+
+	if err := db.DB.Table("Installed_Packages").
+		Where("( (name || '-' || release || '-' ||  version)) in (?)", pkgs).
+		Find(&result); err.Error != nil {
+		c.log.WithField("error", err.Error)
+		return nil, err.Error
+	} else {
+		if len(result) > 0 {
```
omaciel, line 618: Is `result` always an iterable? If so, then it should be safe not to check its length and `for n:= range results` should either go through all items or just skip if results is empty (unless it's `nil`?)

pkg/clients/imagebuilder/client_test.go
```diff
@@ -517,4 +532,220 @@ var _ = Describe("Image Builder Client Test", func() {
...
+					fmt.Sprintf("%s-%s-%s", metadata.InstalledPackages[n].Name, metadata.InstalledPackages[n].Release, metadata.InstalledPackages[n].Version))
+			}
+
+			a, err := client.ValidatePackages(metadataPackages)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(a).To(BeNil())
+		})
+
+		It("should return error ", func() {
+			db.DB.Exec("ALTER TABLE installed_packages DROP COLUMN name")
```
omaciel, line 744: Won't this cause major headache if run in parallel or if the order ever changes or if another test needs to save to that table?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+	for n := range metadata.InstalledPackages {
+		metadataPackages = append(metadataPackages, fmt.Sprintf("%s-%s-%s", metadata.InstalledPackages[n].Name, metadata.InstalledPackages[n].Release, metadata.InstalledPackages[n].Version))
+	}
+
+	var pkg map[uint]*models.InstalledPackage
+	var cip []models.CommitInstalledPackages
+
+	packagesExistsMap := make(map[string]bool)
+
+	if feature.DedupPackage.IsEnabled() {
```
omaciel, line 473: Unittest covers this branch?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+		pkg, err = c.ValidatePackages(metadataPackages)
+		if len(pkg) > 0 && err == nil {
```
omaciel, line 475: Unittest covers this branch?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
-		pkg := models.InstalledPackage{
-			Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
-			Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
-			Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
-			Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+
+		if feature.DedupPackage.IsEnabled() {
+			if packagesExistsMap[metadata.InstalledPackages[n].Name] {
+				continue
+			} else {
```
omaciel, line 488: Unittest covers this branch?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+				pkg := models.InstalledPackage{
+					Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
+					Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
+					Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
+					Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+				}
+				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
+			}
+
+		} else {
```
omaciel, line 498: Unittest covers this branch?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,71 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+				Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+			}
+			image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
 		}
-		image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
 	}
+
 	image.Commit.OSTreeCommit = metadata.OstreeCommit
+	db.DB.Omit("Image.InstalledPackages.*").Save(image.Commit)
+	if feature.DedupPackage.IsEnabled() &&
```
omaciel, line 511: Unittest covers this branch?

pkg/clients/imagebuilder/client.go
```diff
@@ -546,11 +596,31 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*S
 		}).Error(new(PackageRequestError))
 		return nil, new(PackageRequestError)
 	}
-	var searchResult SearchPackageResult
+	var searchResult models.SearchPackageResult
 	err = json.Unmarshal(respBody, &searchResult)
 	if err != nil {
```
omaciel, line 601: Unittest covers this branch?

omaciel (changes requested):
We can talk about it and go over together if you think it might make it easier?

pkg/clients/imagebuilder/client.go
ldjebran: we need this map to check later if the name is already saved, because  the map returned with ValidatePackages have ID as keys, so we need to build a map of just the existing packages.
but Yes @omaciel ValidatePackages can return a map of packagename as keys and we will not need packagesExistsMap at all (and also dupPackages).

pkg/clients/imagebuilder/client.go
acosferreira: yes we have 5 diff scenarios created covering those pices

pkg/clients/imagebuilder/client.go
acosferreira: yes we have 5 diff scenarios created covering those pices

pkg/clients/imagebuilder/client.go
acosferreira: yes we have 5 diff scenarios created covering those pices

pkg/clients/imagebuilder/client_test.go
acosferreira: added the after each to make sure we are in same context

adeliaferreira (approved):
/lgtm

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,64 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
-		pkg := models.InstalledPackage{
-			Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
-			Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
-			Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
-			Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+		metadataPackages = append(metadataPackages,
+			fmt.Sprintf("%s-%s-%s", metadata.InstalledPackages[n].Name, metadata.InstalledPackages[n].Release, metadata.InstalledPackages[n].Version))
+	}
+
+	var packagesExistsMap map[string]*models.InstalledPackage
```
loadtheaccumulator, line 468: Should these lines be inside the feature flag?

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,64 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+					Arch: metadata.InstalledPackages[n].Arch, Name: metadata.InstalledPackages[n].Name,
+					Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
+					Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
+					Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+				}
+				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
+			}
+
+		} else {
+			fmt.Printf("\n passei aqui: %v\n", n)
```
loadtheaccumulator, line 492: Remove debug print statement.

pkg/clients/imagebuilder/client.go
```diff
@@ -463,16 +458,64 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
...
+				Release: metadata.InstalledPackages[n].Release, Sigmd5: metadata.InstalledPackages[n].Sigmd5,
+				Signature: metadata.InstalledPackages[n].Signature, Type: metadata.InstalledPackages[n].Type,
+				Version: metadata.InstalledPackages[n].Version, Epoch: metadata.InstalledPackages[n].Epoch,
+			}
+			image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
+		}
+	}
+
 	image.Commit.OSTreeCommit = metadata.OstreeCommit
+	db.DB.Omit("Image.InstalledPackages.*").Save(image.Commit)
```
loadtheaccumulator, line 504: Is there functionality that needs to exist outside of the feature flag? If yes, should there be two separate PRs?

loadtheaccumulator (changes requested):
Just some questions and minor changes

pkg/clients/imagebuilder/client.go
acosferreira: no, this needs to happen regardless of feature flag.

omaciel (approved):
/lgtm

Looks like my concerns/suggestions have been addressed.

pkg/clients/imagebuilder/client.go
acosferreira: we need to declare to be able to use it

de1987 (approved):

pkg/clients/imagebuilder/client.go
loadtheaccumulator: I'm still iffy on introducing a feature in one PR and its flagged replacement in the same PR, but not a show stopper for me.

loadtheaccumulator (approved):
I'm still iffy on introducing a feature in a PR and its flagged replacement in the same PR, but not a show stopper for me.

PR #1887, opened 2023-02-14
Add Glitchtip support

tpapaioa:
# Description
This PR implements GlitchTip for error tracking:
- add new `GlitchtipDsn` configuration parameter, which is read as a Vault secret when deployed via Clowder
- add new dependency `github.com/getsentry/sentry-go`
- initialize the Sentry client in `main.go`, so that all exceptions caught by the client in `main()` will be sent to GlitchTip

Normal functionality is unaffected, and I've verified that a `panic()` added after https://github.com/RedHatInsights/edge-api/blob/cc3990138b72f90ca592cf893b8b72770be3ba4c/main.go#L179 is successfully reported to the edge-api-stage GlitchTip project for PR checks run against stage.

FIXES: THEEDGE-2974

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,10 +25,12 @@ import (
 	"github.com/redhatinsights/edge-api/pkg/routes"
 	"github.com/redhatinsights/edge-api/pkg/services"
 	edgeunleash "github.com/redhatinsights/edge-api/unleash"
+	feature "github.com/redhatinsights/edge-api/unleash/features"
 
 	"github.com/redhatinsights/edge-api/config"
 
 	"github.com/Unleash/unleash-client-go/v3"
+	"github.com/getsentry/sentry-go"
 	"github.com/go-chi/chi"
 	"github.com/go-chi/chi/middleware"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -167,6 +168,16 @@ func main() {
 	initDependencies()
 
 	cfg := config.Get()
+
+	// Set up Sentry client for GlitchTip error tracking
+	sentry.Init(sentry.ClientOptions{
```
omaciel, line 173: Perhaps we could create a FeatureFlag for this so that it's easier to test, enable/disable? Otherwise, looks good

de1987: Agree

tpapaioa: @omaciel @de1987 I've added a new feature flag (currently toggled off) in Unleash and added a check before enabling the client.

unleash/features/feature.go
```diff
@@ -23,6 +23,11 @@ type Flag struct {
 	EnvVar string
 }
 
+// GLITCHTIP LOGGING FLAGS
+
+// GlitchtipLogging is the feature flag for reporting errors to GlitchTip
+var GlitchtipLogging = &Flag{Name: "edge-management.glitchtip_logging", EnvVar: "GLITCHTIP_LOGGING"}
```
omaciel, line 29: Where is GLITCHTIP_LOGGING defined?

tpapaioa: Environment parameter overrides for feature flags are checked here: 

https://github.com/RedHatInsights/edge-api/blob/1604ec8ca789f1dc52bd2b83bfeddd75b35800a9/unleash/features/feature.go#L101

de1987 (approved):

omaciel (approved):
/lgtm

PR #1892, opened 2023-02-15
THEEDGE-3037 Fix wrong HTTP status when trying to update an image that is not the latest

de1987:
# Description
The edge-api was returning the wrong HTTP status when trying to update an image that is not the latest.
These changes fix the issue.

FIXES: [THEEDGE-3037](https://issues.redhat.com/browse/THEEDGE-3037)

## Type of change

What is it?

- [ ] Documentation update
- [x] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/images_test.go
+++ b/pkg/routes/images_test.go
@@ -25,7 +25,10 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/redhatinsights/edge-api/pkg/clients/imagebuilder/mock_imagebuilder"
+	"github.com/redhatinsights/edge-api/pkg/common/seeder"
 	"github.com/redhatinsights/edge-api/pkg/dependencies"
+	EdgeErrors "github.com/redhatinsights/edge-api/pkg/errors"
 	"github.com/redhatinsights/edge-api/pkg/models"
 	"github.com/redhatinsights/edge-api/pkg/services/mock_services"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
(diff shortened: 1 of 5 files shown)


pkg/routes/images_test.go
```diff
@@ -1732,3 +1735,147 @@ var _ = Describe("Images Route Tests", func() {
...
+		router.Route("/images", MakeImagesRouter)
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Context("POST /images/{imageId}/update", func() {
+		When("using the latest image with valid information", func() {
+			It("should create image update successfully", func() {
```
mgold1234, line 1781: do you mean that the image update successfully?
I think maybe the 'create' is uncesussery here

pkg/services/images.go
```diff
@@ -683,7 +684,7 @@ func (s *ImageService) processImage(ctx context.Context, id uint) {
 		s.log.WithField("error", err.Error()).Error("Failed creating commit for image")
 	}
 
-	if image.Commit.Status == models.ImageStatusSuccess {
+	if image.Commit != nil && image.Commit.Status == models.ImageStatusSuccess {
```
omaciel, line 687: I feel that we should add tests for this dual conditional statement. It doesn't look like we testing `processImage` directly (yes, I know it's a 'private' function :P

pkg/services/images.go
```diff
@@ -971,7 +972,7 @@ func (s *ImageService) cleanFiles(kickstart string, isoName string, imageID uint
 
 // UpdateImageStatus updates the status of an commit and/or installer based on Image Builder's status
 func (s *ImageService) UpdateImageStatus(image *models.Image) (*models.Image, error) {
-	if image.Commit.Status == models.ImageStatusBuilding {
+	if image.Commit != nil && image.Commit.Status == models.ImageStatusBuilding {
```
omaciel, line 975: I feel that we should add tests for this dual conditional statement. It doesn't look like we testing `UpdateImageStatus` directly

omaciel (changes requested):
Looking good and it'll look even better with more tests :)

pkg/routes/images_test.go
de1987: it's not updating an image, it's in the step that creates the update itself
any suggestion to improve the description?

pkg/services/images.go
de1987: I agree with you, after carefully analyzing the tests I fixed the mocks and removed those changes

pkg/routes/images_test.go
```diff
@@ -1732,3 +1735,144 @@ var _ = Describe("Images Route Tests", func() {
...
+				composeImage.ImageSetID = &imageSet.ID
+
+				var body bytes.Buffer
+				err := json.NewEncoder(&body).Encode(&updateImage)
+				Expect(err).ToNot(HaveOccurred())
+
+				req, err := http.NewRequest("POST", fmt.Sprintf("/images/%d/update", image.ID), &body)
+				Expect(err).ToNot(HaveOccurred())
+
+				mockImageBuilder.EXPECT().ComposeCommit(gomock.Eq(&composeImage)).DoAndReturn(func(image *models.Image) (*models.Image, error) {
```
ldjebran, line 1815:
```suggestion
				mockImageBuilder.EXPECT().ComposeCommit(gomock.Eq(&composeImage)).Return(image, nil)
```
we do not need the function here

ldjebran (commented):
lgtm, left only one comment

pkg/routes/images_test.go
de1987: hey @ldjebran thank you for your suggestion
for this one, it's required because by using `Return` only the main code loses the pointer reference and causes panic
with `DoAndReturn` is able to get the pointer and just return it

I double-check with your suggestion and it's causing panic

omaciel (approved):
/lgtm

pkg/routes/images_test.go
ldjebran: and with the pointer 

```go
mockImageBuilder.EXPECT().ComposeCommit(gomock.Eq(&composeImage)).Return(&image, nil)
```

ldjebran: after an offline  discussion and debugging session we came on that we need this function, and we need to return the image received by ComposeCommit that we cannot control from our test.
resolving question

ldjebran (approved):
lgtm, Thanks @de1987

PR #1908, opened 2023-02-20
Improve coverage to search package method

adeliaferreira:
# Description
Add a new test case to cover error on web service on search package function

FIXES: THEEDGE-3058

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor


pkg/clients/imagebuilder/client_test.go
```diff
@@ -85,6 +85,18 @@ var _ = Describe("Image Builder Client Test", func() {
 		Expect(err).To(BeNil())
 		Expect(res.Meta.Count).To(Equal(0))
 	})
+	It("test web service error while search a package", func() {
```
mgold1234, line 88: maybe better to write 'while there is no package to search" WDYT?

adeliaferreira: This new test case will test the http fail status ( Bad request) from web service. Maybe "test web service failure on search package" could be a better string?

AaronH88: "StatusBadRequest" seems to be the most important part of this test, and in my opinion, the test name should reflect this.
Consider something like "test web service when package search returns StatusBadRequest"

adeliaferreira: Sounds good to me the @AaronH88 suggestion, are you good with this nome @mgold1234 ?

adeliaferreira: It's already update, if you can recheck, it'll be amazing. Thanks everyone

AaronH88 (approved):
/lgtm

pkg/clients/imagebuilder/client_test.go
```diff
@@ -85,6 +85,18 @@ var _ = Describe("Image Builder Client Test", func() {
...
 		Expect(res.Meta.Count).To(Equal(0))
 	})
+	It("test web service when package search returns StatusBadRequest", func() {
+		ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusBadRequest)
+			fmt.Fprintln(w, `{"error":{"count":0}}`)
+		}))
+		defer ts.Close()
+		config.Get().ImageBuilderConfig.URL = ts.URL
```
ldjebran, line 95: we need to restore the initial config imagebuilder url

adeliaferreira: sorry, don't understood, can you explain better?

adeliaferreira: Added, thanks for your suggestion

AaronH88 (approved):
/lgtm

ldjebran (approved):
lgtm

PR #1912, opened 2023-02-22
images: Fixes and add services/images.go coverage.

ldjebran:
# Description
- Replace some function to use global service FilesService and ProducerServices.
- Replace exec.Cmd by the more testable BuildCommand.
- Fixes some minor bugs in the code.
- Add more coverage to services/images.go

FIXES: THEEDGE-3062

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -12,8 +12,8 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"os/exec"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -81,6 +81,7 @@ func NewImageService(ctx context.Context, log *log.Entry) ImageServiceInterface 
 		RepoService:     NewRepoService(ctx, log),
 		ProducerService: kafkacommon.NewProducerService(),
 		TopicService:    kafkacommon.NewTopicService(),
+		FilesService:    NewFilesService(log),
 	}
 }
 
@@ -93,6 +94,7 @@ type ImageService struct {
 	RepoService     RepoServiceInterface
 	ProducerService kafkacommon.ProducerServiceInterface
 	TopicService    kafkacommon.TopicServiceInterface
+	FilesService    FilesService
 }
 
 // GetImageReposFromDB return ThirdParty repo of image by OrgID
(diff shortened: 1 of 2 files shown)


mgold1234 (approved):
/lgtm

pkg/services/images.go
```diff
@@ -869,7 +870,7 @@ func (s *ImageService) addSSHKeyToKickstart(sshKey string, username string, kick
 	td := UnameSSH{sshKey, username}
 
 	s.log.WithField("templatesPath", cfg.TemplatesPath).Debug("Opening file")
-	t, err := template.ParseFiles(cfg.TemplatesPath + "templateKickstart.ks")
+	t, err := template.ParseFiles(path.Join(cfg.TemplatesPath, "templateKickstart.ks"))
```
omaciel, line 873: Love it! :)

pkg/services/images_test.go
```diff
@@ -2121,4 +2141,493 @@ var _ = Describe("Image Service Test", func() {
...
+				// disable feature edge-management.completion_events
+				os.Unsetenv("FEATURE_COMPLETION_EVENTS")
+			})
+
+			It("should produce event models.EventTypeEdgeOstreeRepoCompleted", func() {
+				expectedRepo := models.Repo{URL: faker.URL(), Status: models.RepoStatusSuccess}
+				err := db.DB.Create(&expectedRepo).Error
+				Expect(err).ToNot(HaveOccurred())
+				mockRepoBuilder.EXPECT().ImportRepo(gomock.AssignableToTypeOf(&models.Repo{})).Return(&expectedRepo, nil)
+				mockProducerService.EXPECT().ProduceEvent(
```
omaciel, line 2212: Where do we assert that the event returned is the expected one? I'm sure it's my incomplete understanding of the code...

ldjebran: here we are testing that ProduceEvent is called by the proper args , the two first args we know that values must be as we set it here, for the third arg , we can not get the real value but we know that it must be of type models.CRCCloudEvent. For the returned value we only assume that it should return nil , and the mock will return nil (as we are not testing Produce event, Produce event is tested in it's own module ).

@omaciel is this is what you were looking for?

ldjebran: also if we will assume that error is returned from ProduceEvent,  the assertion on error should be the same as the returned from ProduceEvent, this is on the next test https://github.com/RedHatInsights/edge-api/pull/1912#discussion_r1114648932

omaciel: Thanks @ldjebran for the explanation.

PR #1932, opened 2023-03-01
remove validation of env to set gpg verify; Define always as false;

acosferreira:
# Description
rollback validation on gpg verify and set default as true

FIXES: THEEDGE-3051 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -180,11 +180,10 @@ func CreateEdgeAPIConfig() (*EdgeConfig, error) {
 	}
 	options.SetDefault("FeatureFlagsService", os.Getenv("FEATURE_FLAGS_SERVICE"))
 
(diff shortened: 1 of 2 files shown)


omaciel (approved):
/lgtm

Do we need the template to be resigned?

de1987 (approved):

de1987:
> /lgtm
> 
> Do we need the template to be resigned?

should edge management have your own signature?

acosferreira:
@omaciel no, the playbook now is parameterized, and now the parameter will be always false instead based on environment 
The playbook should be re-signed only when we change anything inside of them.

PR #1944, opened 2023-03-08
Fix commits save

acosferreira:
# Description
Fix sql statment for psql

FIXES: THEEDGE-2648

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/clients/imagebuilder/client.go
+++ b/pkg/clients/imagebuilder/client.go
@@ -486,7 +486,8 @@ func (c *Client) GetMetadata(image *models.Image) (*models.Image, error) {
 				}
 				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, pkg)
 			} else {
-				image.Commit.InstalledPackages = append(image.Commit.InstalledPackages, *packagesExistsMap[metadata.InstalledPackages[n].Name])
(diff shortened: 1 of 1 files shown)


pkg/clients/imagebuilder/client.go
```diff
@@ -588,8 +589,10 @@ func (c *Client) SearchPackage(packageName string, arch string, dist string) (*m
 func (c *Client) ValidatePackages(pkgs []string) (map[string]*models.InstalledPackage, error) {
 	var result []models.InstalledPackage
 	setOfPackages := make(map[string]*models.InstalledPackage)
-	if err := db.DB.Table("Installed_Packages").
-		Where("( (name || '-' || release || '-' ||  version)) in (?)", pkgs).
+	if err := db.DB.Model(&models.InstalledPackage{}).
```
mgold1234, line 592: can you add unit test for that?

acosferreira: we already have the unit tests, however the test was passing due sqlite instances handle the previous sql statement with no error.

loadtheaccumulator (approved):

PR #1951, opened 2023-03-13
content-sources: API return HMS repositories

ldjebran:
# Description

- To preserve compatibility: Return HMS repositories as thirdparty repos models with arbitrary/virtual ids, except for the ones already selected we need to supply their real ids , this happens when we update an image, and we supply an imageID in url param. The resolution is to have virtual ids superior to any id known in the current context. 
- Introduced a function GetImageByIDExtended that allow to have the image only with the data that are needed, and to not overcomplicate/overload this function for simple usage, when gormDB is not supplied (is nil) the core image is returned without any data queried.
- All the code is 100 % covered.

FIXES: THEEDGE-3158

## Type of change

What is it?

- [ ] Documentation update
- [X] Tests update
- [ ] Refactor

# Checklist:


pkg/routes/thirdpartyrepo.go
```diff
@@ -144,8 +148,144 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+	return response, err
+}
+
+func getImageFromURLParam(w http.ResponseWriter, r *http.Request, gormDB *gorm.DB) (*models.Image, error) {
+	ctxServices := dependencies.ServicesFromContext(r.Context())
+	imageIDString := r.URL.Query().Get("imageID")
+	if imageIDString == "" {
+		// when imageID not Defined return a nil image
+		return nil, nil
+	}
```
de1987, line 199: is this error handled somewhere else?

pkg/routes/thirdpartyrepo_test.go
```diff
@@ -392,4 +397,281 @@ var _ = Describe("ThirdPartyRepos basic routes", func() {
...
+
+	Context("#GetAllContentSourcesRepositories", func() {
+		var orgID string
+		var image models.Image
+		var mockRepositoriesService *mock_repositories.MockClientInterface
+		var mockImageService *mock_services.MockImageServiceInterface
+		var ctrl *gomock.Controller
+		var listRepositoryResponse repositories.ListRepositoriesResponse
+		var thirdPartyRepo models.ThirdPartyRepo
+
```
de1987, line 409: could we have to check when the `imageID` is missing?
WDYT?

de1987 (changes requested):

pkg/routes/thirdpartyrepo.go
ldjebran: This is not an error, this mean we have no error , and no image

de1987 (approved):

pkg/routes/thirdpartyrepo_test.go
ldjebran: we have a test without image , listing with filtering by name

pkg/routes/thirdpartyrepo.go
```diff
@@ -144,8 +148,144 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+				if repo.ID > maxImageRepoIDValue {
+					maxImageRepoIDValue = repo.ID
+				}
+			}
+		}
+	}
+
+	repos := make([]models.ThirdPartyRepo, 0, len(response.Data))
+	for ind, ContentSourcesRepo := range response.Data {
+		// calculate the id to set , to not have conflict with the saved one on db,
```
mgold1234, line 259:
```suggestion
		// calculate the id to set , that will not have a conflict with the saved one on db,
```

pkg/services/images.go
```diff
@@ -1143,6 +1147,32 @@ func (s *ImageService) addImageExtraData(image *models.Image) (*models.Image, er
 	return image, nil
 }
 
+// GetImageByIDExtended retrieves an image by its identifier and use the supplied gormDB instance
```
mgold1234, line 1150:
```suggestion
// GetImageByIDExtended retrieves an image by its identifier and uses the supplied gormDB instance
```

pkg/services/images.go
```diff
@@ -1143,6 +1147,32 @@ func (s *ImageService) addImageExtraData(image *models.Image) (*models.Image, er
...
+// This is more customisable function version, that will get only the needed data, if gormDB is nil, we get an image without
```
mgold1234, line 1151:
```suggestion
// This is more customizable function version, that will get only the needed data, if gormDB is nil, we get an image without
```

pkg/routes/thirdpartyrepo.go
```diff
@@ -144,8 +148,144 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+		// use an ID superior of any known one in the current context, add overall repos count so that we will not conflict with
+		// ids used on different pages even if we change pagination.Limit when changing pages
+		virtualIDValue := maxImageRepoIDValue + uint(response.Meta.Count+ind)
+		edgeRepo := models.ThirdPartyRepo{
```
acosferreira, line 263: Should we include also the new fields?

ldjebran: for the UI  it's not needed , but for the API generally, this may be needed.
Will add them.

pkg/routes/thirdpartyrepo.go
ldjebran: ok , will rephrase it .

pkg/routes/thirdpartyrepo.go
```diff
@@ -144,8 +148,144 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+		return
+	}
+
+	image, err := getImageFromURLParam(w, r, db.DB.Preload("ThirdPartyRepositories"))
+	if err != nil {
+		// getImageFromURLParam already logged and responded to request
+		return
+	}
+
+	// create a map of the image (if defined) for repos uuid to repos id and define the maximum id value
```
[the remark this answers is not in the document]
ldjebran: here we mean if image was defined (in the past)  and not to define in the present or future.

pkg/services/images.go
ldjebran: done

pkg/services/images.go
ldjebran: done

pkg/routes/thirdpartyrepo.go
ldjebran: done

pkg/routes/thirdpartyrepo.go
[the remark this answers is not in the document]
ldjebran: done

pkg/routes/thirdpartyrepo.go
```diff
@@ -144,8 +148,144 @@ func createRequest(w http.ResponseWriter, r *http.Request) (*models.ThirdPartyRe
...
+	}
+
+	return response, err
+}
+
+func getImageFromURLParam(w http.ResponseWriter, r *http.Request, gormDB *gorm.DB) (*models.Image, error) {
+	ctxServices := dependencies.ServicesFromContext(r.Context())
+	imageIDString := r.URL.Query().Get("imageID")
+	if imageIDString == "" {
+		// when imageID not Defined return a nil image
```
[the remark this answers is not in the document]
ldjebran: done

pkg/routes/thirdpartyrepo.go
ldjebran: done

ldjebran:
@acosferreira @mgold1234 thanks you for your review, all comments has been applied, please revisit.
@de1987 thanks for your review, your approval has been dismissed because of new commits, please revisit.

adeliaferreira (approved):
LGTM

PR #1960, opened 2023-03-16
Add description to ConfigureHttpClient function

mgold1234:
# Description

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,5 +58,3 @@ go/pkg/mod/*
 /build/
 /coverage.json
 /coverage.html
-
-rebase.sh
--- a/pkg/clients/client.go
+++ b/pkg/clients/client.go
@@ -34,7 +34,9 @@ func GetOutgoingHeaders(ctx context.Context) map[string]string {
 	return headers
 }
 
-func ConfigureHttpClient(client *http.Client) *http.Client {
+// ConfigureClientWithTlsPath return https client in case there is TLS CA path,
+// otherwise return the original unmodified client and logs an error
+func ConfigureClientWithTlsPath(client *http.Client) *http.Client {
 	cfg := config.Get()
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -58,3 +58,5 @@ go/pkg/mod/*
 /build/
 /coverage.json
 /coverage.html
+
+rebase.sh
```
loadtheaccumulator, line 62: Please remove this rebase.sh script.

rebase.sh
```diff
@@ -0,0 +1,6 @@
+
```
loadtheaccumulator, line 1: Please remove this file.

pkg/clients/client.go
```diff
@@ -34,6 +34,8 @@ func GetOutgoingHeaders(ctx context.Context) map[string]string {
 	return headers
 }
 
+// ConfigureHttpClient return https client in case there is TLS CA path,
```
loadtheaccumulator, line 37: I would reword to say it adds a TLS config to an http client.
This function is called on the assumption there is a TLS CA Path
Line below looks awesome, but you could mention that if there is no TLS CA, it logs an error and returns the original unmodified client.

loadtheaccumulator (changes requested):

rebase.sh
ldjebran, line 1: _New github feature , to comment on the whole file :smiley:_ 

Please remove this file.

loadtheaccumulator:
It looks like this grew in scope from just adding a description. Please separate them so that a problem with one does not interfere with the other.

pkg/clients/client.go
mgold1234: @loadtheaccumulator added the changes

loadtheaccumulator (approved):
lgtm

ldjebran (approved):

PR #1976, opened 2023-03-27
add a fix to upgrade playbook in case of customer change the configuration

mgold1234:
the configuration file.
related to https://issues.redhat.com/browse/THEEDGE-3223?focusedId=21963643&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21963643

# Description

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/templates/template_playbook_dispatcher_ostree_rebase_payload_prod.test.yml
+++ b/templates/template_playbook_dispatcher_ostree_rebase_payload_prod.test.yml
@@ -19,29 +19,35 @@
       contenturl={{ repo_content_url }}
       tls-client-key-path=/etc/pki/consumer/key.pem
       tls-client-cert-path=/etc/pki/consumer/cert.pem
+      {{ proxy_details.stdout }}
     insights_signature_exclude: "/vars/insights_signature,/vars/update_number,/vars/ostree_remote_name,/vars/ostree_changes_refs,/vars/os_tree_ref,/vars/repo_url,/vars/repo_content_url,/vars/ostree_gpg_verify"
     insights_signature: !!binary |
       TFMwdExTMUNSVWRKVGlCUVIxQWdVMGxIVGtGVVZWSkZMUzB0TFMwS1ZtVnljMmx2YmpvZ1IyNTFV
-      RWNnZGpFS0NtbFJTVlpCZDFWQldTOVZaSFJqZG5jMU9FUXJhalZ3VGtGUmFuUkdVUzhyUzJwMVVu
-      WXdaRVJuWmpBemFHVkVWUzg1ZGpsRmFtaGhTVWxJT0VwSU1rUUtRMXBMVTFsRldHeFVkRlpzVEdK
-      RFprOXlORFZqWWs1dU5tVmlaVVVyTUdSMllVbDJUelZIUjNCUGVtTndWak56UkM5MGJEaGxWVTQy
-      T0RsNVJHOUpjQXA0VW5ScFVtUlVZVlJqVDFFeE5XYzBTRGxxTkdvek5uTkhWbVpSUWxsdk9XSm1P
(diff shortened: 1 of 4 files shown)


loadtheaccumulator:
What happens if a line other than proxy is added?

de1987:
> What happens if a line other than proxy is added?

agree with Jonathan
and if it's empty? or null?

Also, what is the security it could imply?

acosferreira (changes requested):
validates if has all the required keys present 
and should just change the contenturl value to the updated value.
this will prevent to change the file in case the user add any other confi besides the proxy

loadtheaccumulator:
> ... should just change the contenturl value to the updated value. this will prevent to change the file in case the user add any other confi besides the proxy

+1 

@mgold1234 That is what I was recommending in our chat. Any values passed via the playbook should be added/modified instead of overwriting the entire file with customizations read from local and added back.

templates/template_playbook_dispatcher_ostree_rebase_payload_prod.test.yml
```diff
@@ -11,41 +11,42 @@
     os_tree_ref: "rhel/9/x86_64/edge"
     ostree_gpg_verify: "false"
     ostree_gpg_keypath: "/etc/pki/rpm-gpg/"
-    ostree_remote_template: |
-      [remote "{{ ostree_remote_name }}"]
```
ldjebran, line 15: I think we are missing the config section

ldjebran (commented):
awesome resolution, but we should be also able  to create the file if it's missing 
and of course with the config section.
```
[remote “rhel-edge”]


```

templates/template_playbook_dispatcher_ostree_rebase_payload_prod.test.yml
ldjebran: is "ostree_remote_name" var is a constant or may change over updates ?

ldjebran: I think we should
 what  do you think @loadtheaccumulator @acosferreira

loadtheaccumulator: @mgold1234 @ldjebran I think if file exists would be a good check. You would do it programmatically out of habit, so probably a good practice to continue here.

loadtheaccumulator: Take a look at backup and create options.

templates/template_playbook_dispatcher_ostree_upgrade_payload.yml
```diff
@@ -11,41 +11,42 @@
...
+      dldVWmFVa2RsTURSek0zRkJUMnB0YzNWeldYWndTVWt4VHk5WGRXZFhkbVlyUldndlIxSTVjRkZv
+      VlhCTVdFNUhTVmRMWlVGamJEY0tZVXhOTTNneU1YTjNTWFpLYTFwcGNGTnlabFJaYUdkRVFVUTVW
+      MVV6ZEM5UVpFVTJOM0EyWTNnNVFrRnFUWEF4YUZsV1VETlNUMk5NUXpORUswdFlNd3B6WW14T05X
+      eGxXSEJQYXowS1BYRkJWVlVLTFMwdExTMUZUa1FnVUVkUUlGTkpSMDVCVkZWU1JTMHRMUzB0Q2c9
       PQ==
   tasks:
     - name: apply templated ostree remote config
-      ansible.builtin.copy:
-        content: "{{ ostree_remote_template }}"
+      ansible.builtin.lineinfile:
```
loadtheaccumulator, line 38: Yeah. Please backup the file before modifying, and take a look at the create option to create the file if it doesn't exist.

loadtheaccumulator (changes requested):

loadtheaccumulator:
https://github.com/RedHatInsights/edge-api/pull/1994 replaces this PR.

PR #1986, opened 2023-04-03
content-sources: add client CreateRepository.

ldjebran:
# Description
The CreateRepository the repositories function that will be used to create repository in content-sources.

100% coverage

FIXES: THEEDGE-3208

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/clients/repositories/clients_test.go
+++ b/pkg/clients/repositories/clients_test.go
@@ -545,3 +545,166 @@ func TestGetRepositoryByUUID(t *testing.T) {
 		})
 	}
 }
+
+// ErrFaultyWriter the error returned by FaultWriter's Write function
+var ErrFaultyWriter = errors.New("faulty writer expected write error")
+
+// FaultyWriter a struct that implement a Writer that return always error when writing
+type FaultyWriter struct{}
+
+func (f *FaultyWriter) Write(_ []byte) (n int, err error) {
+	return 0, ErrFaultyWriter
+}
+
+func TestCreateRepository(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/clients/repositories/clients_test.go
```diff
@@ -545,3 +545,166 @@ func TestGetRepositoryByUUID(t *testing.T) {
...
+			Name:           "should create repository successfully",
+			RepoToCreate:   repoToCreate,
+			HTTPStatus:     http.StatusCreated,
+			IOReadAll:      io.ReadAll,
+			NewJSONEncoder: json.NewEncoder,
+			Response:       &responseRepo,
+			ExpectedError:  nil,
+		},
+		{
+			Name:           "should return error when http status is not 201",
```
mgold1234, line 593: do you know what is the status error you get? 
maybe specify it in the Name test?

mgold1234: at this specific test you check StatusBadRequest                   = 400, so we can define it in the test name as well

ldjebran: The most important in this test is that the returned status is not 201, I can put 500 as returned status and the result will be the same, so the most important here is not what we will receive as returned status, but how our code will react when the status is different from 201.

ldjebran: Also you can read in the code that we are checking if the returned http status is different from 201, and not to equal to 400.

mgold1234 (approved):
/lgtm

PR #2011, opened 2023-04-20
content-sources: update migration preparation script to comply with content-sources rules.

ldjebran:
# Description
Update the migration preparation script:
- with default data collection limit = 100
- with default pages to process = 100, to prevent any forever loop, even if the logic does not allow that.

FIXES: THEEDGE-3262

Put this as "do not merge" to enable feature migration flag  before merge. 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/cmd/migraterepos/repairrepos/repairrepos.go
+++ b/cmd/migraterepos/repairrepos/repairrepos.go
@@ -10,7 +10,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// ErrMigrationNotAvailable error returned when the migration feature flag is disabled
 var ErrMigrationNotAvailable = errors.New("migration is not available")
+
+// DefaultDataLimit the default data limit to use when collecting data
+var DefaultDataLimit = 100
+
+// DefaultMaxDataPageNumber the default data pages to handle as preventive way to enter an indefinite loop
+var DefaultMaxDataPageNumber = 100
 
 // RepairUrls add slash to urls to conform to content-sources urls
 // any url like "http://example.com" became "http://example.com/"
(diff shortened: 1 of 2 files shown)


mgold1234 (approved):
/lgtm

cmd/migraterepos/repairrepos/repairrepos.go
```diff
@@ -25,31 +32,40 @@ func RepairUrls() (int, error) {
...
+	for page < DefaultMaxDataPageNumber {
+		var repoURLs []RepoURL
+		if err := db.DB.Debug().Model(&models.ThirdPartyRepo{}).
+			Select("url, count(url) as count").
+			Where("url NOT LIKE '%/' AND url IS NOT NULL AND url != ''").
+			Group("url").
+			Limit(DefaultDataLimit).
+			Scan(&repoURLs).Error; err != nil {
+
+			log.WithField("error", err.Error()).Error("error when retrieving urls without slashes")
```
de1987, line 46: Thanks for working on those fixes @ldjebran 
1- It seems the total number of pages is fixed, I think would be a better approach if we dynamically get it from the database, because we could have less or more data than just 100.

2- Another observation, I would guess the data retrieved from the database will be always the same batch because the query is not using `offset`.

WDYT?

cmd/migraterepos/repairrepos/repairrepos.go
```diff
@@ -115,30 +131,40 @@ func RepairDuplicates() error {
...
+		var reposData []RepoData
+		if err := db.DB.Debug().Model(&models.ThirdPartyRepo{}).
+			Select("org_id, url, count(url) as count").
+			Where("url IS NOT NULL AND url != ''").
+			Group("org_id, url").
+			Having("count(url) > 1").
+			Limit(DefaultDataLimit).
+			Scan(&reposData).Error; err != nil {
+			log.WithField("error", err.Error()).Error("error when collecting orgs duplicate urls")
+		}
```
de1987, line 146: Same here

de1987 (changes requested):
I'm concerned about the two observations I added.
Could you check?

cmd/migraterepos/repairrepos/repairrepos.go
ldjebran: we do not need that as with each iteration the data will change

cmd/migraterepos/repairrepos/repairrepos.go
ldjebran: same as my first comment

cmd/migraterepos/repairrepos/repairrepos.go
ldjebran: the data of the first iteration will disappear as to be repaired

ldjebran: on each iteration we have a new data set to repair, the unit-test has to fail in you case

de1987 (approved):

PR #2086, opened 2023-06-10
add repo file upload retry with configurable number of attempts

loadtheaccumulator:
* configure number of repo file upload attempts
* retry failed file uploads up to configured # of attempts

# Description
Individual repo files can fail to upload.
This PR adds a configurable number of attempts to successfully upload.

FIXES: THEEDGE-2510

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


loadtheaccumulator:
In draft. Tests next.

loadtheaccumulator:
Moved break to fix test.

loadtheaccumulator:
Seeing sporadic test failures for coverage (which we run multiple times in various tests???), so setting back to draft.

loadtheaccumulator:
Resolved sporadic race condition with using "for i" in the go func. Renamed to "attempt". Taking this off draft.

pkg/services/files/uploader.go
```diff
@@ -92,13 +93,28 @@ type uploadDetails struct {
...
-		fname, err := p.uploader.UploadFileWithACL(p.fileName, p.uploadPath, acl)
-		if err != nil {
-			u.log.WithFields(log.Fields{"fname": fname, "count": p.count, "error": err.Error()}).Error("Error uploading file")
+		// attempt to upload a file multiple times before erroring
+		for attempt := 1; attempt < attemptLimit; attempt++ {
+			fname, err := p.uploader.UploadFileWithACL(p.fileName, p.uploadPath, acl)
+			// log on file upload failure and retry
+			if err != nil {
+				u.log.WithFields(log.Fields{"fname": fname, "attempt": attempt, "count": p.count, "error": err.Error()}).Error("Error uploading file")
+				time.Sleep(1 * time.Second)
```
ldjebran, line 105: can we put this sleep retry in a constant ?

pkg/services/files/uploader_test.go
```diff
@@ -138,6 +138,7 @@ var _ = Describe("Uploader Test", func() {
 			var initialRepoTempPath string
 
 			BeforeEach(func() {
+				config.Init()
```
ldjebran, line 141: maybe the correct refactor will be as :

```
			var conf *config.EdgeConfig
                        ...  
			BeforeEach(func() {
				conf = config.Get()
```

what do you think ?

ldjebran (commented):
looks good, left 2 comment , also we need a unittest like "should upload repo"  one  
 name "should upload repo with retry on failure " with the first "s3Client.EXPECT().PutObject" will return err and the second 
"s3Client.EXPECT().PutObject" will succeed.

pkg/services/files/uploader.go
loadtheaccumulator: Yeah. I added the atempts as configurable, and can make that so as well.

pkg/services/files/uploader_test.go
loadtheaccumulator: I can look at that again, but I think I tried it. Oddly enough, the race condition being thrown is between the config.Init() calls in downloader_test.go in and the config.Get() in uploader.go. It considers the Init() a write and the Get a read() and they aren't consistent in the order they run when testing. I did a number of things, but adding the "Write" in uploader_test.go before the call to uploader.go resolved the race. I don't see the problem in the app itself.

ldjebran: think this the testing framework,  some time weird that some vars get initiated on when BeforeEach occur even if the vars are initiated before calling BeforeEach function 

resolving comment then.

ldjebran (dismissed):
/lgtm

acosferreira (approved):
lgtm

PR #2093, opened 2023-06-14
images: FIXES orphan images raised when building images.

ldjebran:
# Description
This orphan images in this context happen when we delete an image-set when an image is building. 
When we delete an image-set we delete all images first, but if one of them is building, when the final status is returned we are saving that image that is passed through all the function (the image has have the deleted_at not nil), and this restore the image automatically and left it orphan as image-set is already deleted. 

The current PR is reloading the image from db at each iteration, the other changes are related to make this PR testable, without race condition returned from the testing framework because of the interrupt detection procedure.

FIXES: THEEDGE-3414

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:


mgold1234 (approved):
/lgtm

PR #2120, opened 2023-07-03
cleanup: Setup openshift CronJob to never restart on failure.

ldjebran:
# Description

- Logically we do not need this job to be restarted on failure. The job failed with "BackoffLimitExceeded", so there is a persistent error. when restarted we loose our pods logs, and we are incapable to understand why this happened. https://access.redhat.com/solutions/6965050
- Increased the limit to 30 instead of 10, as we have ~ 26000 of records to handle and with max page = 1000 we will be able to handle that number.
- Changed the scheduler to run in two hours from now for debugging.

FIXES: https://issues.redhat.com/browse/THEEDGE-3449

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


adeliaferreira (approved):
Lgtm

mgold1234 (approved):
/lgtm

PR #2152, opened 2023-07-25
Implement the api doc for create device-groups

acosferreira:
# Description
In the context of the swaggo lib implement the api doc for thirdpartyrepo endpoints.
FIXES:
THEEDGE-3379
THEEDGE-3386

## Type of change

What is it?

- [X] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ go/pkg/mod/*
 /build/
 /coverage.json
 /coverage.html
+
+#docfiles
+api/
--- a/pkg/models/devicegroups_api.go
+++ b/pkg/models/devicegroups_api.go
@@ -1,0 +1,32 @@
+package models
+
+// CreateDeviceGroupAPI is the /device-group POST endpoint struct for openapi.json auto-gen
+type CreateDeviceGroupAPI struct {
+	Name    string                    `json:"name" example:"my-device-group"` // the device group name
+	Type    string                    `json:"type" example:"static"`          // the device group type
+	Devices []DeviceForDeviceGroupAPI `json:"DevicesAPI,omitempty"`           // Devices of group
+
+} // CreateDeviceGroup
+
+// DeviceImageInfoAPI is a record of group with the current images running on the device
+type DeviceImageInfoAPI struct {
+	Name            string `json:"name" example:"my-image-name"`
+	Version         int    `json:"version" example:"1"`
+	Distribution    string `json:"distribution" example:"RHEL-92"`
+	UpdateAvailable bool   `json:"update Available" example:"true"`
+	CommitID        uint   `json:"commitID" example:"2"`
+} // DeviceImage
+
+// DeviceForDeviceGroupAPI is a device array expected to create groups with devices
+type DeviceForDeviceGroupAPI struct {
+	UUID string `json:"UUID" example:"68485bb8-6427-40ad-8711-93b6a5b4deac"`
+} // Device
+
+// DeviceGroupDevicesAPI is the /device-group return endpoint struct for openapi.json auto-gen
+type DeviceGroupDevicesAPI struct {
+	Account     string `json:"Account"`
+	OrgID       string `json:"org_id"`
+	Name        string `json:"Name"`
+	Type        string `json:"Type"`
+	ValidUpdate bool   `json:"ValidUpdate" `
+} // DeviceGroup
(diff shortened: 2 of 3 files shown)


de1987 (approved):

.gitignore
```diff
@@ -58,3 +58,6 @@ go/pkg/mod/*
 /build/
 /coverage.json
 /coverage.html
+
+#docfiles
+api/
```
mgold1234, line 63: please remove this file from pr

pkg/routes/devicegroups.go
```diff
@@ -304,15 +304,16 @@ func GetAllDeviceGroups(w http.ResponseWriter, r *http.Request) {
...
+// @ID 			 CreateDeviceGroup
+// @Description  Create Device Group for an account.
 // @Tags         Device Groups
 // @Accept       json
 // @Produce      json
-// @Param        body	body	models.Image	true	"request body"
-// @Success      200 {object} models.SuccessPlaceholderResponse
-// @Failure      400 {object} errors.BadRequest
-// @Failure      500 {object} errors.InternalServerError
+// @Param        body	body models.CreateDeviceGroupAPI	true	"request body"
```
mgold1234, line 313: those params are missing:
       parameters:
         - name: sort_by
           in: query
           description: "fields: created_at, name, updated_at. To sort DESC use - before the fields."
           schema:
             type: string
         - name: name
           in: query
           description: "field: filter by name"
           schema:
             type: string
         - name: created_at
           in: query
           description: "field: filter by create date"
           schema:
             type: string
         - name: updated_at
           in: query
           description: "field: filter by update date"
           schema:
             type: string
         - name: limit
           in: query
           description: 'field: return number of groups until limit is reached.'
           schema:
             type: integer
         - name: offset
           in: query
           description: 'field: return number of groups beginning at the offset.'
           schema:
             type: integer

mgold1234 (commented):
looks good

mgold1234:
I think its better to squash all commits to one commit

pkg/routes/devicegroups.go
acosferreira: @mgold1234 this is the endpoint to post
I dont use this parameters for the create, i believe should be added on the list of groups.
What do you think?

.gitignore
acosferreira: I've added once we should not submit the changed files after run the make swaggo. 
This will prevent us to submit wrong

de1987: @mgold1234 As the `/api` directory should not be committed, I think it makes sense to be here.

loadtheaccumulator: I'm thinking the .gitignore will work great. We can override it if we need to update the README file.

pkg/routes/devicegroups.go
loadtheaccumulator: @mgold1234 it looks like the query in your comment might have been copied from the GET section above this POST. While it's possible to use it in a POST, the query would typically be associated with a GET.

https://github.com/RedHatInsights/edge-api/blob/a2fa0f7fe2e078050f5f05dbd2d8eb474f9e5a18/cmd/spec/path.yaml#L1611-L1638

pkg/models/devicegroups_api.go
```diff
@@ -0,0 +1,32 @@
...
+
+// DeviceForDeviceGroupAPI is a device array expected to create groups with devices
+type DeviceForDeviceGroupAPI struct {
+	UUID string `json:"UUID" example:"68485bb8-6427-40ad-8711-93b6a5b4deac"`
+} // Device
+
+// DeviceGroupDevicesAPI is the /device-group return endpoint struct for openapi.json auto-gen
+type DeviceGroupDevicesAPI struct {
+	Account     string `json:"Account"`
+	OrgID       string `json:"org_id"`
```
loadtheaccumulator, line 28: Is Account or OrgID necessary for a consumer of the endpoint?
It's typically provided via header.

loadtheaccumulator (changes requested):
I just have the question about the Account and OrgID.

pkg/models/devicegroups_api.go
acosferreira: the account and orgId is being exposed in the return of post.
This struct i for the 200 response, should we remove that?

loadtheaccumulator: awesome. thanks.

loadtheaccumulator (approved):

loadtheaccumulator (changes requested):
Please squash the commits.

loadtheaccumulator (approved):
lgtm

ldjebran (approved):
/lgtm

PR #2169, opened 2023-08-02
add static delta state db read and write

loadtheaccumulator:
* add static delta state db write
* add static delta state db read
* add associated positive tests

# Description
Add static delta state database read and write

FIXES: THEEDGE-3483

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/models/staticdelta.go
+++ b/pkg/models/staticdelta.go
@@ -1,5 +1,11 @@
 package models
 
+import (
+	"github.com/redhatinsights/edge-api/pkg/db"
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
 // StaticDeltaState models the state of an existing static delta
 type StaticDeltaState struct {
 	Model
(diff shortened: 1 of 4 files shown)


acosferreira (approved):
lgtm

pkg/models/staticdelta.go
```diff
@@ -17,9 +23,50 @@ const (
...
+		"org_id": sds.OrgID,
+		"name":   sds.Name,
+		"url":    sds.URL,
+		"status": sds.Status}).Debug("Found a static delta")
+
+	return sds, nil
+}
+
+// SaveToStore writes a static delta state to the database
+func (sds *StaticDeltaState) SaveToStore(edgelog *log.Entry) error {
```
ldjebran, line 62: maybe just Save ?

pkg/models/staticdelta.go
```diff
@@ -17,9 +23,50 @@ const (
...
+	edgelog.Info("Updating static delta state record")
+	result := db.DB.Save(&sds)
+	if result.Error != nil {
+		edgelog.Error(result.Error)
+
+		return result.Error
+	}
```
ldjebran, line 69: maybe like this ?

```suggestion
	if err := db.DB.Save(&sds).Error; err != nil {
		edgelog.WithField("error", err.Error()).Error("error occurred while saving static delta")
		return result.Error
	}
```

pkg/models/staticdelta.go
loadtheaccumulator: I started with save but was looking to make it clear where the data is going since we have local, DB, and S3 involved with deltas, but I'm good with whichever for state.

pkg/models/staticdelta.go
ldjebran: it's not a blocker , just a suggestion to make the code more compact :smiley:

pkg/models/staticdelta.go
ldjebran: I think that in the context of the model it's just save (we can save the model only to db store), this can be used for other models too.

ldjebran: but it's not a blocker too :smiley:

ldjebran: already approved the PR, feel free to resolve, if you do not want to change

pkg/models/staticdelta.go
ldjebran: already approved the PR, feel free to resolve, if you do not want to change

loadtheaccumulator: Thanks for the feedback, @ldjebran !!!!

pkg/models/staticdelta.go
loadtheaccumulator: @ldjebran I’m not married to SaveToStore, so I can update it with the next PR. :-)
I should change ReadFromStore as well. Thoughts? I went back and forth on Get, Read, State, yada yada.

ldjebran: simple words make it clean (as far as we remain in the context of a model, there is no ambiguity)

Read Save Refresh ... and more GetState SetState

PR #2209, opened 2023-08-30
migration-groups: create inventory groups client.

ldjebran:
# Description
 - create inventory groups client
 - add uuid to devices groups models that will correspond to inventory group id, that way we will be aware about the already migrated groups, and choose only those without uuid for migration, also that will create a link between local groups and inventory groups for later hosts adding (as hosts are added by group uuid).

 FIXES: https://issues.redhat.com/browse/THEEDGE-3553

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/clients/inventorygroups/client.go
+++ b/pkg/clients/inventorygroups/client.go
@@ -1,0 +1,356 @@
+package inventorygroups
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	url2 "net/url"
+	"strconv"
+
+	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/edge-api/pkg/clients"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var ErrParsingURL = errors.New("error occurred while parsing raw url")
+
+var ErrGroupsRequestResponse = errors.New("inventory groups request error response")
+
+var ErrGroupNotFound = errors.New("inventory group not found")
+
+var ErrGroupNameIsMandatory = errors.New("inventory group name is mandatory")
+
+var ErrGroupUUIDIsMandatory = errors.New("inventory group uuid is mandatory")
+
+var ErrGroupHostsAreMandatory = errors.New("inventory group hosts are mandatory")
+
+// BasePath the inventory groups base path
+const BasePath = "api/inventory/v1/groups"
+
+// DefaultPerPage the default per page limit when listing inventory groups
+const DefaultPerPage = 50
+
+// IOReadAll The io body reader
+var IOReadAll = io.ReadAll
+
+// NewJSONEncoder  create a new json encoder
+var NewJSONEncoder = json.NewEncoder
+
+// Group the type representation returned from inventory groups endpoint
+type Group struct {
+	ID        string `json:"id"`
+	Name      string `json:"name"`
+	OrgID     string `json:"org_id"`
+	HostCount int    `json:"host_count"`
+}
+
+// Response the result type returned when getting a list of inventory groups
+type Response struct {
+	Total   int     `json:"total"`
+	Count   int     `json:"count"`
+	Page    int     `json:"page"`
+	PerPage int     `json:"per_page"`
+	Results []Group `json:"results"`
+}
+
+// ListGroupsParams the parameters used for list request
+type ListGroupsParams struct {
+	Name     string
+	Page     int
+	PerPage  int
+	OrderBy  string
+	OrderHow string
+}
+
+type CreateGroup struct {
+	Name    string   `json:"name"`
+	HostIDS []string `json:"host_ids,omitempty"`
+}
+
+// ClientInterface is an Interface to make request to inventory Groups
+type ClientInterface interface {
+	GetBaseURL() (*url2.URL, error)
+	GetGroupByName(name string) (*Group, error)
+	GetGroupByUUID(groupUUID string) (*Group, error)
+	CreateGroup(groupName string) (*Group, error)
+	AddHostsToGroup(groupUUID string, hosts []string) (*Group, error)
+	ListGroups(requestParams ListGroupsParams) ([]Group, error)
+}
+
+// Client is the implementation of an ClientInterface
+type Client struct {
+	ctx context.Context
+	log *log.Entry
+}
+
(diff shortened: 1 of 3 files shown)


pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if requestParams.Page == 0 {
+		requestParams.Page = 1
+	}
+	if requestParams.PerPage == 0 {
+		requestParams.PerPage = DefaultPerPage
+	}
```
de1987, line 117: I would suggest also checking if someone tries to use a negative number.
WDYT?

pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+	if res.StatusCode != http.StatusOK {
+		c.log.WithFields(log.Fields{"statusCode": res.StatusCode, "responseBody": string(body)}).Error("inventory groups error response")
+		return nil, ErrGroupsRequestResponse
+	}
+
+	var response Response
+	err = json.Unmarshal(body, &response)
+	if err != nil {
+		c.log.WithField("error", err.Error()).Error("error occurred when unmarshalling response body")
```
de1987, line 164: if we have this type of error, having the body in the logs would be great for investigating the issue.

pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	response, err := c.ListGroups(ListGroupsParams{Name: name, PerPage: 1, Page: 1, OrderBy: "name", OrderHow: "ASC"})
+	if err != nil {
+		c.log.WithFields(log.Fields{"group-name": name, "error": err.Error()}).Error("failed when calling to ListGroups")
+		return nil, err
+	}
+	if len(response.Results) == 0 || response.Results[0].Name != name {
+		// return group not found when no group returned,
+		// or when the first found group has name different from the requested one.
+		c.log.WithField("group-name", name).Error("group not found")
```
de1987, line 184: If we have a scenario that returns more than one group
Adding the number of responses in the logs would be great for investigating it
WDYT?

pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+		c.log.WithFields(log.Fields{"statusCode": res.StatusCode, "responseBody": string(body)}).Error("inventory group error response")
+		return nil, ErrGroupsRequestResponse
+	}
+
+	var response Response
+	err = json.Unmarshal(body, &response)
+	if err != nil {
+		c.log.WithField("error", err.Error()).Error("error occurred when unmarshalling response body")
+		return nil, err
+	}
```
de1987, line 233: if we have this type of error, having the `body` in the logs would be great for investigating the issue.

pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+		c.log.WithFields(log.Fields{"statusCode": res.StatusCode, "responseBody": string(body)}).Error("inventory create group error response")
+		return nil, ErrGroupsRequestResponse
+	}
+
+	var group Group
+	err = json.Unmarshal(body, &group)
+	if err != nil {
+		c.log.WithField("error", err.Error()).Error("error occurred when unmarshalling response body")
+		return nil, err
+	}
```
de1987, line 292: if we have this type of error, having the `body` in the logs would be great for investigating the issue.

pkg/clients/inventorygroups/client.go
```diff
@@ -0,0 +1,355 @@
...
+	for key, value := range headers {
+		req.Header.Add(key, value)
+	}
+	client := clients.ConfigureClientWithTLS(&http.Client{})
+	res, err := client.Do(req)
+	if err != nil {
+		c.log.WithField("error", err.Error()).Error("inventory create group request error")
+		return nil, err
+	}
+	defer res.Body.Close()
```
de1987, line 275: I notice this code is repeating in the client implementation, I think we can create a reusable API rest caller function.
WDYT?

pkg/models/devicegroups.go
```diff
@@ -21,6 +21,7 @@ type DeviceGroup struct {
 	Type        string   `json:"Type" gorm:"default:static;<-:create"`
 	Devices     []Device `faker:"-" json:"Devices" gorm:"many2many:device_groups_devices;"`
 	ValidUpdate bool     `json:"ValidUpdate" gorm:"-:all"`
+	UUID        string   `json:"uuid,omitempty" gorm:"index"`
```
de1987, line 24: just to confirm, does `UUID` come from inventory?
Thanks

pkg/clients/inventorygroups/client.go
ldjebran: yes you are right , but I do not think that should be implemented in the context of the current PR , also this is the last implementation of a client :smiley:

pkg/clients/inventorygroups/client.go
ldjebran: the page is one and perpage is one too , so the number is always one.

pkg/models/devicegroups.go
ldjebran: yes correct, this is the id of the group in inventory /groups and it's in the format of uuid

pkg/clients/inventorygroups/client.go
ldjebran: done

pkg/clients/inventorygroups/client.go
ldjebran: done

pkg/clients/inventorygroups/client.go
ldjebran: done

ldjebran:
@de1987 addressed and responded to all you comment please revisit

pkg/clients/inventorygroups/client.go
ldjebran: done

ldjebran:
@de1987 thank you for the review, addressed and responded to all you comment please revisit.

de1987 (approved):

acosferreira (commented):
lgtm

acosferreira (approved):

PR #2228, opened 2023-09-13
add debug to service repo builder call for static delta

loadtheaccumulator:
# Description
Last of the debug tests added to UpdateService.BuildUpdateRepo() before adding static delta check logic

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor


loadtheaccumulator:
NOTE: code coverage is irrelevant in this change because the feature flags will go away and additional testing will be included with the actual logic in the next PR.

pkg/services/updates.go
```diff
@@ -363,16 +365,117 @@ func NewTemplateRemoteInfo(update *models.UpdateTransaction) TemplateRemoteInfo
...
+	// 			make permanent with the static delta logic
+	if feature.StaticDeltaDev.IsEnabled() {
+		// grab the original commit URL
+		updateCommit, err := s.commitService.GetCommitByID(update.CommitID, update.OrgID)
+		if err != nil {
+			return nil, err
+		}
+		update.Repo.URL = updateCommit.Repo.URL
+
+		s.log.WithField("updateTransaction", update).Info("services.BuildUpdateRepo StaticDeltaDev setup info")
```
ldjebran, line 408: as understand when updating major version we are going todo a rebase instead of upgrade on the device,
 maybe we should save all and return here in case of change ref.

what do you think ?

loadtheaccumulator: @ldjebran  100%.
This currently falls through and saves after static delta is gen'd or skipped below on line 460-475.
I'm essentially setting the repo to the original image repo and then overriding if a static delta is gen'd with the idea of save all as it returns at the bottom of the func. I could probably go ahead and remove the return on line 457, but was mainly looking for logs in this PR and dropping in the if delta exists logic in the next PR.

loadtheaccumulator: @ldjebran I'm on the same page on the skipping static delta if ref changes. I'll add that logic in with the if exists on next PR. Thanks again!

ldjebran (approved):
lgtm, when waiting for next PR :smile:

acosferreira (approved):
lgtm

PR #2235, opened 2023-09-17
add delta check and state read/write

loadtheaccumulator:
* add check for existing delta and refs changes
* add static delta state db read writes
* fix some FIXMEs used for dev
* add additional debugging

# Description
add static delta check and db read writes

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

### DIFF
--- a/pkg/services/repobuilder.go
+++ b/pkg/services/repobuilder.go
@@ -86,14 +86,15 @@ func (rb *RepoBuilder) BuildUpdateRepo(id uint) (*models.UpdateTransaction, erro
 		return nil, errors.New("repo unavailable")
 	}
 
-	// temporary to see actual values in various scenarios
-	// 1. single system update
-	// 2. multiple system update
-	// 3. update of system after rollback (current logic assumes updating from booted commit)
-	if feature.StaticDeltaDev.IsEnabled() {
+	// Skipping this process if the feature flag is enabled.
+	// Without static-deltas, this just downloads a commit repo and re-uploads as an update repo.
+	// This will retain the original commit repo URL as the update URL.
+	// (e.g., for a 2GB commit, this saves 4GB+ in traffic and local disk space on the pod)
+	if !feature.SkipUpdateRepo.IsEnabled() {
+		rb.log.Info("Starts building update repo...")
+
 		rb.log.WithField("updateTransaction", update).Info("Update transaction dev info")
 
-		// FIXME: break out to getFromCommit() and getToCommit()
 		fromCommitHash := "undefined"
 		fromCommit := &models.StaticDeltaCommit{
 			Rev: fromCommitHash,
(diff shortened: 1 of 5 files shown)


loadtheaccumulator:
I used github codespaces for this commit and PR.
Seeing weird lint errors.

pkg/services/repobuilder.go
```diff
@@ -108,119 +109,188 @@ func (rb *RepoBuilder) BuildUpdateRepo(id uint) (*models.UpdateTransaction, erro
...
 		rb.log.WithFields(log.Fields{
 			"old_commit_count": oldCommitCount,
 			"from_commit_hash": fromCommit.Rev,
 			"to_commit_hash":   toCommit.Rev,
 		}).Info("Static delta commit dev info")
 
-		// FIXME: StaticDeltaStatusNotImplemented is for dev only. Replace with actual status when done.
 		staticDeltaState := &models.StaticDeltaState{
-			Status: models.StaticDeltaStatusNotImplemented,
+			Name:   getStaticDeltaName(fromCommit.Rev, toCommit.Rev),
```
ldjebran, line 120: what will happen when 

fromCommit.Rev == "undefined"  or when fromCommit.Rev == ""

?

pkg/services/repobuilder.go
```diff
@@ -108,119 +109,188 @@ func (rb *RepoBuilder) BuildUpdateRepo(id uint) (*models.UpdateTransaction, erro
...
-					filepath.Clean(filepath.Join(stageCommitPath, "repo")))
-				if err != nil {
-					rb.log.WithField("error", err.Error()).Error("Error pulling static deltas")
-					return nil, err
+			return nil, fmt.Errorf("error chdir :: %s", err.Error())
+		}
+
+		// If there are any old commits, we need to download them all to be merged
+		// into the update commit repo
+		for _, commit := range update.OldCommits {
```
ldjebran, line 202: are we in contradiction here with line 105 where we take into consideration the first commit in the list , and here we are looping over,
should we consider only the first one?
what if the list of OldCommits is empty from the start ?

pkg/services/updates.go
```diff
@@ -367,21 +365,20 @@ func NewTemplateRemoteInfo(update *models.UpdateTransaction) TemplateRemoteInfo
 	}
 }
 
+func getStaticDeltaName(fromCommitHash string, toCommitHash string) string {
```
ldjebran, line 368: may be move this static function to staticDeltas model file

pkg/services/updates.go
```diff
@@ -396,26 +393,57 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
...
+		s.log.WithField("commit_id", update.CommitID).Debug("Getting commit by id")
+		result := db.Org(orgID, "").Joins("Repo").First(&systemCommit, update.CommitID)
+		if result.Error != nil {
+			s.log.WithField("error", result.Error.Error()).Error("Error searching for commit by commitID")
+			return nil, result.Error
+		}
+		s.log.WithField("commit", systemCommit).Debug("Commit retrieved")
+
+		fromCommit := &models.StaticDeltaCommit{
+			Rev: systemCommit.OSTreeCommit,
```
ldjebran, line 416: what happen when systemCommit.OSTreeCommit == ""  ?

pkg/services/repobuilder.go
loadtheaccumulator: @ldjebran Yeah. That's what I've been looking at with all of the testing. So far, [0] is the latest and count has always been 1. What I'd like to do is just validate the commits exist and pass them in as from_commit and to_commit--skipping this loop altogether.

pkg/services/updates.go
loadtheaccumulator: 100%. I copied it from my package and it was with the struct. Can definitely do that here too.

pkg/services/repobuilder.go
loadtheaccumulator: Added check coming up...

pkg/services/updates.go
loadtheaccumulator: I'll add a check for this to the genStaticDelta check with changes refs and delta doesn't exist above.

pkg/services/repobuilder.go
loadtheaccumulator: @ldjebran there are already a few moving parts and this section still works. thoughts on passing the to/from commits in the next PR?

pkg/services/repobuilder.go
loadtheaccumulator: check added

pkg/services/updates.go
loadtheaccumulator: done

pkg/services/updates.go
loadtheaccumulator: done

pkg/services/updates.go
```diff
@@ -396,26 +459,29 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
...
-		}
-		update.Repo.URL = updateCommit.Repo.URL
+	genStaticDelta := false
 
-		s.log.WithField("updateTransaction", update).Info("services.BuildUpdateRepo StaticDeltaDev setup info")
-	}
+	s.log.WithField("genStaticDelta", genStaticDelta).Debug("initial state of genStaticDelta")
 
-	genStaticDelta := false
+	if feature.StaticDeltaShortCircuit.IsEnabled() {
```
ldjebran, line 466: What StaticDeltaShortCircuit mean ?

pkg/services/updates.go
```diff
@@ -396,26 +459,29 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
 		return nil, result.Error
 	}
 
-	// FIXME: this temp short circuit will set the update repo to the original image repo
-	// 			make permanent with the static delta logic
-	if feature.StaticDeltaDev.IsEnabled() {
```
ldjebran, line 401: maybe we should delete this feature flag if not used anymore 
what do you think ?

pkg/services/repobuilder.go
ldjebran: cannot see where this is implemented, can you put the link ?

pkg/services/updates.go
loadtheaccumulator: just wiring in before the static delta kicks off.

pkg/services/repobuilder.go
loadtheaccumulator: updates.go L396-398 (for some reason won't give me the direct link).

I added a function to validate all of these items before the call to repobuilder. it will return and set genStaticDelta false so it will skip repobuilder.

ldjebran (approved):
lgtm

acosferreira (approved):
lgtm

PR #2253, opened 2023-10-02
add wait for ready to static delta process

loadtheaccumulator:
# Description
Add a "wait for ready" to the static delta process
Moved logic under a single dev feature flag
Removed some debug logging
Refactored status logic into a single switch

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

### DIFF
--- a/pkg/services/updates.go
+++ b/pkg/services/updates.go
@@ -438,10 +438,20 @@ func checkStaticDeltaPreReqs(edgelog *log.Entry, orgID string, update models.Upd
 	return staticDelta, nil
 }
 
+func saveUpdateTransaction(edgelog *log.Entry, updateTransaction *models.UpdateTransaction) error {
+	edgelog.WithField("repo", updateTransaction.Repo.URL).Info("Update repo URL")
+	updateTransaction.Repo.Status = models.RepoStatusSuccess
+	if err := db.DB.Omit("Devices.*").Save(&updateTransaction).Error; err != nil {
+		return err
+	}
+	err := db.DB.Omit("Devices.*").Save(&updateTransaction.Repo).Error
+
+	return err
+}
+
 // BuildUpdateRepo determines if a static delta is necessary and calls the repo builder
 func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.UpdateTransaction, error) {
 	var update *models.UpdateTransaction
-	var err error
 	var sderr error
 	var staticDelta *models.StaticDelta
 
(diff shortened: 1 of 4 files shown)


pkg/services/updates.go
```diff
@@ -462,105 +472,128 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
...
+			sdUpdate, sdErr := s.waitForStaticDeltaReady(update, staticDelta)
+			if sdErr != nil {
+				return nil, sdErr
+			}
+			return sdUpdate, nil
+		}
 	}
 
-	s.log.WithField("genStaticDelta", genStaticDelta).Debug("final state of genStaticDelta")
+	// FIXME: this is the current code falling through if feature flags aren't enabled
```
ldjebran, line 532: suppose we do not want with static deltas for moment, is this going to fail ?

ldjebran: or should we  make sure  feature.SkipUpdateRepo.IsEnabled()  to be false ?

pkg/services/updates.go
```diff
@@ -462,105 +472,128 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
...
-		if err := db.DB.Omit("Devices.*").Save(&update).Error; err != nil {
-			return nil, err
-		}
-		if err := db.DB.Omit("Devices.*").Save(&update.Repo).Error; err != nil {
-			return nil, err
+		if saveErr := saveUpdateTransaction(s.log, update); saveErr != nil {
+			return nil, saveErr
 		}
+	case <-errChan:
+		update.Repo.URL = staticDelta.ToCommit.URL
```
ldjebran, line 583: so we are setting the target commit repo url on any building error ?

pkg/services/updates.go
```diff
@@ -462,105 +472,128 @@ func (s *UpdateService) BuildUpdateRepo(orgID string, updateID uint) (*models.Up
...
 
-		return update, nil
+		s.log.WithField("repo", update.Repo.URL).Info("Update repo URL")
+		update.Repo.Status = models.RepoStatusSuccess
```
ldjebran, line 586: The repo building failed and we are setting the repo url and status as success , how will know that the problem is with the repo building , as we returning error and the device  update will not happen anyway ?

pkg/services/updates.go
loadtheaccumulator: That's the idea. Error but continue with the update anyway.

pkg/services/updates.go
loadtheaccumulator: Yeah. Let's talk about this one. There will be static delta errors, but there is a repo (the original image repo or a static delta repo if successful). I left that so it will continue as-is. We can rework that if necessary.

pkg/services/updates.go
loadtheaccumulator: I was thinking we would leave feature.SkipUpdateRepo.IsEnabled() true to skip the static delta now.

loadtheaccumulator: Or StaticDeltaDev feature flag to false and skip the switch logic to just use the image repo like we are currently without the delta.

loadtheaccumulator: feature.SkipUpdateRepo should probably be renamed with StaticDelta instead of Update

ldjebran: Ok , let's cleanup all this feature flags in next clean up PR

pkg/services/updates.go
ldjebran: I was thinking that we are returning an error , but realized that we are not :)

ldjebran (approved):
lgtm

pkg/services/updates.go
```diff
@@ -438,10 +438,20 @@ func checkStaticDeltaPreReqs(edgelog *log.Entry, orgID string, update models.Upd
 	return staticDelta, nil
 }
 
+func saveUpdateTransaction(edgelog *log.Entry, updateTransaction *models.UpdateTransaction) error {
+	edgelog.WithField("repo", updateTransaction.Repo.URL).Info("Update repo URL")
+	updateTransaction.Repo.Status = models.RepoStatusSuccess
+	if err := db.DB.Omit("Devices.*").Save(&updateTransaction).Error; err != nil {
+		return err
+	}
+	err := db.DB.Omit("Devices.*").Save(&updateTransaction.Repo).Error
```
de1987, line 447: for code consistency, I saw there are 5 calls for this function and 4 already set the `Repo.Status` to `Success` before calling it
I would suggest to have `updateTransaction.Repo.Status = models.RepoStatusSuccess` outside of this function, which also makes sense because of the function name.
@loadtheaccumulator WDYT?

de1987: Also, it will be important if needed to save `Repo` with `Status` different than `Success`

de1987 (approved):

PR #2378, opened 2024-01-22
device-update: implement deviceUpdateImagesFilters

ldjebran:
# Description
Implement deviceUpdateImagesFilters to allow to filter images for device update. The issue raised when the user click on update for a device, on the device update view there is a filter that allow to filter update target images, the request was sent to backend, but as the backend feature was not implement, as the filter feature was not implemented, the same result was returned. 

FIXES: https://issues.redhat.com/browse/THEEDGE-3795

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

# Checklist:

### DIFF
--- a/pkg/routes/devices.go
+++ b/pkg/routes/devices.go
@@ -33,7 +33,7 @@ func MakeDevicesRouter(sub chi.Router) {
 	sub.Route("/{DeviceUUID}", func(r chi.Router) {
 		r.Use(DeviceCtx)
 		r.Get("/dbinfo", GetDeviceDBInfo)
-		r.With(common.Paginate).Get("/", GetDevice)
+		r.With(common.Paginate).With(ValidateDeviceUpdateImagesFilterParams).Get("/", GetDevice)
 		r.With(common.Paginate).Get("/updates", GetUpdateAvailableForDevice)
 		r.With(common.Paginate).Get("/image", GetDeviceImageInfo)
 	})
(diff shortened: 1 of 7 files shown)


pkg/routes/devices.go
```diff
@@ -102,6 +102,83 @@ var devicesFilters = common.ComposeFilters(
...
+			deviceUpdateImagesFilters.Created = createdDated.Format(common.LayoutISO)
+		}
+		// any date parse error should have been caught in validation step
+	}
+
+	// define int filters
+	for _, key := range []string{"version", "additional packages", "all packages", "systems running"} {
+		paramValue := query.Get(key)
+		if paramValue != "" {
+			value, err := strconv.Atoi(paramValue)
```
omaciel, line 157: Maybe ignore `err` since it isn't really useful in this case? `value, _ := strconv.Atoi(paramValue)`?

pkg/services/devices.go
```diff
@@ -241,9 +242,55 @@ func (s *DeviceService) GetUpdateAvailableForDevice(device inventory.Device, lat
 	}
 
 	var images []models.Image
-	query := db.DB.Limit(limit).Offset(offset).Where("Image_set_id = ? and Images.Status = ? and Images.Id > ?",
+	query := db.DB.Debug().Limit(deviceUpdateImagesFilters.Limit).Offset(deviceUpdateImagesFilters.Offset).Where("Image_set_id = ? and Images.Status = ? and Images.Id > ?",
```
omaciel, line 245: Do we need to keep `Debug`?

omaciel (approved):
/lgtm

Left a couple of questions/suggestions.

pkg/routes/devices.go
ldjebran: I think we will need it, just in case we call it outside the api, it will very informative instead of a panic :)

pkg/services/devices.go
ldjebran: I needed that when delopping to ensure the sql was what I expected, will remove it

ldjebran: removed :)

omaciel (approved):
/lgtm

acosferreira (approved):
lgtm

PR #2443, opened 2024-03-06
fix empty Principal when using service account

loadtheaccumulator:
# Description
Replaces empty Principal element with OrgID when API request uses service token instead of user/pass.
Playbook Dispatcher errors with required Principal field being empty instead of len > 0

FIXES: THEEDGE-3907

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/services/updates.go
+++ b/pkg/services/updates.go
@@ -287,12 +287,18 @@ func (s *UpdateService) CreateUpdate(id uint) (*models.UpdateTransaction, error)
 	for _, device := range update.Devices {
 		device := device // this will prevent implicit memory aliasing in the loop
 		// Create new &DispatcherPayload{}
+		principal := identity.User.Username
+		if len(identity.User.Username) == 0 {
+			// FIXME: identity.ServiceAccount.Username when middleware identity v2.0 is prod, until then use update.OrgId
+			principal = update.OrgID
+		}
+
 		payloadDispatcher := playbookdispatcher.DispatcherPayload{
 			Recipient:    device.RHCClientID,
 			PlaybookURL:  playbookURL,
 			OrgID:        update.OrgID,
 			PlaybookName: "Edge-management",
-			Principal:    identity.User.Username,
+			Principal:    principal,
 		}
 		s.log.WithField("playbook_dispatcher", payloadDispatcher).Debug("Calling playbook dispatcher")
 		exc, err := s.PlaybookClient.ExecuteDispatcher(payloadDispatcher)


pkg/services/updates.go
```diff
@@ -287,12 +287,18 @@ func (s *UpdateService) CreateUpdate(id uint) (*models.UpdateTransaction, error)
...
+			principal = update.OrgID
+		}
+
 		payloadDispatcher := playbookdispatcher.DispatcherPayload{
 			Recipient:    device.RHCClientID,
 			PlaybookURL:  playbookURL,
 			OrgID:        update.OrgID,
 			PlaybookName: "Edge-management",
-			Principal:    identity.User.Username,
+			Principal:    principal,
```
ezr-ondrej, line 301: Is principal mandatory in the dispatcher payload?

pkg/services/updates.go
```diff
@@ -287,12 +287,18 @@ func (s *UpdateService) CreateUpdate(id uint) (*models.UpdateTransaction, error)
 	for _, device := range update.Devices {
 		device := device // this will prevent implicit memory aliasing in the loop
 		// Create new &DispatcherPayload{}
+		principal := identity.User.Username
+		if len(identity.User.Username) == 0 {
+			// FIXME: identity.ServiceAccount.Username when middleware identity v2.0 is prod, until then use update.OrgId
```
ezr-ondrej, line 292: When is that? Why do we hold off?
Service accounts are already available and identity v2 is backwards compatible if I get it right, so can we push v2 middleware now?

ezr-ondrej (commented):
I don't love to fallback to org.id, so couple questions to figure out if it's necessary, but if it is, lets go with it :)

pkg/services/updates.go
loadtheaccumulator: The readme lists v2 as experimental... https://github.com/RedHatInsights/platform-go-middlewares/tree/master and branches labelled beta, but I could have overlooked something.

ezr-ondrej (approved):
Ah, got the snag, the migration won't be so straight forward, lets do it in a follow up and just fix this for now :+1:
Thank you @loadtheaccumulator ! :)

pkg/services/updates.go
ezr-ondrej: gotcha

pkg/services/updates.go
loadtheaccumulator: yes

PR #2457, opened 2024-03-16
fix: don't depend on unleash-proxy

ezr-ondrej:
# Description

Unleash proxy is deprecated and unused anyway IIUC.

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


ezr-ondrej:
We probably do need it :shrug: I just have no idea why and it's deprecated, but I guess we can roll with it for now.

PR #2475, opened 2024-04-03
fix: defer log flush in main functions

lzap:
This pull request fixes the issue with log flushing in the main functions of the `cleanup`, `migraterepos`, and `images_iso` packages. Previously, the log was not being flushed properly, which could lead to incomplete or missing log entries. This PR adds the necessary code to ensure that the log is flushed before the application exits.

_Generated by AI._


lzap:
Oh `os.Exit` overrides `defer` I had no idea it works this way. Irrelevant then.

PR #2476, opened 2024-04-03
fix: do not log readiness checks

lzap:
SSIA

This line is logged several times every second for no reason.


ezr-ondrej (approved):
Yup, thanks @lzap :+1:

PR #2479, opened 2024-04-03
add unattended and sudoers options to blueprint

loadtheaccumulator:
# Description
Add unattended and sudoers options to the blueprint for ISO request

FIXES: THEEDGE-3922

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/clients/imagebuilder/client.go
+++ b/pkg/clients/imagebuilder/client.go
@@ -68,10 +68,16 @@ type OSTree struct {
 
 // Customizations is made of the packages that are baked into an image
 type Customizations struct {
-	Packages            *[]string     `json:"packages"`
-	PayloadRepositories *[]Repository `json:"payload_repositories,omitempty"`
-	Users               []User        `json:"users,omitempty"`
-	Subscription        *Subscription `json:"subscription,omitempty"`
+	Installer           *CustomInstaller `json:"installer,omitempty"`
+	Packages            *[]string        `json:"packages"`
+	PayloadRepositories *[]Repository    `json:"payload_repositories,omitempty"`
+	Users               []User           `json:"users,omitempty"`
+	Subscription        *Subscription    `json:"subscription,omitempty"`
+}
+
+type CustomInstaller struct {
+	SudoNoPasswd []string `json:"sudo-nopasswd"`
+	Unattended   bool     `json:"unattended"`
 }
 
 // Repository is the record of Third Party Repository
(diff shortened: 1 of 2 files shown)


pkg/clients/imagebuilder/client.go
```diff
@@ -341,7 +348,13 @@ func (c *Client) ComposeInstaller(image *models.Image) (*models.Image, error) {
 	}
 	users := make([]User, 0)
 	if feature.PassUserToImageBuilder.IsEnabled() && image.Installer != nil && image.Installer.Username != "" && image.Installer.SSHKey != "" {
-		users = append(users, User{Name: image.Installer.Username, SSHKey: image.Installer.SSHKey})
+		user := &User{
```
ezr-ondrej, line 351: We define it as pointer here, but then dereferencing it... I like the inline style actually, I'm all for keeping it :)
```
users = append(users, User{
                        Name:   image.Installer.Username,
			SSHKey: image.Installer.SSHKey,
			Groups: []string{image.Installer.Username, "wheel"},
		}
```

If we need the extra var lets just have the it as value, not a pointer :)

```
user := User{ .... }
users = append(users, user)
```
```

File: pkg/clients/imagebuilder/client.go [hunk 2]
```diff
@@ -367,6 +380,16 @@ func (c *Client) ComposeInstaller(image *models.Image) (*models.Image, error) {
...
 	}
+
+	customInstaller := &CustomInstaller{
+		Unattended:   true,
+		SudoNoPasswd: []string{"@wheel"},
+	}
+
+	if feature.BlueprintCustomInstaller.IsEnabled() {
+		req.Customizations.Installer = customInstaller
+	}
```
ezr-ondrej, line 391: I think we can inline it and that way not initiate when not needed :)

```suggestion
	if feature.BlueprintCustomInstaller.IsEnabled() {
		req.Customizations.Installer = &CustomInstaller{
			Unattended:   true,
			SudoNoPasswd: []string{"@wheel"},
		}
	}
```

ezr-ondrej (commented):
Styling nitpicks, the change make sense but I didn't tested it.

pkg/clients/imagebuilder/client.go
loadtheaccumulator: works for me. coming up...

pkg/clients/imagebuilder/client.go
loadtheaccumulator: works for me. coming up...

ezr-ondrej (approved):
Cool :+1:
Can we merge, or do we need to synchronize with the removal of the installer hook?

PR #2489, opened 2024-04-09
Disable codecov GHA

lzap:
Code coverage GHA is broken and currently blocking us.

I propose to remove this, I am not fan of code coverage. One can have 100% coverage with unit tests, yet the application can be still broken for end users.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -92,22 +92,3 @@ jobs:
 
       - name: Validate spec-file
         run: python3 -m openapi_spec_validator ${{ github.workspace }}/cmd/spec/openapi.json
-
-  gocoverage:
-    name: Codecov
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        name: Checkout edge-api
-
-      - name: Run Go Test Coverage output without fdo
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
> I propose to remove this, I am not fan of code coverage. One can have 100% coverage with unit tests, yet the application can be still broken for end users.

I think the benefit of codecov is: bot reminding you you've not added tests :)
I'm ok to merge this as you're right codecov has issues with PRs from forks, it's easy fix tho - just exposing the token publicly, which is not a huge issue IMO :)

lzap:
> I think the benefit of codecov is: bot reminding you you've not added tests :)

Go ahead and figure out how to configure the thing not to bug us when you add a single line and coverage goes 0.1% down or something. I still thing coverage for unit tests does not bring anything to the table. Some code do not need unit tests, it is up to humans (reviewers) to decide.

ezr-ondrej:
Yeah I agree that the coverage target should not be 100%, it should rather be a: hey there, haven't you forgotten to write a test for this? (which I usually answer yes I forgot, because I do not do TDD xD)
Tho I agree the value is minimal and if it's a struggle to get it working, lets ditch it :)

PR #2444, opened 2024-03-06
deps: use platform-go-middlewares v2

ezr-ondrej:
# Description
Migrate to identity v2 for better support of service accounts.

## Type of change

What is it?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

Needs https://github.com/RedHatInsights/edge-api/pull/2457

PR #2491, opened 2024-04-09
Log Gorm through logrus

lzap:
This pull request adds logging for Gorm using logrus. It includes a new GormLogger struct and a new function `prettyfier` in the logger package. The `InitLogger` function in the logger package is also updated to use the new GormLogger.

I created helper functions because `db.DB.WithContext(r.Context())` feels quite long. I am adding it at just few places tho, it is not realistic to add it everywhere. Only when we will debugging a bug on stage, we can add the context through the code path.

Query logs now contain latency, so we can easily pull long SQL queries on stage.

I will do a followup and create a prometheus metric for that to see how slow SQL queries really are in all envs.

### DIFF
--- a/logger/gorm_logger.go
+++ b/logger/gorm_logger.go
@@ -1,0 +1,73 @@
+package logger
+
+import (
+	"context"
+	"time"
+
+	"github.com/redhatinsights/platform-go-middlewares/request_id"
+	"github.com/sirupsen/logrus"
+	glogger "gorm.io/gorm/logger"
+)
+
+// GormLogger is a custom logger for GORM
+type GormLogger struct {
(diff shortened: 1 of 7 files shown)


ezr-ondrej (approved):
Looks pretty sane :)

logger/gorm_logger.go
```diff
@@ -0,0 +1,73 @@
+package logger
+
+import (
+	"context"
+	"time"
+
+	"github.com/redhatinsights/platform-go-middlewares/request_id"
```
lzap, line 7: @ezr-ondrej fyi this uses middlewares now

https://github.com/RedHatInsights/edge-api/pull/2444

PR #2495, opened 2024-04-11
Correct readiness and logging

lzap:
This pull request fixes an issue with the readiness status endpoint and improves logging in the code.

The app considered itself ready when openapi JSON file was present. Readiness should be used in a different way, for example for migrations, shutdown or other things. When not in use, it is much cleaner to simply return 200. Our infra requires it so that is the best thing to do.

More about this concept at:

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes


ezr-ondrej (approved):
Wow, that was interesting code indeed :stuck_out_tongue:

lzap:
> Wow, that was interesting code indeed 😛

I mean it was not terribly wrong, just not too useful. And it was leaving a log line on every ready request, there is a lot of them.

ezr-ondrej:
Yeah ... a lot of code, that is not too useful smells... I'm big proponent of least code possible :stuck_out_tongue:

PR #2497, opened 2024-04-12
Use FieldLogger and mockgen make target

lzap:
Logging is done through Event while logrus provides a clean interface called `FieldLogger` to do the same. Make use of that.

This pull request also adds a make target for generating mock files using mockgen. This will make it easier to generate and update mocks for different files in the project.

The mock changes had to be done because after FieldLogger change, mocks had to be regenerated.

I have no idea how the team were maintaining regeneration of these mocks but I could not find any script in this repo. Because two mock files are misplaced (one is actually present twice), I do believe this was manual work.

_Review commit by commit for cleaner experience._


lzap:
I also added back unit tests which were removed in https://github.com/RedHatInsights/edge-api/commit/06cfa7c55d24ad1f0c7932fefc43f337e4b6a011 as the intention was only to disable coverage reports. Finally, I made sure go-checkout actually uses cache to speed up compilation and test runs.

loadtheaccumulator (approved):
lgtm

PR #2499, opened 2024-04-15
Add simple job scheduler

lzap:
This pull request adds a simple job scheduler. It is a code ported from

https://github.com/RHEnVision/provisioning-backend/tree/main/pkg/worker

with memory-only implementation for now. Meaning jobs will be executed still on api pods, but if we decide to in the future, we can change it to SQL-based or Redis-based queue.

Kafka is not a good job queueing solution because load distribution across consumers is hash-based with default implementation being round-robin. Any sort of queueing business logic must be coded on top of that so there is no benefit. Redis, on the other hand, provide native queues. Or it can be SQL-based similarly to image builder's composer (job table, notify/subscribe for events).

For this reason, I propose to remove the WIP event-based Kafka solution and go with simplest possible in-process solution. As we get rid of big tasks and add Pulp integration, we may not need to move job processing into individual pods, but if we need to it is a trivial thing to. The implementation can be taken from provisioning service so this would be low-effort.

This should be an easy change as long as all arguments that are passed into background processing logic in the current codebase can be put into a struct that can be serialized later on (e.g. JSON/GOB). It is very important that all input data is not shared (e.g. no pointers to the "godlike" global state services in the `dependencies` package) as in the future workers may be running on different pods.

By default, `edge-api` spawns 10 managed worker goroutines with up to 10 queued jobs in total. Each job will expire after 4 hours (I hope to shorten this soon). The code logic must pass context into clients (e.g. HTTP client, SQL client) in order to receive a context error when that happens, at least into long-running steps. These numbers were made up, we need to tune these numbers once we implement Pulp, or even before if we choose to do this before that.

The proposed queue solves also context cancellation, panic and interrupts properly by providing a second (failure) handler that is called when job expires, is cancelled gracefully or interrupt signal is sent when k8s decides to kill the pod.

For easier unit testing, a separate implementation can be added which will use zero goroutines, the code will be executed straight in the `Enqueue` function and all the others (like `Start`, `Stop` or `Stats`) will be no-operations. This way, we can run tests in a consistent environment. Not part of this PR tho.

**How to use the scheduler**

A `Job` represents unit of work, it has a type, random uuid, identity and correlation (request_id typically) and an empty interface - anything that can be serialized. There are two interfaces for working with jobs, `Enqueuer` provides `Enqueue` method which adds a job to the queue and then `Worker` where handler functions get registered. A worker   can be started, gracefully stopped and provides statistics which can be monitored.

Each job type must have two handler functions registered:

* job handler: a function which is called to process a job. The input is the `Job` type, there is no output or error - all errors must be handled by the function (e.g. logged, written to db).
* failure handler: a function which is called when
  * job was running too long (timeout, 4 hours by default)
  * pod is shutting down gracefully (context cancel triggered by `Stop` method)
  * unhandled panic occurred in the job handler function
  * interrupt occured when k8s are about to kill the pod

Statistics provide two values:

* amount of active jobs (buys workers)
* amount of enqueued jobs (jobs waiting to be processed)

There is a hard limit of workers (10 in this proposal) which allows us to control the load on pods as well as hard limit of jobs which can be put into queue (10 as well). If this limit is reached, enqueuer will start blocking so requests may timeout. This means we will notice the problem in the monitoring soon as well as user will be informed something is not right.

As part of this PR I added metrics for HTTP requests (latency, result codes) and the new job queue so we can see amount of jobs entering the scheduler, number of finished/timeouted/panicked jobs and also latency.

I am also adding yet another (I know) feature flag to turn the new scheduler for "retry image build" operation. I have chosen this one because it is very easy to test it, just do `curl-stage POST /edge/v1/images/74265/retry` to schedule a new rebuild. My plan is to turn this on for my stage account and start testing this.

### DIFF
--- a/internal/middleware/prometheus.go
+++ b/internal/middleware/prometheus.go
@@ -1,0 +1,71 @@
+package middleware
+
+// Chi routes aware middleware, taken from https://github.com/766b/chi-prometheus
+// License: Apache 2.0
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var buckets = []float64{100, 200, 500, 750, 1000, 2000, 5000}
+
+const (
+	metricNameHTTPRequestTotal    = "edge_http_request_total"
+	metricNameHTTPRequestDuration = "edge_http_request_duration_ms"
+)
+
+// Middleware is a handler that exposes prometheus metrics for the number of requests,
+// the latency and the response size, partitioned by status code, method and HTTP path.
+type Middleware struct {
+	reqs    *prometheus.CounterVec
+	latency *prometheus.HistogramVec
+}
+
+// NewPatternMiddleware returns a new prometheus Middleware handler that groups requests by the chi routing pattern.
+// EX: /users/{firstName} instead of /users/bob
+func NewPatternMiddleware() func(next http.Handler) http.Handler {
+	var m Middleware
+	m.reqs = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name:        metricNameHTTPRequestTotal,
+			Help:        "HTTP requests count partitioned by numeric status code, text status code, method and HTTP path (chi route)",
+			ConstLabels: prometheus.Labels{"service": "edge-api"},
+		},
+		[]string{"code", "status_code", "method", "path"},
+	)
+	prometheus.MustRegister(m.reqs)
+
+	m.latency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        metricNameHTTPRequestDuration,
+		Help:        "Request duration partitioned by numeric status code, text status code, method and HTTP path (chi route)",
+		ConstLabels: prometheus.Labels{"service": "edge-api"},
+		Buckets:     buckets,
+	},
+		[]string{"code", "status_code", "method", "path"},
+	)
+	prometheus.MustRegister(m.latency)
+	return m.patternHandler
+}
+
+func (c Middleware) patternHandler(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
+		next.ServeHTTP(ww, r)
+
+		rctx := chi.RouteContext(r.Context())
+		routePattern := strings.Join(rctx.RoutePatterns, "")
+		routePattern = strings.ReplaceAll(routePattern, "/*/", "/")
+
+		c.reqs.WithLabelValues(strconv.Itoa(ww.Status()), http.StatusText(ww.Status()), r.Method, routePattern).Inc()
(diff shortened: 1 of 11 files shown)


ezr-ondrej:
Would it make sense to extract into separate module?
Or just depend on the implementation in provisioning for now?
I'm not strictly against copy just trying to see if we can do without it :)

loadtheaccumulator:
First pass comments:
No worries about adding another feature flag. They are necessary with our CI/CD system. we just have to be diligent about cleanup after things are baked-in and proven.

"dependencies" change +100!
Simplified unit testing, yea!
I like all the points you've made here.

I can give more background on the intended use for Kafka, but in a nutshell, the EDA effort wasn't specifically solving a queue issue. We're solving that with Pulp now, but we'll still need something like this simple job scheduler to replace the current commit-wait-loop and restart after Interrupt processes. SIDENOTE: I've already been working on PRs to remove the deprecated EDA code.

My first concern was additional complexity to support, but in looking at all of it I think the benefits outweigh any of that concern. **There's a lot here. In your opinion, is there anything in this PR that can be used outside of this specific feature flag and could be separate supporting PRs that go in first and serve other areas of the code?**

lzap:
> My first concern was additional complexity to support, but in looking at all of it I think the benefits outweigh any of that concern. There's a lot here. In your opinion, is there anything in this PR that can be used outside of this specific feature flag and could be separate supporting PRs that go in first and serve other areas of the code?

The last thing I want to add to this project is complexity. The current status of background processing is: randomly spawned goroutines without graceful shutdown (there is some SIGINT handling code at lest to write the status to the DB record) and more importantly without any scalability limits. The proposal adds proper queue with managed workers which can be limited (10 per pod at the moment) and graceful shutdown. Every feature of the worker is covered with test.

I can indeed separate the worker from the initial job (`RetryImageJob`) into separate PR, however, I would like to at least test it on a real world scenario. Because when anyone design something without actually using it, it is hardly correct. Oh I just noticed tests are green, I broke it with my last rebase, need to fix that.

loadtheaccumulator (approved):
let's give it a go. lgtm

internal/middleware/prometheus.go
```diff
@@ -0,0 +1,71 @@
...
+		start := time.Now()
+		ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
+		next.ServeHTTP(ww, r)
+
+		rctx := chi.RouteContext(r.Context())
+		routePattern := strings.Join(rctx.RoutePatterns, "")
+		routePattern = strings.ReplaceAll(routePattern, "/*/", "/")
+
+		c.reqs.WithLabelValues(strconv.Itoa(ww.Status()), http.StatusText(ww.Status()), r.Method, routePattern).Inc()
+		c.latency.WithLabelValues(strconv.Itoa(ww.Status()), http.StatusText(ww.Status()), r.Method, routePattern).Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
```
ezr-ondrej, line 68: I'm not saying it's useless, but we do have these from the gateway already.
I think this is still useful as the timing on the gateway and internal can differ, just saying :)

ezr-ondrej (approved):
Looks sane and should be safe if we turn it on only for stage at first :crossed_fingers:

PR #2501, opened 2024-04-15
Build all artifacts on CI

lzap:
SSIA

It is a good idea to fail early if the project does not build.

I also added another commit which further improves handling of temporary database files - uses defer trick to not to leave files behind when tests fail.


ezr-ondrej (approved):
Nice improvement :+1:

PR #2504, opened 2024-04-18
Add feature flag check with context

lzap:
This pull request adds a new function `CheckFeatureCtx` to the `feature` package that allows checking if a given feature is available with context. It also includes changes to the `main.go` and `images.go` files to use the new function.

I wonder how features were previously added per customer account. I heard that some customers had specific workflows.

### DIFF
--- a/unleash/features/feature.go
+++ b/unleash/features/feature.go
@@ -5,12 +5,13 @@
 package feature
 
 import (
+	"context"
 	"os"
 
 	"github.com/Unleash/unleash-client-go/v4"
 	unleashCTX "github.com/Unleash/unleash-client-go/v4/context"
-
 	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/platform-go-middlewares/v2/identity"
 )
 
 // FeatureCustomRepos is the const of the custom repo feature flag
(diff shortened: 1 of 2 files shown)


unleash/features/feature.go
```diff
@@ -161,20 +162,67 @@ func CheckFeature(feature string, options ...unleash.FeatureOption) bool {
...
+func OrgID(ctx context.Context) string {
+	value := ctx.Value(orgIDKey)
+	if value == nil {
+		return ""
+	}
+	return value.(string)
+}
+
+// WithOrgID returns context copy with trace id value.
+func WithOrgID(ctx context.Context, id string) context.Context {
```
ezr-ondrej, line 199: Could we get it from the identity, that's already stored in the context?
I'm not sure, but I think we already store even the orgId.

loadtheaccumulator: The orgid should already be in the identity. We could just pass the orgid to Unleash for every call and let the strategy determine if it's going to use it on the server side.

loadtheaccumulator, line 199: The orgid should already be in the identity. We could just pass the orgid to Unleash for every call and let the strategy determine if it's going to use it on the server side.

lzap: We could, but are you sure we want to decode base64 and parse JSON on every feature check through the request? There could be possibly dozens of checks per request, there maybe already are but not using context right now. What if we refactor this later and forget about it? I would rather parse it once in the middleware and store it, the cost is very low.

lzap: The same concern as above. Storing it parsed in the context again is I believe safer to avoid spending CPU cycles on this later.

lzap:
Rebased, I am passing `ctx` correctly now (thanks linter/Ondra) and use `UserId` correctly now.

unleash/features/feature.go
ezr-ondrej: I believe there are actually two identities in the context now.
Encoded: https://github.com/RedHatInsights/edge-api/blob/843ed53dfd19c185c3c636491a3f174de29e129f/pkg/routes/common/identity.go#L29
Unencoded: https://github.com/RedHatInsights/platform-go-middlewares/blob/77250f342961c26847b0f97178d718d0854eba36/identity/identity.go#L164

And we have a function that Gets OrgId from the unencoded:
https://github.com/RedHatInsights/edge-api/blob/843ed53dfd19c185c3c636491a3f174de29e129f/pkg/routes/common/org_id.go#L26

unleash/features/feature.go
```diff
@@ -161,20 +162,63 @@ func CheckFeature(feature string, options ...unleash.FeatureOption) bool {
 	return false
 }
 
+// CheckFeature checks to see if a given feature is available with context
+func CheckFeatureCtx(ctx context.Context, feature string, options ...unleash.FeatureOption) bool {
+	cfg := config.Get()
+	orgID := OrgID(ctx)
```
ezr-ondrej, line 168: I'd prefer using
https://github.com/RedHatInsights/edge-api/blob/843ed53dfd19c185c3c636491a3f174de29e129f/pkg/routes/common/org_id.go#L26

unleash/features/feature.go
ezr-ondrej: I think it's mostly about the code clarity. We'd have the OrgId three times set on the context.
That can become messy quite fast. Which one is set and ready to use?

Also Jonathan do you mean we should pass the context to all the toggles and let those toggle definitions decide if they need to use it?

That would be a good change IMO, but bit more distruptive as we'd need to change all the places we call the `CheckToggle` function.

lzap: Yeah I haven’t noticed we have that as well, sure.

lzap: I haven’t realized there is also the parsed identity, of course let’s use that one.

lzap:
And of course import cycle problem with the `common` package.

If you can, avoid `common` packages. A rule from the Effective Go documentation...

ezr-ondrej:
I think we can add the helper in the `platform-go-middlewares`: https://github.com/RedHatInsights/platform-go-middlewares/pull/80

lzap:
I rebased, since I do not have GoLand license anymore and to my surprise there are no good command line tools for moving methods to different packages, I used the deprecated method described in: https://go.dev/talks/2016/refactor.article

unleash/features/feature.go
```diff
@@ -155,20 +156,71 @@ func CheckFeature(feature string, options ...unleash.FeatureOption) bool {
 	return false
 }
 
+// CheckFeature checks to see if a given feature is available with context
+func CheckFeatureCtx(ctx context.Context, feature string, options ...unleash.FeatureOption) bool {
+	cfg := config.Get()
+	uctx := unleashCTX.Context{}
+	orgID, err := ci.GetOrgIDFromContext(ctx)
```
ezr-ondrej, line 163: What about `identity.GetIdentity(ctx).Identity.OrgID` ? :thinking:
We can easily handle empty string instead of error.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #2514, opened 2024-04-22
identity: Set the Original identity only once

ezr-ondrej:
# Description

We are already setting the Raw encoded identity here:

https://github.com/RedHatInsights/platform-go-middlewares/blob/48aa4a66e621e08b2cac8030e0da9b9e00efc780/identity/identity.go#L305

This removes setting it again in dependancies.Middleware

## Type of change

What is it?

- [x] Refactor


lzap (approved):
Good catch.

PR #2515, opened 2024-04-23
identity: Remove SetOriginalIdentity

ezr-ondrej:
# Description

## Type of change

What is it?

- [x] Refactor


lzap (approved):
Nice, I think we can also get rid of `GetOriginalIdentity` too.

PR #2517, opened 2024-04-23
Add logging of HTTP response code and latency

lzap:
This pull request adds logging of the HTTP response code and latency in the `logMiddleware` function. It provides more visibility into the request handling process and helps in troubleshooting and performance monitoring.


pkg/services/images.go
```diff
@@ -1472,12 +1473,15 @@ func (s *ImageService) RetryCreateImage(ctx context.Context, image *models.Image
 		return nil
 	}
 	if feature.JobQueue.IsEnabledCtx(ctx) {
+		orgID := identity.GetIdentity(ctx).Identity.OrgID
+		s.log.Infof("Enqueuing RetryCreateImageJob for org %s", orgID)
```
ezr-ondrej, line 1477: Shouldn't the orgId be logged automatically in fields?

ezr-ondrej: ugh, right, we should then I guess :)
I can look into that.

ezr-ondrej (approved):
Thanks @lzap, might help for sure ;)

PR #2519, opened 2024-04-24
eda remove services code

loadtheaccumulator:
# Description
Remove code for EDA image services

FIXES: HMS-3992

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor


Makefile
```diff
@@ -79,8 +79,6 @@ build:
 	go build -o build/edge-api-migrate-repositories cmd/migraterepos/main.go
 	go build -o build/edge-api-migrate-groups cmd/migrategroups/main.go
 	go build -o build/edge-api-ibvents cmd/kafka/main.go
-	go build -o build/edge-api-images-build pkg/services/images_build/main.go
```
loadtheaccumulator, line 82: @lzap Are we running tests outside of the container and inside the container?

loadtheaccumulator: Ah. compiling earlier in the pipeline?

lzap (approved):
Wow that is a decent amount of code. I am gonna trust the pipeline in this case.

Makefile
lzap: Yeah I added this recently, it is more of a technical problem - when you compile and run tests then GHA can cache the whole go cache to speedup subsequent builds. Also it is pretty fast with the cache and we get notified early if it does not compile. Finally, errors from the compiler are much nicer than from the linter.

PR #2531, opened 2024-05-01
add a toggle to log fields for dev and debug

loadtheaccumulator:
# Description
Adding a feature toggle for local development to log fields

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/unleash/features/feature.go
+++ b/unleash/features/feature.go
@@ -6,7 +6,9 @@ package feature
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/Unleash/unleash-client-go/v4"
 	unleashCTX "github.com/Unleash/unleash-client-go/v4/context"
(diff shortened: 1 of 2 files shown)


unleash/features/feature.go
```diff
@@ -178,6 +183,18 @@ func CheckFeatureCtx(ctx context.Context, feature string, options ...unleash.Fea
 	return false
 }
 
+// IsEnabledLocal returns a bool directly from the environment. Use before Unleash is init'd
+func (ff *Flag) IsEnabledLocal() bool {
```
loadtheaccumulator, line 187: Might be nice to dump print statement and leverage this method in IsEnabled() and IsEnabledCtx().
It would allow true/false in the env var value instead of defined or not.

lzap (approved):
Honestly, I think for this use case I think we should just remove my change alltogether. This is really no big deal.

I would find this approach to be useful for the noisy SQL debug records tho, this is really not too useful with Gorm.

unleash/features/feature.go
loadtheaccumulator: I'd like to just roll it in to IsEnabled() and skip Unleash if it isn't init'd.

PR #2533, opened 2024-05-02
update dev podman compose files

loadtheaccumulator:
# Description
Updating podman compose files for dev purposes only.
Updated README.
Added Containerfile-fedora
Updated postgres version from 10 to 12
Added recent feature flag examples to env template

FIXES: 

## Type of change

What is it?

- [x] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/podman/Containerfile-fedora
+++ b/podman/Containerfile-fedora
@@ -1,0 +1,78 @@
+############################################
+# STEP 1: build executable edge-api binaries
+############################################
+#FROM registry.access.redhat.com/ubi8/go-toolset:1.20 AS edge-builder
+FROM quay.io/fedora/fedora:39 AS edge-builder
+
+RUN dnf install -y go
+
+#WORKDIR $GOPATH/src/github.com/RedHatInsights/edge-api/
+
+COPY . .
+# Use go mod
+ENV GO111MODULE=on
+# Fetch dependencies.
+# Using go get requires root.
+USER root
+RUN go get -d -v
+
+# Build the binary.
+#RUN go build -tags=fdo -o /usr/bin/edge-api
+RUN go build -o /usr/bin/edge-api
+
+
+# Build the migration binary.
+RUN go build -o /usr/bin/edge-api-migrate cmd/migrate/main.go
+RUN go build -o /usr/bin/edge-api-wipe cmd/db/wipe.go
+RUN go build -o /usr/bin/edge-api-migrate-device cmd/db/updDb/set_account_on_device.go
+
+# Run the doc binary
+RUN go install github.com/swaggo/swag/cmd/swag@latest
+RUN ~/go/bin/swag init --generalInfo api.go --o ./api --dir pkg/models,pkg/routes --parseDependency
+RUN go run cmd/swagger2openapi/main.go  api/swagger.json api/openapi.json
+
+# Build the microservices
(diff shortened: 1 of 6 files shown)


podman/README.md
```diff
@@ -17,8 +16,127 @@ The containers described below share a common environment file for setting appli
...
+### Registries
+
+The default postgres used in the compose files is available from registry.redhat.io. You might need to authenticate to be able to pull the image.
+
+    podman login registry.redhat.io
+
+### Setup Data Directories
+
+    mkdir -p ~/devdata/postgresql
+    chmod 775 ~/devdata/postgresql
```
lzap, line 33: There is no need to create explicit directories, in fact, this is going to be painful in terms of SELinux and/or UNIX permissions when IDs do not match.

I suggest to simply let `podman` to manage volumes by itself, it can do better job of doing that.

podman/README.md
```diff
@@ -17,8 +16,127 @@ The containers described below share a common environment file for setting appli
...
+### Setup Data Directories
+
+Set up the data directories that will be mounted as volumes with the following commands:
+
+    mkdir -p ~/devgo
+    chmod 775 ~/devgo
+    mkdir -p ~/devdata/repos
+    mkdir ~/devdata/vartmp
+    mkdir ~/devdata/gocache
+    chmod -R 775 ~/devdata/
```
lzap, line 101: Again, I suggest to let podman do this for you. It is as simple as doing this:

```yaml
volumes:
  pg-backend:

services:
  backend-db:
    restart: on-failure
    image: docker.io/postgres:15.1
    profiles:
      - backend
      - backend-dev
    healthcheck:
      test: [ "CMD-SHELL", "pg_isready -d provisioning -U postgres && sleep 5" ]
      interval: 2s
      timeout: 3s
      retries: 10
    environment:
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres
      - POSTGRES_DB=provisioning
    volumes:
      - pg-backend:/var/lib/postgresql/data
    expose:
      - 5432
```

Podman will take care of creating the volume and managing it in the correct SELinux context. You can see those via `podman volumes list` I think, they will live somewhere in `.local/containers/storage` whatever, I do not remember.

podman/Containerfile-fedora
```diff
@@ -0,0 +1,78 @@
...
+# kickstart inject requirements
+COPY scripts/fleetkick.sh /usr/local/bin
+RUN chmod +x /usr/local/bin/fleetkick.sh
+COPY templates/templateKickstart.ks /usr/local/etc
+
+# template to playbook dispatcher
+#COPY templates/template_playbook_dispatcher_ostree_upgrade_payload.yml /usr/local/etc
+
+USER 1001
+CMD ["edge-api"]
```
lzap, line 78: I wish we could build on top of the official container file to keep the environment as close as possible to stage/prod.

lzap (changes requested):
Some ideas, I haven’t tested this yet.

I wonder - does it make sense to only containerize dependencies (Kafka+Postgres) and then build from source directly? This is approach I used the most when working on provisioning because I did not like how long it took for the container to build.

But if you are aiming for a full compose even with frontend, then it probably makes sense to have it all containerized.

PR #2544, opened 2024-05-14
jobs: job testing endpoints

lzap:
I wanted to test out the job scheduler, so I created two dummy endpoints which serve the only purpose - to call a noop job. One that is a success, one that has panic which leads to failure handler being called. It all works fine. I would like to get this to stage to see it working there first before I continue testing the image rebuild.

### DIFF
--- a/pkg/jobs/memory.go
+++ b/pkg/jobs/memory.go
@@ -2,6 +2,7 @@ package jobs
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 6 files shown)


pkg/jobs/memory.go
```diff
@@ -230,7 +230,7 @@ func (w *MemoryWorker) processJob(ctx context.Context, job *Job, wid uuid.UUID)
 		start := time.Now()
 		h(ctx, job)
 		elapsed := time.Since(start)
-		logger.Infof("Job %s of type %s completed in %s seconds", job.ID, job.Type, elapsed.Seconds())
+		logger.Infof("Job %s of type %s completed in %.02f seconds", job.ID, job.Type, elapsed.Seconds())
 		metrics.JobProcessedCount.WithLabelValues(string(job.Type), "finished").Inc()
 		metrics.BackgroundJobDuration.WithLabelValues(string(job.Type)).Observe(elapsed.Seconds())
 	} else {
```
lzap, line 236: Few changes I noticed when opened the file.

ezr-ondrej (approved):
other routes are in routes, but given these are not public routes, I think this is fine.

Let's do it :+1:

PR #2545, opened 2024-05-15
jobs: pass identity into retry job

lzap:
Identity was not passed, thus the job fails as it needs org_id for SQL.


pkg/jobs/job.go
```diff
@@ -108,7 +108,10 @@ func initJobContext(origCtx context.Context, job *Job) (context.Context, logrus.
 
 	id, err := identity.DecodeIdentity(job.Identity)
 	if err != nil {
-		logrus.WithContext(ctx).WithError(err).Warnf("Error decoding identity: %s", err)
+		logrus.WithContext(ctx).
+			WithError(err).
+			WithField("identity", job.Identity).
+			Warnf("Error decoding identity: %s", err)
 		id = identity.XRHID{}
 	}
```
lzap, line 116: Also sends identity into log in case decoding fails. Should be safe.

pkg/jobs/memory.go
```diff
@@ -70,12 +70,11 @@ func (w *MemoryWorker) Enqueue(ctx context.Context, job *Job) error {
...
-	_, logger := initJobContext(ctx, job)
-
 	if job.ID == uuid.Nil {
 		job.ID = uuid.New()
 	}
 
+	_, logger := initJobContext(ctx, job)
 	logger.WithField("job_args", job.Args).Infof("Enqueuing job %s of type %s", job.ID, job.Type)
 	w.q <- job
 	w.sen.Add(1)
```
lzap, line 80: Fixes null UUID being logged in the log statement there.

pkg/services/images.go
```diff
@@ -1332,28 +1332,33 @@ func init() {
...
 func (s *ImageService) RetryCreateImage(ctx context.Context, image *models.Image) error {
-	s.log = s.log.WithFields(log.Fields{"imageID": image.ID, "commitID": image.Commit.ID})
+	logger := log.WithContext(ctx).WithFields(log.Fields{"imageID": image.ID, "commitID": image.Commit.ID})
+
 	// recompose commit
 	image, err := s.ImageBuilder.ComposeCommit(image)
 	if err != nil {
-		s.log.WithField("error", err.Error()).Error("Failed recomposing commit")
+		logger.WithField("error", err.Error()).Error("Failed recomposing commit")
 		return err
```
lzap, line 1341: This is a good example of how logger should be used with context. Unfortunately, most of the service methods do not have `ctx` argument. We should be gradually adding them when needed.

Those service types should have never been introduced, feels like a Java engineer was writing Go for the first time. This method should have been just a function.

pkg/services/images.go
```diff
@@ -1332,28 +1332,33 @@ func init() {
...
 	if feature.JobQueue.IsEnabledCtx(ctx) {
 		orgID := identity.GetIdentity(ctx).Identity.OrgID
-		s.log.Infof("Enqueuing RetryCreateImageJob for org %s", orgID)
+		logger.Infof("Enqueuing RetryCreateImageJob for org %s", orgID)
 		job := jobs.Job{
-			Type: "RetryCreateImageJob",
-			Args: &RetryCreateImageJob{ImageID: image.ID},
+			Type:     "RetryCreateImageJob",
+			Args:     &RetryCreateImageJob{ImageID: image.ID},
+			Identity: identity.GetRawIdentity(ctx),
```
lzap, line 1354: This is the line.

lzap:
I am also dropping the custom "human readable" logger I wrote, it is not very useful honestly. Finally, adding a custom hook that will always ensure fields from context are present if context is present. This will make sure all jobs do have `job_id` logged properly.

logger/logger.go
```diff
@@ -82,12 +49,10 @@ func InitLogger(writer io.Writer) {
 			FieldMap: log.FieldMap{
 				log.FieldKeyTime: "@timestamp",
 			},
-			CallerPrettyfier: prettyfier,
 		})
-	} else if !feature.LogVerboseFields.IsEnabledLocal() {
-		log.SetFormatter(&plainFormatter{})
 	}
```
lzap, line 53: Dropping the custom formatter, prettyfier and the flag.

pkg/jobs/job.go
```diff
@@ -103,12 +104,30 @@ func WithJobID(ctx context.Context, id string) context.Context {
...
+}
+
+// WithCorrID returns context copy with trace id value.
+func WithCorrID(ctx context.Context, id string) context.Context {
+	return context.WithValue(ctx, corrIDCtxKey, id)
+}
+
 func initJobContext(origCtx context.Context, job *Job) (context.Context, logrus.FieldLogger) {
 	ctx := WithJobID(origCtx, job.ID.String())
+	ctx = WithCorrID(ctx, job.CorrelationID)
```
ezr-ondrej, line 123: I'm not sure about this, but don't we have a function that populates that field for a request context already and wouldn't we want to use the same (so it can be read the same in logger for example).

ezr-ondrej (approved):
Just a nit about whether we need a separate context value for jobs, not sure about that :)

pkg/jobs/job.go
ezr-ondrej: ugh wait, that is bit too gracious, there can be user in the identity and also it's quite large field, right?

pkg/jobs/job.go
lzap: This is intentionally copy as jobs might get moved to another process in the future.

ezr-ondrej: Yeah, something like that! :) It could still be set up explicitly, that's fine, just use the same field, so log can just read that given field wherever we are.

pkg/jobs/job.go
lzap: Ok removed.

pkg/jobs/job.go
ezr-ondrej: Right, dat one does not have an setter. Gotcha, lets have a separete one then :+1:

logger/ctx_hook.go
```diff
@@ -0,0 +1,59 @@
...
+		e.Data["correlation_id"] = cid
+	}
+
+	rid := request_id.GetReqID(e.Context)
+	if rid != "" {
+		e.Data["request_id"] = rid
+	}
+
+	return nil
+}
```
lzap, line 59: Just to be clear, this bit of code will ensure that if context is passed into logger, none of the fields above (org/job/request/user id) needs to be set, it will be done automatically. Meaning:

```
logrus.WithContext(ctx).WithField("org_id", "this is not needed anymore").
```

However, when logrus is not used with context, none of the fields will be added:

```
logrus.WithField("org_id", "still needed").
```

PR #2547, opened 2024-05-15
jobs: improve logging

lzap:
This further improves logging.

First off, it puts context into the dependency log pointers, so all common fields are guaranteed to be there.

Second, `logrus.WithContext(ctx)` is used in those methods in image building where context is present.

Finally, version commit and time is added in each message for better user experience in kibana. With this, you can tell immediately when new version was deployed on stage.

I am not sure why my Go insist on adding `go.sum` entries, I hope its okay. Not sure why, I have the same major version as it is used in `go.mod` yet it wants me to use newer versions of libraries. Not sure.


logger/ctx_hook.go
```diff
@@ -53,6 +90,7 @@ func (h *ctxHook) Fire(e *logrus.Entry) error {
 	rid := request_id.GetReqID(e.Context)
 	if rid != "" {
 		e.Data["request_id"] = rid
+		e.Data["requestId"] = rid
 	}
```
lzap, line 94: It is a mess in the codebase, instead of fixing all places, this will set both names.

pkg/dependencies/main.go
```diff
@@ -43,7 +43,7 @@ type EdgeAPIServices struct {
 func Init(ctx context.Context) *EdgeAPIServices {
 	account, _ := common.GetAccountFromContext(ctx)
 	orgID, _ := common.GetOrgIDFromContext(ctx)
-	log := log.WithFields(log.Fields{
+	log := log.WithContext(ctx).WithFields(log.Fields{
 		"requestId": request_id.GetReqID(ctx),
 		"accountId": account,
 		"orgID":     orgID,
```
lzap, line 49: This is the line.

pkg/services/images.go
```diff
@@ -1673,19 +1673,19 @@ func (s *ImageService) GetMetadata(image *models.Image) (*models.Image, error) {
...
+		log.WithContext(ctx).WithField("error", tx.Error.Error()).Error("Error saving image")
 		return nil, c, tx.Error
 	}
 	tx = db.DB.Save(&image.Installer)
 	if tx.Error != nil {
-		s.log.WithField("error", tx.Error.Error()).Error("Error saving installer")
+		log.WithContext(ctx).WithField("error", tx.Error.Error()).Error("Error saving installer")
 		return nil, c, tx.Error
 	}
 	image, err := s.ImageBuilder.ComposeInstaller(image)
```
lzap, line 1691: For the record, this is how it should be done, however, it is not possible to refactor this everywhere. The old method via `s.log` will work as well, no need to change it just for nothing. I just saw the opportunity because, for some reason, context is passed as `ctx` argument in these methods. In 95% other cases, it is not present so `s.log` is the only option.

logger/ctx_hook.go
ezr-ondrej: :( xD

ezr-ondrej (approved):
Thank you! :orange_heart:

PR #2548, opened 2024-05-15
jobs: store raw identity

lzap:
Image retry call fails because raw identity is not available in the job context, thus image builder client errors out.

### DIFF
--- a/pkg/jobs/job.go
+++ b/pkg/jobs/job.go
@@ -128,6 +128,7 @@ func initJobContext(origCtx context.Context, job *Job) (context.Context, logrus.
 		id = identity.XRHID{}
 	}
 	ctx = identity.WithIdentity(ctx, id)
+	ctx = identity.WithRawIdentity(ctx, job.Identity)
 
 	return ctx, logrus.WithContext(ctx).WithFields(
 		logrus.Fields{
--- a/pkg/services/images.go
+++ b/pkg/services/images.go
@@ -1309,20 +1309,21 @@ type RetryCreateImageJob struct {
 }
 
 func RetryCreateImageJobHandler(ctx context.Context, job *jobs.Job) {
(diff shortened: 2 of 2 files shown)


pkg/services/images.go
```diff
@@ -1309,20 +1309,21 @@ type RetryCreateImageJob struct {
...
 
-	s.processImage(ctx, args.ImageID, DefaultLoopDelay, false)
+	err := s.processImage(ctx, args.ImageID, DefaultLoopDelay, false)
+	if err != nil {
+		log.WithContext(ctx).Errorf("Process image returned error: %s", err)
+	}
 }
 
 func RetryCreateImageFailHandler(ctx context.Context, job *jobs.Job) {
-	s := NewImageService(ctx, log.StandardLogger()).(*ImageService)
```
ezr-ondrej, line 1319: nice start of the cleanup :stuck_out_tongue:

ezr-ondrej (approved):
Nice, thanks @lzap ! :orange_heart:

pkg/services/images.go
lzap: It is just failure handler that updates a database when job fails. Service is not needed here, it was only used for logging.

pkg/jobs/job.go
```diff
@@ -128,6 +128,7 @@ func initJobContext(origCtx context.Context, job *Job) (context.Context, logrus.
 		id = identity.XRHID{}
 	}
 	ctx = identity.WithIdentity(ctx, id)
+	ctx = identity.WithRawIdentity(ctx, job.Identity)
```
lzap, line 131: This is the line.

As always, everything else are my attempts to improve logging.

loadtheaccumulator (approved):
lgtm

PR #2550, opened 2024-05-15
jobs: fix hook order

lzap:
I added the context hook after cloudwatch hook, thus everything worked correctly locally and in RHOS terminal, but not in Kibana.


loadtheaccumulator (approved):
Lgtm

PR #2553, opened 2024-05-16
services: dbx helper

lzap:
Adds context to most if not all service `db.DB` or `db.Org` calls. This ensures SQL logs are correlated too.


pkg/services/services.go
```diff
@@ -26,3 +28,8 @@ func NewService(ctx context.Context, log log.FieldLogger) Service {
 func (s *Service) SetLog(log log.FieldLogger) {
 	s.log = log
 }
+
+// DB returns db.DBx with context stored in the service
+func (s *Service) DB() *gorm.DB {
+	return db.DBx(s.ctx)
+}
```
lzap, line 35: This is the helper that is used in this patch. The other `db.Orgx` was introduced earlier.

lzap:
Rebased to use `DBx/Orgx` methods instead, this is a tad cleaner.

ezr-ondrej (approved):
:orange_heart: it  :+1:

ezr-ondrej:
Eh the coverage is flaky, lets remove it and use snyk here instead: https://github.com/RedHatInsights/edge-api/pull/2555

ezr-ondrej (approved):
Thanks @lzap ! :orange_heart:

PR #2561, opened 2024-05-20
fix: race condition in kafka producer

lzap:
Read was not protected by the mutex. Randomly saw this in the codebase, probably no effect on behavior.


ezr-ondrej (approved):
Definitelly makes sense, nice catch! :+1:

PR #2563, opened 2024-05-21
refactor: job queue for more go routines

lzap:
This splits job queue into two - slow and fast, the latter being the default. This will prevent job starving when long-running jobs take all the resources (worker goroutines) and will not allow faster jobs to complete. The job size and worker pool size is guesswork for now, prometheus will show us numbers once we turn this on.

I reviewed all goroutines and transformed the code to use the job queue except four occurences which are soon going away with the pulp migration. Here is the list of new jobs:

```
$ ag NewAndEnqueue | egrep -o '\w+Job' | sort -u
CreateRepoForImageJob
CreateUpdateAsyncJob
ProcessImageJob
ProcessInstallerJob
ResumeCreateImageJob
RetryCreateImageJob
SyncDevicesWithInventoryJob
```

I made a guess on which jobs are likely slow and I use slow queue for them, we can update this later once we have some data from prometheus.

### DIFF
--- a/pkg/jobs/job.go
+++ b/pkg/jobs/job.go
@@ -23,12 +23,28 @@ type JobType string
 // will be called.
 type JobHandler func(ctx context.Context, job *Job)
 
+// JobQueue represents a queue where job is enqueued.
+type JobQueue int
+
+const (
+	// SlowQueue is a queue for jobs that can be processed slowly.
+	SlowQueue JobQueue = iota
+
+	// FastQueue is a queue for jobs that should be processed quickly.
+	FastQueue
+)
+
 // Job represents a single job. It is a message that is sent to a worker.
 // Job arguments are serialized so do not store pointers - workers must not share any
 // memory with the caller.
 type Job struct {
 	// Random UUID for logging and tracing. It is generated randomly by Enqueue function when blank.
 	ID uuid.UUID
+
+	// Job queue. Fast queue is for jobs that should be processed quickly, slow queue is for jobs
+	// that can be processed slowly. This allows better resource management. When not specified,
+	// job is enqueued to the slow queue.
+	Queue JobQueue
 
 	// Job type or "queue"
 	Type JobType
@@ -44,9 +60,10 @@ type Job struct {
 }
 
 // New creates new job and sets identity and correlation id from passed context.
-func New(ctx context.Context, jobType JobType, args any) *Job {
+func New(ctx context.Context, jobType JobType, jq JobQueue, args any) *Job {
 	return &Job{
 		ID:            uuid.New(),
+		Queue:         jq,
 		Type:          jobType,
 		Identity:      identity.GetRawIdentity(ctx),
 		CorrelationID: request_id.GetReqID(ctx),
@@ -56,6 +73,11 @@ func New(ctx context.Context, jobType JobType, args any) *Job {
 
 var ErrJobNotFound = errors.New("job not found")
 var ErrHandlerNotFound = errors.New("handler not registered")
+
+// IgnoredJobHandler is a handler that does nothing. It is used when no handler is registered for a job.
+var IgnoredJobHandler JobHandler = func(_ context.Context, _ *Job) {
+	// do nothing
+}
 
 // JobEnqueuer sends Job messages into worker queue.
 type JobEnqueuer interface {
--- a/pkg/jobs/memory.go
+++ b/pkg/jobs/memory.go
@@ -16,17 +16,20 @@ import (
 )
 
 type Config struct {
-	QueueSize int
-	Workers   int
-	Timeout   time.Duration
-	IntSignal os.Signal
+	FastQueueSize int
+	SlowQueueSize int
+	FastWorkers   int
+	SlowWorkers   int
+	Timeout       time.Duration
+	IntSignal     os.Signal
 }
 
 type MemoryWorker struct {
 	cfg Config
 	hs  map[JobType]JobHandler
 	fhs map[JobType]JobHandler
-	q   chan *Job
+	qF  chan *Job // fast
+	qS  chan *Job // slow
 	oc  *sync.Once
 	wg  sync.WaitGroup
 	cf  []context.CancelFunc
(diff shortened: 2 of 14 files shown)


pkg/jobs/job.go
```diff
@@ -57,6 +74,10 @@ func New(ctx context.Context, jobType JobType, args any) *Job {
 var ErrJobNotFound = errors.New("job not found")
 var ErrHandlerNotFound = errors.New("handler not registered")
 
+var IgnoredJobHandler JobHandler = func(_ context.Context, _ *Job) {
+	// do nothing
+}
+
```
lzap, line 80: Some jobs do not need failure handler, so this noop function is used instead.

pkg/jobs/memory.go
```diff
@@ -89,16 +100,32 @@ func (w *MemoryWorker) Start(ctx context.Context) {
...
-		go w.dequeueLoop(gctx, uid)
+		go w.dequeueLoop(gctx, uid, w.qS)
 	}
 
+	// Start stats loop
+	w.wg.Add(1)
+	gctx, cf := context.WithCancel(ctx)
+	w.cf = append(w.cf, cf)
+	go w.statsLoop(gctx)
+
```
lzap, line 128: Statistics are now a separate goroutine.

pkg/routes/images_test.go
```diff
@@ -1597,7 +1605,7 @@ var _ = Describe("Images Route Tests", func() {
 
 				mockImagesService.EXPECT().GetImageByID(strconv.Itoa(int(image.ID))).Return(&image, nil)
 				expectedError := errors.New("unknown error")
-				mockImagesService.EXPECT().CreateInstallerForImage(gomock.Any(), &image).Return(nil, nil, expectedError)
+				mockImagesService.EXPECT().CreateInstallerForImage(gomock.Any(), &image).Return(nil, expectedError)
```
lzap, line 1609: Since `CreateInstallerForImage` was introduced in the image interface and it is called from separate goroutine, this is needed in the mock.

pkg/services/images.go
```diff
@@ -49,7 +49,8 @@ type ImageServiceInterface interface {
 	UpdateImageStatus(image *models.Image) (*models.Image, error)
 	SetErrorStatusOnImage(err error, i *models.Image)
 	CreateRepoForImage(context.Context, *models.Image) (*models.Repo, error)
-	CreateInstallerForImage(context.Context, *models.Image) (*models.Image, chan error, error)
+	CreateInstallerForImage(context.Context, *models.Image) (*models.Image, error)
+	ProcessInstaller(ctx context.Context, image *models.Image) error
```
lzap, line 53: The `CreateInstallerForImage` method was split into two, the `ProcessInstaller` is called asynchronously from some callsites.

pkg/services/images.go
```diff
@@ -814,22 +840,19 @@ func (s *ImageService) processImage(ctx context.Context, id uint, loopDelay time
...
+				log.WithContext(ctx).WithField("error", err.Error()).Error("Failed creating installer for image")
 			}
+			err = s.ProcessInstaller(ctx, image)
 			if err != nil {
 				s.SetErrorStatusOnImage(err, image)
 				log.WithContext(ctx).WithField("error", err.Error()).Error("Failed creating installer for image")
 			}
+			if image.Status == models.ImageStatusInterrupted {
+				return err
+			}
```
lzap, line 855: The channel as a return value was a reason to split this. This was not good Go code at all.

pkg/services/updates.go
```diff
@@ -526,9 +558,14 @@ func (s *UpdateService) waitForStaticDeltaReady(update *models.UpdateTransaction
...
 
+			if ctx.Err() != nil {
+				edgelog.Error("timeout waiting for static delta to be ready")
+				break
+			}
+
 			time.Sleep(sleepTime)
 		}
-	}(s.log, staticDelta.State)
+	}(ctx, s.log, staticDelta.State)
```
lzap, line 568: This wait code could possibly loop forever, added timeout and cancel.

lzap:
Squash on merge.

loadtheaccumulator (changes requested):
Thanks for diving in on that! I made a suggestion or two. Another overall suggestion is separate the jobqueue feature flag into a feature flag per major function (e.g., images, device/update, etc) if submitting changes to multiple features in a single PR. And maybe splitting the infinite loop code changes into a separate PR so they can be tested (and reverted/fixed) independently without getting mixed in with the original purpose of the PR.

pkg/jobs/memory.go
```diff
@@ -89,16 +100,32 @@ func (w *MemoryWorker) Start(ctx context.Context) {
 	w.cfm.Lock()
 	defer w.cfm.Unlock()
 
-	logrus.WithContext(ctx).Infof("Starting %d job workers", w.cfg.Workers)
-	for i := 0; i < w.cfg.Workers; i++ {
+	logrus.WithContext(ctx).Infof("Starting %d fast job workers", w.cfg.FastWorkers)
```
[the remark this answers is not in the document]
lzap: That is a great approach in general and why I believe this can be a dogma for error messages (because of sentry/glitchtip), I believe for info/debug messages the rule can be loosened a bit. Here I see no value in providing the "N" number in a separate field.

pkg/services/updates.go
```diff
@@ -526,9 +558,14 @@ func (s *UpdateService) waitForStaticDeltaReady(update *models.UpdateTransaction
 				break DeltaLoop
 			}
 
+			if ctx.Err() != nil {
+				edgelog.Error("timeout waiting for static delta to be ready")
```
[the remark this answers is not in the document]
lzap: Yeah adding.

loadtheaccumulator:
I'm sure https://github.com/RedHatInsights/edge-api/pull/2576 probably broke this one with removing static delta all together.

pkg/jobs/job.go
lzap: Yeah used AI to generate one.

pkg/jobs/memory.go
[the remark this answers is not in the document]
lzap: The rule of thumb in my books is that I want a separate field if I need to search or filter records in a log system like Kibana.

pkg/jobs/memory.go
```diff
@@ -76,7 +82,12 @@ func (w *MemoryWorker) Enqueue(ctx context.Context, job *Job) error {
 
 	_, logger := initJobContext(ctx, job)
 	logger.WithField("job_args", job.Args).Infof("Enqueuing job %s of type %s", job.ID, job.Type)
-	w.q <- job
+
+	if job.Queue == FastQueue {
+		w.qF <- job
+	} else {
+		w.qS <- job
+	}
```
lzap, line 90: There are two queues now, fast is used for quick jobs.

lzap:
Rebased Ready for re-review.

loadtheaccumulator (approved):

lzap:
Feel free to merge **after** we confirm prod is stable. This patch might break things. TY!

PR #2565, opened 2024-05-23
refactor: preallocate package diff maps

lzap:
This is a wrap-up of my memory optimalisation research. I was mainly focusing on image building, this was not complete investigation of all parts of the application. Only what is on fire right now - image building.

In the first patch, I comment out pprof profiler, I am done with the process. Keeping the three lines in the codebase in case we need this again.

Second, it looks like the biggest offender during my (quite narrow) testing were package diff functions, while it could be optimized further, I am just preallocating maps to the specific size - this is always a good practice and this will save some copying.

The biggest issue is that kubernetes count cached pages into memory limit of a pod and while the application takes around 200-300 MB of memory when doing an image build, Linux filesystem cache goes up significantly (2-6 GB) which can ultimately lead to OOM kill of a pod. The third patch is an attempt to prevent that by using direct linux I/O kernel flag `O_DIRECT` which causes the read and write operations to bypass Linux kernel cache. I searched for only the relevant `.Open` and `os.Create` statements and added the flag there by changing this to the underlaying `os.FileOpen` call. This affects downloading and uploading of ISO files, tar extracting and S3 uploading, I did not do the change for irrelevant file operations (e.g. kickstart, tests, temporary files).

Most of this code will be removed once we implement the planned Pulp integration.

I noticed one file handler leak, fixed that as well.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,7 +15,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime/debug"
-	"strings"
 	"syscall"
 	"time"
 
@@ -126,10 +125,10 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		http.ServeFile(w, r, cfg.OpenAPIFilePath)
 	})
 
-	// Non-production routes
-	if strings.Contains(config.Get().EdgeAPIBaseURL, "stage") {
-		route.Mount("/api/edge/debug", middleware.Profiler())
-	}
+	// Uncomment to enable profiling
+	//if strings.Contains(config.Get().EdgeAPIBaseURL, "stage") {
+	//route.Mount("/api/edge/debug", middleware.Profiler())
+	//}
 
 	// Authenticated routes
 	authRoute := route.Group(nil)
--- a/main.go
+++ b/main.go
@@ -125,11 +125,6 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		http.ServeFile(w, r, cfg.OpenAPIFilePath)
 	})
 
-	// Uncomment to enable profiling
-	//if strings.Contains(config.Get().EdgeAPIBaseURL, "stage") {
-	//route.Mount("/api/edge/debug", middleware.Profiler())
-	//}
-
 	// Authenticated routes
 	authRoute := route.Group(nil)
 	if cfg.Auth {
--- a/pkg/services/devices.go
+++ b/pkg/services/devices.go
@@ -363,8 +363,8 @@ func (s *DeviceService) GetUpdateAvailableForDevice(device inventory.Device, lat
 }
 
 func getPackageDiff(a, b []models.InstalledPackage) []models.InstalledPackage {
-	diff := make([]models.InstalledPackage, 0)
-	pkgs := make(map[string]models.InstalledPackage)
+	var diff []models.InstalledPackage
+	pkgs := make(map[string]models.InstalledPackage, len(b))
 	for _, pkg := range b {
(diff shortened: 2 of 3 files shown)


pkg/services/devices.go
```diff
@@ -378,7 +378,7 @@ func getPackageDiff(a, b []models.InstalledPackage) []models.InstalledPackage {
 
 func getVersionDiff(new, old []models.InstalledPackage) []models.InstalledPackage {
 	var diff []models.InstalledPackage
-	oldPkgs := make(map[string]models.InstalledPackage)
+	oldPkgs := make(map[string]models.InstalledPackage, len(old))
 	for _, pkg := range old {
 		oldPkgs[pkg.Name] = pkg
 	}
```
lzap, line 384: This simply pre-allocates map so no copying is necessary - we know the length of the input slices in advance.

Turns out these functions are on the top of memory allocations, they do about 30% of all allocations of the application during my testing.

pkg/services/files.go
```diff
@@ -79,7 +80,7 @@ func NewFilesService(log log.FieldLogger) FilesService {
 // GetFile returns the file given a path
 func (s *LocalFilesService) GetFile(path string) (io.ReadCloser, error) {
 	path = "/tmp/" + path
-	f, err := os.Open(filepath.Clean(path))
+	f, err := os.OpenFile(filepath.Clean(path), os.O_RDONLY|syscall.O_DIRECT, 0)
```
lzap, line 83: The rest is the same - changing `Open` or `Create` calls to `OpenFile` which allows passing flags. Double check if I did not make a mistake:

* `Open` should be `OpenFile(filename, os.O_RDONLY, 0)`
* `Create` should be `OpenFile(filename, O_RDWR|O_CREATE|O_TRUNC, 0666)` 

Taken directly from the Go codebase.

pkg/services/files/uploader.go
```diff
@@ -201,10 +204,7 @@ func (u *S3Uploader) UploadFileWithACL(fname string, uploadPath string, acl stri
 		u.log.WithField("error", err.Error()).Error("Error uploading to AWS S3")
 		return "", err
 	}
-	if err := f.Close(); err != nil {
-		u.log.WithField("error", err.Error()).Error("Error closing file")
-		return "", err
-	}
+
```
lzap, line 207: This was a file handler leak in case the S3 PUT operation encountered an error.

lzap:
Ah tests are failing with `invalid argument` since Linux direct I/O needs to be 512B aligned, okay. There are Go libraries for that but I was hoping only open flags would do the trick. Since most of the code is going away, I am dropping this commit from the PR. So it is only the two problems fixed now.

ezr-ondrej (dismissed):
Nice finds :+1:

main.go
```diff
@@ -126,10 +125,10 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		http.ServeFile(w, r, cfg.OpenAPIFilePath)
 	})
 
-	// Non-production routes
-	if strings.Contains(config.Get().EdgeAPIBaseURL, "stage") {
-		route.Mount("/api/edge/debug", middleware.Profiler())
-	}
+	// Uncomment to enable profiling
+	//if strings.Contains(config.Get().EdgeAPIBaseURL, "stage") {
```
ezr-ondrej, line 129: linter does not like this. it's a two liner, I'd just delete it, maybe add it in `/docs/DEBBUGGING.md` as a tip instead if you want to go fancy.

ezr-ondrej: This needs to be fixed first :)

PR #2569, opened 2024-05-27
fix: Do not pass service context

ezr-ondrej:
# Description

Fixes issue with usage of cancelled context introduced in 6c62f065e77a7adbbd8daaebe53891ad03dc041a


lzap:
followup: https://github.com/RedHatInsights/edge-api/pull/2571

PR #2571, opened 2024-05-27
services: revert DBx/Orgx methods

lzap:
The shared context gets cancelled after a request ends. The DBx helper can only be used with properly passed context.

This reverts commit 6c62f065e77a7adbbd8daaebe53891ad03dc041a.

Rather than trying to find and fix all the background places, I would rather revert the change completely. If we want a SQL trace logging somewhere specifically, context should be passed normally and then `DB` can be renamed to `DBx(ctx)`.


ezr-ondrej (approved):
Agreed :+1:

PR #2574, opened 2024-05-29
fix: fix config init race condition

lzap:
This code is not correct:

```go
func Get() *EdgeConfig {
	if Config == nil {
		var lock = &sync.Mutex{}
		lock.Lock()
		defer lock.Unlock()
		Init()
	}
	return Config
}
```

Problem 1: mutex is created as a local variable, meaning it is stored on the stack, meaning every goroutine has its own copy, meaning it does not protect anything.

Problem 2: read is not protected, what might work on intel x86 right now might not be guaranteed elsewhere. Atomic pointers are available in `atomic.Pointer`, in the meantime, all read and writes should be protected via mutexes.

Problem 3: both `Config` and `Init` functions are exported, meaning, anyone can actually call the code or change the config variable without the mutex.

Now, all these problems do not appear to manifest into any bug because it looks like all of this synchronization code is not useful. Looks like `Init` is called at the start of the process so I am not sure why all of this lazy initialization logic is present when `Init` is exported.

But, I think this deserves to be fixed since it is a small change.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -118,7 +118,8 @@ type loggingConfig struct {
 	Region          string `json:"region,omitempty"`
 }
 
-var Config *EdgeConfig
+var config *EdgeConfig
+var configMu = sync.Mutex{}
 
 // DevConfigFile is a wrapper for local dev kafka edgeConfig
 type DevConfigFile struct {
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Great find, at minimum getting rid of useless mutex is huge benefit.

PR #2577, opened 2024-05-30
remove fdo from app image build

loadtheaccumulator:
# Description
Removing FDO from Dockerfile as first step in removing stale FDO code

FIXES: HMS-4010

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,14 +11,8 @@ ENV GO111MODULE=on
 USER root
 RUN go get -d -v
 
-# interim FDO requirements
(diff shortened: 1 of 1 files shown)


lzap (approved):
I see many other FDO references in the codebase, so I think we can take a stab on this since this is essentially not reachable code I assume.

loadtheaccumulator:
Definitely just the first step. I have a list in the ticket on what else can be deleted.

PR #2591, opened 2024-06-07
jobqueue: handle TERM signal not just INT

lzap:
It appears that kubernetes sends TERM signal not INT signal during pod restart, this was a typo on my side what I really meant was TERM. This patch adds handling of both signals.

This fixes the issue when app is restarted and image build is still running, the failure handler was not called thus leaving the image hanging in the pending state forever.

### DIFF
--- a/pkg/jobs/memory.go
+++ b/pkg/jobs/memory.go
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/google/uuid"
@@ -21,7 +22,7 @@ type Config struct {
 	FastWorkers   int
 	SlowWorkers   int
 	Timeout       time.Duration
-	IntSignal     os.Signal
+	IntSignal     []os.Signal
 }
 
 type MemoryWorker struct {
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Good find, thanks! :+1:

pkg/jobs/memory.go
```diff
@@ -60,7 +61,7 @@ func NewMemoryClient() *MemoryWorker {
 		FastWorkers:   100,
 		SlowWorkers:   10,
 		Timeout:       2 * time.Hour,
-		IntSignal:     os.Interrupt,
+		IntSignal:     []os.Signal{os.Interrupt, syscall.SIGTERM, syscall.SIGSTOP},
```
lzap, line 64: This is the default list of signals job queue considers to run failure handler:

* TERM used by k8s
* INT when you hit ctrl-c in local development
* STOP is rarely used in Linux I added it only for completenes (k8s does not appear to use it)

lzap:
Btw where is a unit test but for obvious reasons, it uses SIGUSR1 signal to simulate the use case. You cannot kill the test if you want it to pass :-D

lzap:
Merging JQ is not yet enabled so end to end testing will not test it. I want to test this manually on stage.

ezr-ondrej:
> Btw where is a unit test but for obvious reasons, it uses SIGUSR1 signal to simulate the use case. You cannot kill the test if you want it to pass :-D

Great self review! xD

PR #2603, opened 2024-06-14
chore: update deps

lzap:
SSIA


lzap:
@loadtheaccumulator any idea what is wrong? I actually see a different failure locally. Looks like some tests have some environmental dependencies.

lzap:
Closing for now. Would be nice to identify which dependency breaks tests.

PR #2618, opened 2024-06-28
Initialize dependencies just once

ezr-ondrej:
# Description

We are initializing dependencies already in https://github.com/RedHatInsights/edge-api/blob/8f736c36b9c69ba2c446e618b6976e8d54213298/main.go#L118

This is redundant, removing it.


ezr-ondrej:
This one is really easy after #2619

PR #2619, opened 2024-06-28
Logger without deps

ezr-ondrej:
# Description

Remove the need for dependencies from logger middleware, there is no benefit, we're overriding all the fields anyway.


ezr-ondrej:
I believe this is a no-brainer and easy to test, weekend pipelines will verify.

PR #2620, opened 2024-06-28
Untangle storage from the need of the dependencies context

ezr-ondrej:
# Description

The heavy initialization of all the services
was done for all routes, this is an attempt to untangle it.

We use context logger directly and initiate the only needed service on demand.


ezr-ondrej:
Lets have the pipeline test this :)

PR #2621, opened 2024-06-28
Load only needed data for transaction

ezr-ondrej:
# Description

We only need Repo URL


ezr-ondrej:
We'll need to test update anyway, this is a major improvement.

PR #2629, opened 2024-07-06
remove fdo routes

loadtheaccumulator:
# Description
Remove FDO routes from code

FIXES: HMS-4010

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor


loadtheaccumulator:
We can consider deprecation of the endpoints with a deprecation header and maybe 409 Gone--then execute on this removal in the future.

lzap:
LGTM

lzap:
Can you also add a migration to remove those tables from the DB? Or if you prefer to do it separately, I think thanks to the automigrate the database is in bad shape and we should do a thorough review and remove all fields/tables/indicies which are not used.

loadtheaccumulator:
> Can you also add a migration to remove those tables from the DB? Or if you prefer to do it separately, I think thanks to the automigrate the database is in bad shape and we should do a thorough review and remove all fields/tables/indicies which are not used.

Yeah. I'll do that in an additional PR. +1 on review of tables and fields

loadtheaccumulator:
Also confirmed with RHEL for Edge team this code is no longer required. I'll merge and test in stage in the AM.

PR #2632, opened 2024-07-08
Log slow queries as warnings

lzap:
This builds on top of the Gorm logger I added and while debug/trace is typically turned off on production, we would like to see slow queries in Kibana - they are subject of further investigation and optimalization. This simple change logs all slower queries than 2 seconds into warning logger so we can investigate them later.


ezr-ondrej (approved):
Nice, thanks :+1:

PR #2633, opened 2024-07-08
Trivial memory cache for update trans

lzap:
We probably want to implement a full Redis cache, but this will do as well:

https://github.com/RHEnVision/provisioning-backend/blob/main/internal/cache/redis.go

A simple memory cache taken from the provisioning service, it was already removed and replaced with Redis as we were caching many more objects in there:

https://github.com/RHEnVision/provisioning-backend/commit/decfb331a2e5642e904bed0dcd3ac41b319eb732

This will help us to identify if it solves the update problem or not.

### DIFF
--- a/main.go
+++ b/main.go
@@ -271,6 +271,8 @@ func main() {
 	jobs.Worker().Start(ctx)
 	defer jobs.Worker().Stop(ctx)
 
+	defer routes.UpdateTransCache.Stop()
+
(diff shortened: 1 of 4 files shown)


lzap:
java.io.IOException: Unexpected termination of the channel

loadtheaccumulator (approved):
lgtm

PR #2636, opened 2024-07-09
Add storage transfer metrics

lzap:
This pull request adds metrics for storage transfer operations. It includes counters for the number of download operations via S3 storage proxy, bytes transferred via S3 storage proxy, and the duration of S3 storage proxy operations. Additionally, it adds a counter for memory cache hit count by operation.


ezr-ondrej (approved):
Thanks! :+1:

PR #2652, opened 2024-07-17
Add content guards for Pulp

lzap:
This pull request adds content guards for Pulp to enhance security and access control. Content guards are created and ensured during the fixture creation process. The guards are associated with the distributions created for the repositories. This ensures that only authorized users can access and modify the content in the repositories.

Usage:

Use NewPulpServiceDefaultDomain method to create a default domain client and create a new domain if it does not exist yet for the OrgID.

Use NewPulpServiceWithDomain or NewPulpService to create a new client.

Use ArtifactsCreatePipe to stream TAR commit directly from IB S3 into Pulp. Make sure to not to buffer the whole file in memory or temporary file, use Reader and Writer interfaces.

Create repo with RepositoriesCreate and ensure content guards are created via ContentGuardEnsure. This method is idempotent and it corrects if content guards are misconfigured. It may not correct all possible states in that case it throws an error but normal cases like username change that will work automatically.

Use DistributionsCreate to create distribution and pass the existing repo and content guards into it.

Finally, use RepositoriesImport to perform a long running Pulp task to create distribution. This method will wait until the task is done in Pulp so it can block for quite some time.

All these methods are ment to be called from a background goroutine (job queue).

### DIFF
--- a/pkg/clients/pulp/guards_header.go
+++ b/pkg/clients/pulp/guards_header.go
@@ -6,20 +6,24 @@ import (
 	"fmt"
 
 	"github.com/google/uuid"
+	"github.com/redhatinsights/edge-api/pkg/ptr"
+	"github.com/sirupsen/logrus"
 )
 
 const JQOrgID = ".identity.org_id"
 
-func formatName(name string) string {
-	return "ORGID=" + name
+func headerGuardName(orgID string) string {
+	return "EDGE_HEADER_ORGID=" + orgID
 }
 
-func (ps *PulpService) HeaderGuardCreate(ctx context.Context, jq, orgID string) (*HeaderContentGuardResponse, error) {
+// HeaderGuardCreate creates a new header guard for the given orgID.
+func (ps *PulpService) HeaderGuardCreate(ctx context.Context, orgID string) (*HeaderContentGuardResponse, error) {
 	req := ContentguardsCoreHeaderCreateJSONRequestBody{
-		Name:        formatName(orgID),
+		Name:        headerGuardName(orgID),
 		HeaderName:  "x-rh-identity",
 		HeaderValue: orgID,
-		JqFilter:    &jq,
+		JqFilter:    ptr.To(JQOrgID),
+		Description: ptr.To("EDGE"),
 	}
 	resp, err := ps.cwr.ContentguardsCoreHeaderCreateWithResponse(ctx, ps.dom, req, addAuthenticationHeader)
 
@@ -34,6 +38,7 @@ func (ps *PulpService) HeaderGuardCreate(ctx context.Context, jq, orgID string) 
 	return resp.JSON201, nil
 }
 
+// HeaderGuardRead returns the header guard with the given ID.
 func (ps *PulpService) HeaderGuardRead(ctx context.Context, id uuid.UUID) (*HeaderContentGuardResponse, error) {
 	req := ContentguardsCoreHeaderReadParams{}
 	resp, err := ps.cwr.ContentguardsCoreHeaderReadWithResponse(ctx, ps.dom, id, &req, addAuthenticationHeader)
@@ -49,6 +54,7 @@ func (ps *PulpService) HeaderGuardRead(ctx context.Context, id uuid.UUID) (*Head
 	return resp.JSON200, nil
 }
 
+// HeaderGuardReadByOrgID returns the header guard for the given orgID.
 func (ps *PulpService) HeaderGuardReadByOrgID(ctx context.Context, orgID string) (*HeaderContentGuardResponse, error) {
 	hgl, err := ps.HeaderGuardList(ctx, orgID)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


pkg/clients/pulp/guards_rbac.go
```diff
@@ -0,0 +1,158 @@
...
+	if errors.Is(err, ErrRecordNotFound) {
+		cg, err = ps.RbacGuardCreate(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		return cg, nil
+	} else if err != nil {
+		u := (*cg.Users)[0].Username
+		if cg.Users == nil || len(*cg.Users) == 0 || u != rbacUsername() {
```
avitova, line 124: We cannot acces `u := (*cg.Users)[0].Username` before checking that `cg.Users != nil` and even `len(*cg.Users) != 0`

pkg/clients/pulp/guards_rbac.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	id := ScanUUID(*resp.JSON201.PulpHref)
+	rresp, err := ps.cwr.ContentguardsCoreHeaderAddRoleWithResponse(ctx, ps.dom, id, rreq, addAuthenticationHeader)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if rresp.JSON201 == nil {
+		return nil, fmt.Errorf("unexpected response: %d, body: %s", resp.StatusCode(), string(resp.Body))
```
avitova, line 106:
```suggestion
		return nil, fmt.Errorf("unexpected response: %d, body: %s", rresp.StatusCode(), string(rresp.Body))
```

avitova (commented):
Left some comments here. It's mostly pointers concerns.

pkg/clients/pulp/guards_header.go
```diff
@@ -63,22 +69,37 @@ func (ps *PulpService) HeaderGuardReadByOrgID(ctx context.Context, orgID string)
...
+		cg, err = ps.HeaderGuardCreate(ctx, orgID)
 		if err != nil {
 			return nil, err
 		}
 
-		return hg, nil
+		return cg, nil
 	} else if err != nil {
+		if cg.JqFilter == nil || *cg.JqFilter != JQOrgID {
+			logrus.WithContext(ctx).Warnf("unexpected Header Content Guard: %s, expected: %s, deleting it", *cg.JqFilter, JQOrgID)
```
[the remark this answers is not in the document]
ezr-ondrej: it should not happen as if cg.JqFilter == nil, it doesn't need to evaluate the second portion of the condition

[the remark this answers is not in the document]
lzap: Ah the deletion is done in the wrong code path - it should be done when a record is successfully found, not when there was a problem with fetching the data (e.g. timeout, network error etc)

pkg/clients/pulp/guards_rbac.go
lzap: Ah copy and paste error, yeah.

pkg/clients/pulp/guards_rbac.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.JSON200 == nil {
+		return nil, fmt.Errorf("unexpected response: %d, body: %s", resp.StatusCode(), string(resp.Body))
+	}
+
+	if resp.JSON200.Count >= DefaultPageSize {
```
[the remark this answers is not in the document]
lzap: I changed all of them, there were 5 of these checks in total.

lzap:
Amended changes, great comments. One bigger change - the "delete and create again" code (three instances) were moved out of the if block because that was an incorrect place for it. Please do not merge yet, Pulp stage is currently down I want to retest the changes.

pkg/clients/pulp/guards_header.go
```diff
@@ -63,22 +69,36 @@ func (ps *PulpService) HeaderGuardReadByOrgID(ctx context.Context, orgID string)
...
 
-		return hg, nil
+		return cg, nil
 	} else if err != nil {
 		return nil, err
 	}
 
-	return hg, nil
+	if cg != nil && (cg.JqFilter == nil || *cg.JqFilter != JQOrgID) {
+		logrus.WithContext(ctx).Warnf("unexpected Header Content Guard: %s, expected: %s, deleting it", *cg.JqFilter, JQOrgID)
```
avitova, line 87: This one we still can not dereference, because it is possible that `cg.JqFilter == nil` here.

pkg/clients/pulp/guards_rbac.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, err
+		}
+
+		return cg, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	if cg.Users == nil || len(*cg.Users) == 0 || (*cg.Users)[0].Username != rbacUsername() {
+		logrus.WithContext(ctx).Warnf("unexpected user for RBAC Content Guard: %s, expected: %s, deleting it", (*cg.Users)[0].Username, rbacUsername())
```
avitova, line 127: Same here, I think. We can not dereference or access` (*cg.Users)[0]` in `Warnf`, because .Users can be nil or the array can be empty.

avitova:
Thanks! Looks good, though I still have some worries around pointers, check the comments. Also, maybe we could add a helper function for `ScanUUID` that has a pointer argument?
Something like

```
func ScanUUIDHelper(href *string) uuid.UUID {
    if href == nil {
        *return something*
    }
    return ScanUUID(*href)
}
```

lzap:
I corrected it all (3 places), nil is being explicitly checked now beforehand, pointers dereferenced using the helper function from the `ptr` package, also changed the ScanUUID function as you recommended and overall the code is simplified. I suggest to see the diff, bunch of various changes. Most of them were result of further testing on stage, the ensure methods were not working properly at all since I was not able to test them yesterday :-D

I had issues with recreating RBAC CG from scratch, so I decided to check if required username is in the guard using `Contains` function. If not, the role is added.

avitova:
Thanks, I have no more comments, looks good to me.

pkg/clients/pulp/href_test.go
```diff
@@ -38,7 +38,9 @@ func TestScanUUID(t *testing.T) {
 
 		for _, tt := range tests {
 			t.Run(tt.name, func(t *testing.T) {
-				actual := ScanUUID(tt.href)
+				// linter G601 workaround
+				href := &tt.href
+				actual := ScanUUID(href)
```
ezr-ondrej, line 43: could we just
```suggestion
				actual := ScanUUID(&tt.href)
```

pkg/clients/pulp/guards_rbac.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if rresp.JSON201 == nil {
+			return nil, fmt.Errorf("unexpected response: %d, body: %s", rresp.StatusCode(), string(rresp.Body))
+		}
+	}
+	return cg, nil
+}
```
lzap, line 136: If the required IB readonly user is not present in a RBAC CG this ensure method does not try to delete it and create from scratch because that does not work - users and permissions are tied together already that is a ton of work. Instead, it makes "best effort" to insert the user into the existing CG. If that fails with an error, an admin must fix this.

Note the RBAC CG is only once per environment unlike header CG which is per-organization. That is why I am not spending too much time on this, the only use case is when this code is called is when username is actually changed. Should rarely happen.

pkg/clients/pulp/href_test.go
lzap: No, the same result, unhappy linter with a false positive. Figuring out a solution.

lzap: This linter was disabled in 1.22 because it is giving too many false positives, like in this case. I have more recent linter locally and it does not report this, so it is a bit slow trial and error approach for me to find out a good solution. What works for sure is to iterate via integer and use `[]` to access elements but that is so dumb to do. Trying now with `ptr.To` which should not be recognized by the linter I hope :-)

lzap: And it worked jeez.

PR #2653, opened 2024-07-18
fixes(HMS-4440): ADR-046 resource limits

lzap:
See ADR-046 for more info.

Review notes: please verify there are no changes except one.

The cleaner job currently has CPU limitation 800m with 400m allocation.

I propose to change this to 500/250 which is more aligned with the de facto standard on the cluster: CPU allocations divisible by 4 or 2. This will allow easier allocation by the scheduler.

I am assuming the cleanup job performs SQL queries, meaning most of the time CPU will be waiting. Thus I am leaning towards lower value than 1000m/500m.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -141,11 +141,11 @@ objects:
           value: ${SUBSCRIPTION_SERVER_URL}
         resources:
           limits:
-            cpu: ${{CPU_LIMIT}}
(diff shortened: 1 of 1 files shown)


ezr-ondrej (approved):
Thanks @lzap ! We need to make sure to update the values in app-interface when releasing (we do update them for prod now)

lzap:
Created an MR, thanks for the reminder.

PR #2663, opened 2024-07-26
Clean orphaned installed_packages (HMS-4477)

ezr-ondrej:
# Description
Clean up the orphaned installed_packages.

FIXES: HMS-4477


ezr-ondrej:
OTH I'm undrafting this.

:warning: This will likely fail as the commit_installed_packages is table with milion of records, so milion ids to match against 28 milion records, that can not be handled by the DB IMO, but it doesn't hurt to try.

loadtheaccumulator (approved):
lgtm

PR #2672, opened 2024-08-01
pulp: implement artifact pulling

lzap:
This pull request implements the functionality to pull artifacts in the Pulp software. It includes changes to the `pulp` package, adding the ability to create distributions, import repositories, and delete artifact versions. Additionally, it introduces a new utility function `ScanRepoFileVersion` to scan the version of a repository file. The changes also include updates to the test cases.


pkg/clients/pulp/repositories.go
```diff
@@ -41,10 +41,14 @@ func (ps *PulpService) RepositoriesImport(ctx context.Context, id uuid.UUID, rep
 		return nil, fmt.Errorf("unexpected response: %d, body: %s", resp.StatusCode(), string(resp.Body))
 	}
 
-	href, err := ps.WaitForTask(ctx, resp.JSON202.Task)
+	hrefs, err := ps.WaitForTask(ctx, resp.JSON202.Task)
 	if err != nil {
 		return nil, err
 	}
+	if len(hrefs) != 1 {
```
loadtheaccumulator, line 48: Are there situation where it would return anything other than 1?

loadtheaccumulator (approved):
lgtm

pkg/clients/pulp/repositories.go
lzap: Not here, other callsites - yes.

PR #2674, opened 2024-08-06
Cleanup orphaned pkg in cte batches

lzap:
This pull request includes changes to clean up orphaned installed packages in batches. Previously, the deletion process was inefficient as it sent a request for every single package. This PR improves the process by deleting orphaned packages in batches, resulting in better performance and efficiency.

The SQL is tested, however, I currently do not have a dev setup with postgres database. Will test tomorrow.

FYI the SQL will fail on SQLite3 it uses a non-standard statement USING:

```
FEATURE_CLEANUP_ORPHAN_COMMITS=1 go run ./cmd/cleanup
```

### DIFF
--- a/cmd/cleanup/cleanuporphancommits/cleanuporphancommits.go
+++ b/cmd/cleanup/cleanuporphancommits/cleanuporphancommits.go
@@ -3,6 +3,7 @@ package cleanuporphancommits
 import (
 	"errors"
 	"strings"
+	"time"
 
 	"github.com/redhatinsights/edge-api/cmd/cleanup/storage"
 	"github.com/redhatinsights/edge-api/config"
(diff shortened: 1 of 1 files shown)


lzap:
I have rewritten the query from

```
with to_delete as (
 select ip.id from installed_packages ip
 full join commit_installed_packages cip on ip.id = cip.installed_package_id
 where cip.commit_id is null limit 30000
)
delete from installed_packages
using to_delete
where to_delete.id = installed_packages.id;
```

to

```
delete from installed_packages
where installed_packages.id in (
	select ip.id from installed_packages ip
	full join commit_installed_packages cip on ip.id = cip.installed_package_id
	where cip.commit_id is null limit ?
);
```

in order to keep tests on sqlite3. I don’t think this will have dramatic impact on performance, what is important is that nothing is actually getting fetched to the client now. All the data transfer is the query and amount of affected rows. That should be faster and the query looks faster to me. Only numbers will prove tho, SQL optimizer in Postgres is crazy good.

loadtheaccumulator (approved):
lgtm

ezr-ondrej (approved):
Alright, lets try it out :shipit:

PR #2679, opened 2024-08-07
Re-apply: Manual migrations, drop timestamps

lzap:
This reverts commit 23be8599e622d1d2337368bb4acd0cd98dfd5f67.

We haven’t noticed the stuck delete operation was still running. We have killed it so we can proceed now.


ezr-ondrej (approved):
Thanks! :+1:

PR #2681, opened 2024-08-07
Make batchsize configurable

lzap:
This pull request adds the ability to configure the batch size for package deletion. Previously, the batch size was hardcoded to 1000, but now it can be set using the `CleanupBatchSize` configuration option. This allows for more flexibility and customization when performing cleanup operations.

The change also contains `backoffLimit` set to 1 according to [k8s docs](https://kubernetes.io/docs/concepts/workloads/controllers/job/). Jobs are supposed to be restarting until they return 0 regardless of the restart policy, it is a bit weird but this is how I understand it. We are unable to kill the old job it keeps poping back.


loadtheaccumulator (approved):
lgtm

PR #2682, opened 2024-08-08
Cleanup with sub select (HMS-4477)

ezr-ondrej:
# Description

Adds second implementation of the deletion of the installed_packages.
This allows us to switch quickly and compare the two.

Refs: HMS-4477


loadtheaccumulator (approved):
lgtm

PR #2683, opened 2024-08-09
Disabling parallel execution of tests

adarshdubey-star:
Switching the flag in-order to have a sequential execution in the pipelines


lzap (approved):
Reason?

adarshdubey-star:
> Reason?

The suite is getting executed twice both sequential and parallel and I believe it's not necessary to overwhelm the pipeline, also we are not able to report the results back to ibutsu portal because of the parallel execution as there is no selection happening in the tekton pipeline.

PR #2685, opened 2024-08-09
Disable constraint/trigger during cleanup

lzap:
I am shooting in the dark here, but there is actually a constraint thanks to a FK on the table that is very slow for deletions:

```
            Table "public.commit_installed_packages"
        Column        |  Type  | Collation | Nullable | Default 
----------------------+--------+-----------+----------+---------
 installed_package_id | bigint |           | not null | 
 commit_id            | bigint |           | not null | 
Indexes:
    "commit_installed_packages_pkey" PRIMARY KEY, btree (installed_package_id, commit_id)
Foreign-key constraints:
    "fk_commit_installed_packages_commit" FOREIGN KEY (commit_id) REFERENCES commits(id)
    "fk_commit_installed_packages_installed_package" FOREIGN KEY (installed_package_id) REFERENCES installed_packages(id)

```

Instead of deleting the foreign key completely, which is quite bit of work because GORM does not allow to disable FKs per table only globally, I am trying a trick that is mentioned in several articles and solutions on the internet. Since FK constraints are implemented via triggers, it could help deletion significantly.

### DIFF
--- a/cmd/cleanup/cleanuporphancommits/cleanuporphancommits.go
+++ b/cmd/cleanup/cleanuporphancommits/cleanuporphancommits.go
@@ -211,7 +211,24 @@ func CleanupOrphanInstalledPackages(gormDB *gorm.DB) error {
 		gormDB = db.DB
 	}
 
-	return cleanupOrphanInstalledPackagesCTE(gormDB)
+	tx := gormDB.Unscoped().Exec(`alter table commit_installed_packages disable trigger all`)
+	if tx.Error != nil {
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
Sure, passes locally on pgsql and unit tests just warn on sqlite, so I'm good with this, it's nicely limited and even if it fails, it just warns.

ezr-ondrej (approved):
Nice, lets put it to a test :)

PR #2686, opened 2024-08-12
Drop fk constraint (HMS-4477)

ezr-ondrej:
# Description

Drop the foreign key constraint for commit_installed_packages temporarily to speed up deletes.


lzap (approved):
Yeah I had the very same idea on Saturday, in fact I have the very same code here. Let’s try it.

PR #2690, opened 2024-08-12
Remove trigger disable cleanup statements

lzap:
This pull request removes the trigger disable cleanup statements from the code.

The app has no SU permissions, it did not work.

PR #2691, opened 2024-08-12
Revert "Disable constraint/trigger during cleanup (HMS-4477)"

ezr-ondrej:
Reverts RedHatInsights/edge-api#2685

We do not have permissions to do this, so this is actually only erroring out.


ezr-ondrej:
OTH we might even want to remove the clean-up completely, it's now useless with the cascade

lzap:
I am removing this here too: https://github.com/RedHatInsights/edge-api/pull/2690

I would rather keep this deletion because I do not trust GORM creating the necessary FKs there were some recent bugs reported on updating existing FKs (that it does not work reliably). We could still have non-cascading FKs in some environments.

PR #2695, opened 2024-08-20
fix old SA1006 lint errors

loadtheaccumulator:
# Description
Fixing simple SA1006 lint issues with formatted print statements and no formatting

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor

### DIFF
--- a/pkg/models/thirdpartyrepo.go
+++ b/pkg/models/thirdpartyrepo.go
@@ -4,7 +4,6 @@ package models
 
 import (
 	"errors"
-	"fmt"
 	"regexp"
 	"strings"
 
@@ -66,7 +65,7 @@ func (t *ThirdPartyRepo) ValidateRequest() error {
 		return errors.New(RepoNameCantBeInvalidMessage)
 	}
 	if !ValidateRepoURL(t.URL) {
-		return fmt.Errorf(InvalidURL)
+		return errors.New(InvalidURL)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/models/thirdpartyrepo.go
```diff
@@ -66,7 +65,7 @@ func (t *ThirdPartyRepo) ValidateRequest() error {
 		return errors.New(RepoNameCantBeInvalidMessage)
 	}
 	if !ValidateRepoURL(t.URL) {
-		return fmt.Errorf(InvalidURL)
+		return errors.New(InvalidURL)
```
ezr-ondrej, line 68: we could have the Error as a constant directly `InvalidURLErr = ....`. That way whoever consumes the package can use errors.Is() ... to tell which error was returned (if interested).
Not a big deal, we'd need to refactor all the errors here, which is probably not what we want to do.

ezr-ondrej: Would be a nice follow-up if you'd like :)

ezr-ondrej (approved):
Thank you! :+1:

pkg/models/thirdpartyrepo.go
loadtheaccumulator: Definitely. Constants are used elsewhere, so we can always add error refactor to the list.

PR #2697, opened 2024-08-21
Updating IQE trigger template for edge backend

adarshdubey-star:
The test runs are not getting reported to ibutsu which causes issues in debugging , as the template is quite old with the set params.

### DIFF
--- a/deploy/stage-post-deployment-tests.yaml
+++ b/deploy/stage-post-deployment-tests.yaml
@@ -1,29 +1,184 @@
 ---
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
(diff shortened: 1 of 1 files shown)


akhil-jha (approved):

ezr-ondrej (approved):
Akhil thinks it's fine, that's good enough for me. If you need a proper review, let me know, but I'm ok to get it in and see if it works :)
For proper review, I'd need to figure out how these templates work :stuck_out_tongue:

adarshdubey-star:
The deployment for eph env is failing for some reason , I see similar failure in the gitlab iqe-edge pipeline as well

PR #2702, opened 2024-09-09
fixes(HMS-4620): add RHEL 8.10

lzap:
SSIA

@loadtheaccumulator can you guide me what else I need to do in order to add RHEL 8.10?

### DIFF
--- a/config/repo_config.go
+++ b/config/repo_config.go
@@ -3,12 +3,15 @@
 package config
 
 // List of supported RHEL distributions
+// See https://github.com/osbuild/image-builder/blob/main/internal/v1/api.yaml#L1083
+// schema component Distributions
 const distRHEL84 = "rhel-84"
 const distRHEL85 = "rhel-85"
 const distRHEL86 = "rhel-86"
 const distRHEL87 = "rhel-87"
 const distRHEL88 = "rhel-88"
 const distRHEL89 = "rhel-89"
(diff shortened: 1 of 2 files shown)


loadtheaccumulator (changes requested):
You have two of the three changes. Just need to add it to the DistributionRefs map at Line 66

loadtheaccumulator (dismissed):
lgtm

loadtheaccumulator:
Ugh. @lzap Test failing, and not sure why 81 hasn't failed before now.

```
 --- FAIL: TestValidateRepo (0.00s)
    --- FAIL: TestValidateRepo/validate_major#01 (0.00s)
        repo_config_test.go:27: "81" not found: ""
FAIL
FAIL	github.com/redhatinsights/edge-api/config	0.006s
```

loadtheaccumulator:
I see the line in the test. It should only be failing in the test, and not elsewhere in the code.
It should get passed through, although, I am not sure what value IB is expecting.

lzap:
Right, it appears that image builder has now two forms for distros: `rhel-89` and `rhel-8.10`. I have rewritten the test to use regular expression to capture both formats.

What is next? Do we need to add this to UI?

loadtheaccumulator (approved):
lgtm

loadtheaccumulator:
UI is next. I can also use 8.10 for my next test this morning via the API

PR #2706, opened 2024-09-18
store commit tar in pulp

loadtheaccumulator:
# Description
Import Image Builder OSTree commit tar file into a Pulp Repository

* Adds internal repostore package
* Updates repobuilder and image to call Pulp or AWS store based on feature flag toggle
* Prepares for unit tests

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


pkg/services/internal/pulpcommit.go
```diff
@@ -0,0 +1,270 @@
...
+type PulpCommit struct {
+	OrgID      string
+	SourceURL  string
+	Domain     *pulpDomain
+	FileRepo   *pulpFileRepository
+	OSTreeRepo *pulpOSTreeRepository
+}
+
+// Store imports an OSTree repo into Pulp
+func (pc *PulpCommit) Store(ctx context.Context, commit models.Commit, edgeRepo *models.Repo) error {
```
lzap, line 28: Parameter named `commit` does not need to be there, only `OrgID` is used. The same for `edgeRepo`, try to minimize dependencies of this package, just use primitive types.

pkg/services/internal/pulpcommit.go
```diff
@@ -0,0 +1,270 @@
...
+	log.WithContext(ctx).Info("Pulp OSTree Repo created with Content Guard and Distribution")
+
+	if err = pc.OSTreeRepo.Import(ctx, pserv, pc.OSTreeRepo.Name, edgeRepo, *pc.FileRepo); err != nil {
+		log.WithContext(ctx).Error("Error importing tarfile into pulp ostree repository")
+
+		return err
+	}
+	log.WithContext(ctx).Info("Image Builder commit tarfile imported into pulp ostree repo from pulp file repo")
+
+	return nil
```
lzap, line 74: The types and interfaces you introduced makes this function worse in my opinion. Readability is king, if I have to choose between this and a single function that can be read top to bottom, I am going for the long function even if I am breaking up the dogma of "function must not be longer than N lines". Functions can always be easily broken down to sub-functions and arguments can be grouped together into structs that are only used as data carriers.

pkg/services/internal/pulpcommit.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+type pulpDomain struct {
+	Name string
+	UUID uuid.UUID
+}
```
lzap, line 80: I see little benefit in this type, domain can be just `string`, in fact that is exactly what the pulp client wants.

pkg/services/internal/pulpcommit.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+type pulpOSTreeRepository struct {
+	ID           uint
+	Name         string
+	CGPulpHref   string
+	PulpHref     string
+	Distribution *pulp.OstreeOstreeDistributionResponse
+}
```
lzap, line 161: Ditto.

pkg/services/internal/pulpcommit.go
```diff
@@ -0,0 +1,270 @@
...
+	edgeRepo.Status = models.RepoStatusSuccess
+
+	edgeRepoURL, _ := url.Parse(edgeRepo.URL)
+	log.WithContext(ctx).WithFields(log.Fields{
+		"status":                edgeRepo.Status,
+		"repo_distribution_url": edgeRepoURL.Redacted(),
+	}).Debug("Repo import into Pulp complete")
+
+	return nil
+}
```
lzap, line 270: This could have been just a single function, granted very long but it could be broken down to:

```
func DoSomething(a, b, c, d) (result, finish, error) {
 doStep1(a, b, c)
 doStep2(b, c, d)
 return result, func(){ deleteRepository(b, c) }, nil
}

func doStep1(a, b, c)

func doStep2(b, c, d)
```

It returns a result (status) that can be stored into database, a function that can be deferred and performs a deletion of resources that are no longer needed and of course an error in case things go wrong. The caller:

```
r, finish, err := DoSomething(a, b, c, d)
if err != nil {
 ...
}
defer finish()
repository.Status = r
repository.Save()
```

A small step above would be introducing `struct` types (without any methods) just to group some arguments together in case the sub-functions have too many arguments. And only if this does not work, I would consider creating a new object type:

```
type Importer struct {
 Input1 type,
 Input2 type,
 Input3 type,
 private1 type,
 private2 type,
 private3 type,
}

func (i *Importer) DoSomething() (r, finish, err)
```

PR #2707, opened 2024-09-18
store commit tar in pulp

loadtheaccumulator:
# Description
Import Image Builder OSTree commit tar file into a Pulp Repository

    Adds internal repostore package
    Updates repobuilder and image to call Pulp or AWS store based on feature flag toggle

FIXES: HMS-4119

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


pkg/services/repostore/pulpstore.go
```diff
@@ -0,0 +1,230 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/redhatinsights/edge-api/config"
+	"github.com/redhatinsights/edge-api/pkg/clients/pulp"
```
lzap, line 12: Nice, without dependency on `models` it is easier to test if you choose to test it on the PulpService level creating a proper mock for the interface. See, you created a mock for the `PulpRepoStore` function which leaves all the code you wrote in this PR untested. I suggest not to do that and create similar mock for the PulpService interface.

Note there is no such interface at the moment, you need to create one. You can even break it up into smaller interfaces if you want - massive interfaces are not welcomed by Go community, tho, we have used them in the past and it is everywhere in this codebase so it is fine.

pkg/services/repostore/pulpstore.go
```diff
@@ -0,0 +1,230 @@
...
+	if err = pserv.FileRepositoriesVersionDelete(ctx, pulp.ScanUUID(&fileRepoVersion), pulp.ScanRepoFileVersion(&fileRepoVersion)); err != nil {
+		return "", err
+	}
+	log.WithContext(ctx).WithField("filerepo_version", fileRepoVersion).Info("Artifact version deleted")
+
+	return repoURL, nil
+}
+
+// creates a domain service for a specific org if one does not already exist
+func createDomainService(ctx context.Context, orgID string) (*pulp.PulpService, error) {
```
lzap, line 73: Why this function does not actually find an existing domain? We only need to create it once per account, it lists domains and then take action only when it is missing. I would expect a function named `findOrCreateDomain` or something like that.

pkg/services/repostore/pulpstore.go
```diff
@@ -0,0 +1,230 @@
...
+		stageDistURL.User = url.UserPassword(cfg.PulpContentUsername, cfg.PulpContentPassword)
+
+		distURL = stageDistURL.String()
+	}
+
+	parsedURL, _ := url.Parse(distURL)
+	log.WithContext(ctx).WithField("distribution_url", parsedURL.Redacted()).Debug("Distribution URL retrieved")
+
+	return distURL, err
+}
```
lzap, line 187: We will be getting rid of RBAC CG in favor of a different one based on TLS cert unfortunately, this will go away.

pkg/services/repostore/pulpstore.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.WithContext(ctx).Error("Error importing tarfile into pulp ostree repository")
+
+		return "", err
+	}
+	parsedURL, _ := url.Parse(repoURL)
+	log.WithContext(ctx).WithField("repo_url", parsedURL.Redacted()).Info("Image Builder commit tarfile imported into pulp ostree repo from pulp file repo")
+
+	// TODO: move this to parallel or end of entire process (post Installer if requested)
+	if err = pserv.FileRepositoriesVersionDelete(ctx, pulp.ScanUUID(&fileRepoVersion), pulp.ScanRepoFileVersion(&fileRepoVersion)); err != nil {
```
lzap, line 64: We discussed this, we cannot use another job because there is no delayed job processing available, it will start immediately. Creating another function and calling it later is possible, but make sure you call it via `defer` keyword to ensure it is also deleted when error occurs. Alternatively, a function with closure can be returned from here.

lzap (changes requested):
This is so much better, now, two major concerns.

First, domain creation does not find existing domain.

Second, I suggest you move mocking one level down from `PulpRepoStore` mock to `PulpService` mock. This way you can cover the newly written tests completely.

pkg/services/repostore/pulpstore.go
loadtheaccumulator: The tarfile is imported into the ostree repo above. Once that is done, there shouldn't be a dependency on this artifact.

pkg/services/repostore/pulpstore.go
loadtheaccumulator: It still does both. I can rename it. Thinking simply domainService().

pkg/services/repostore/pulpstore.go
loadtheaccumulator: I'll delete the comment. I don't think the time it's spending deleting ( a couple of minutes or less at worst) is worth the effort at this point. We can inject it later.

pkg/services/repostore/pulpstore.go
loadtheaccumulator: Mockgen updated the mock for repobuilder.go because of the StoreRepo() call I used instead of going down the unnecessary rabbit hole of updating existing tests for the AWS store when using ImportRepo.

Per the recommendation, there is no testing in this PR.

loadtheaccumulator:
Domain creation leverages the service for finding existing or creating a new domain when it doesn't. I'll update the comment to be more clear on what it's doing. I'm pushing an updated here in 3, 2, ...

lzap (approved):
Looks great.

We agreed to write tests later, probably the way to go will be to mock PulpService to test this code. We just want to start playing around with this on stage.

PR #2709, opened 2024-09-27
Updating the base image to ubi9

adarshdubey-star:
# Description

Updating the base image from ubi8 to ubi9


lzap (approved):
Oh that was easy. Nice. Thanks.

PR #2722, opened 2024-10-29
refactor pulp config

loadtheaccumulator:
# Description
Refactor pulp config

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor


ezr-ondrej:
Looks ok, I'm not sure why we would want a separate structure, when we're extensively naming all that config anyway, but I don't see an issue with it :)

lzap (approved):
I see flat configuration structures as superior to nested ones, they nicely map to ENV variables and it is just more clear.

But I am okay with this if you want it. Merge at will.

loadtheaccumulator:
I'm good with either way. Per our offline discussion... there are multiple options (including cleanenv, etc), but I was thinking sticking with viper and how some of the others are using it would be good until the entire config is refactored and can be more easily ported. I avoided going one or two levels deeper to keep it flat-ish. The pulp-specific load was to avoid breaking existing config use scattered around the app, and we can refactor into that mechanism as we go. Thanks again for the discussion!

PR #2725, opened 2024-10-30
refactor: remove unleash-proxy dependency (HMS-4929)

ezr-ondrej:
# Description

Moved validation function to `config` package.

Removed check for ephemeral environment as there will be instance of unleash with flags off, so no need to return for ephemeral.
`FeatureFlagsService` config was never actually used.

The unleash secrets were passed, but always overriden by clowder.

CheckFeature is now returning early when unleash is not configured.

Fixes: HMS-4929


lzap (approved):
As long as it does work yeah. This is a tad cleaner.

PR #2738, opened 2024-11-06
build(deps): update stub make target

lzap:
I made an attempt to upgrade `gomega`, it requires a manual step of regenerating stubs. But it cannot be done as it requires Go 1.22. So I am at least fixing the `Makefile` it was not regenerating properly when `go.mod` changed.

The second commit gets rid of `x/exp` dependency, no longer needed with Go 1.21.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,6 @@ require (
 	github.com/spf13/viper v1.19.0
 	github.com/stretchr/testify v1.9.0
 	go.openly.dev/pointy v1.3.0
-	golang.org/x/exp v0.0.0-20240719175910-8a7402abbf56
 	gorm.io/driver/postgres v1.5.9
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/redhatinsights/edge-api
 
-go 1.21
+go 1.22
+
+toolchain go1.23.2
```
lzap, line 5: Damn I accidentally made this change. Do not merge.

lzap:
Okay github.com/onsi/gomega cannot be bumped anymore neither major nor minor version until Go UBI Toolbox is on 1.22. Then we can upgrade it manually (regenerating of stubs is necessary).

In other commit I at least remove `x/exp` dependency.

ezr-ondrej (approved):
Thanks ! :+1:

PR #2763, opened 2024-12-13
add helper function for switching between pulp and aws repos (HMS-4121)

loadtheaccumulator:
# Description
Add a helper method for requesting the correct repo URL when switching between AWS and Pulp as the backing store.
Reduce the feature flag call to one location and call the helper where the correct repo URL is needed.
Enables sending the URL of the Pulp parent repo to Image Builder when an update is requested

FIXES: HMS-4121

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -5,6 +5,7 @@ package models
 import (
 	"errors"
 
+	feature "github.com/redhatinsights/edge-api/unleash/features"
 	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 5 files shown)


pkg/models/commits.go
```diff
@@ -64,6 +65,18 @@ type Repo struct {
...
+//
+//	this avoids the feature flag everywhere repo.URL is used
+//	also using GetURL (not URL) to avoid changing the struct used by Gorm
+func (r Repo) GetURL() string {
+	if feature.PulpIntegration.IsEnabled() && r.PulpStatus == RepoStatusSuccess {
+		return r.PulpURL
+	}
+
+	return r.URL
+}
```
lzap, line 78: This could be just `URL` but is fine getters are all around this repo:

https://go.dev/doc/effective_go#Getters

lzap (approved):

pkg/models/commits.go
loadtheaccumulator: @lzap Yeah. I used that guidance when choosing GetURL, but could have misread. I wanted to avoid changing the field to lowercase since there might be implications elsewhere. 

"If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner. The use of upper-case names for export provides the hook to discriminate the field from the method."

loadtheaccumulator: Ah.  Yeah. Not much point in a getter for an exported. In the next PR, I'll rename that to be less getter for a single field and more specific to what it really is. I have to add some logic to differentiate between hosts for internal and external anyway, so it'll be good opportunity.

PR #2769, opened 2024-12-31
Add single tree repo updates

loadtheaccumulator:
# Description
Adds update of a single tree in an existing repo

FIXES: HMS-4121

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -26,7 +26,7 @@ const (
 	RepoStatusError = "ERROR"
 	// RepoStatusPending is for when the repo process is starting
 	RepoStatusPending = "PENDING"
-	// RepoStatusSkipped is for when a Repo is available to the user
+	// RepoStatusSkipped is for when a Repo is available to the user (post commit build)
 	RepoStatusSkipped = "SKIPPED"
 	// RepoStatusSuccess is for when a Repo is available to the user
 	RepoStatusSuccess = "SUCCESS"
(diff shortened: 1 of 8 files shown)


loadtheaccumulator:
Fixing a bug preventing 2nd+ update from using the single tree. The first update works, but second update on gets lost because of the way AWS S3 repos stored before. Fix on the way post holiday. Setting this to draft until then.

podman/container-compose-dbmigrate.yml
```diff
@@ -8,3 +11,6 @@ services:
         - ~/dev/git/RedHatInsights/edge-api:/opt/app-root/src:Z
       working_dir: /opt/app-root/src
       command:  go run cmd/migrate/main.go
+      secrets:
+        - source: edgemgmt_config
+          target: /tmp/edgemgmt_config.json
```
lzap, line 16: Is this relevant?

lzap:
Looks good.

podman/container-compose-dbmigrate.yml
loadtheaccumulator: Only in that it helps with a local db migration (since one is needed here) if using this same podman-compose setup, but not specific to the code change.

lzap (approved):
One comment regarding the timeout - I suggest to make it even longer.

pkg/models/commits.go
```diff
@@ -26,10 +26,12 @@ const (
 	RepoStatusError = "ERROR"
 	// RepoStatusPending is for when the repo process is starting
 	RepoStatusPending = "PENDING"
-	// RepoStatusSkipped is for when a Repo is available to the user
+	// RepoStatusSkipped is for when a Repo is available to the user (post commit build)
 	RepoStatusSkipped = "SKIPPED"
 	// RepoStatusSuccess is for when a Repo is available to the user
 	RepoStatusSuccess = "SUCCESS"
+	// RepoStatusWaiting is for when the repo process has not started (pre commit build)
+	RepoStatusWaiting = "WAITING"
```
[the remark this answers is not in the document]
loadtheaccumulator: You are not wrong. I removed the bits where it was being used.

PR #2780, opened 2025-01-23
fix type conversion panic on job queue failure

loadtheaccumulator:
# Description
Fix an interface conversion panic in the job queue failure handler for resume create image where it was passing ProcessImageJob instead of ResumeCreateImageJob

* Fixed interface conversion panic
* Changed failure state to Error instead of Interrupted to avoid infinite loop of resumed builds on repeatable errors

FIXES: HMS-5383

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor


lzap (approved):
Yeah this is a flaw of the API, Go generics do not allow generic types only methods. So there is no easy workaround for this I can think of.

PR #2781, opened 2025-01-23
refactor: introduce logging library

lzap:
Swaps logrus with `log/slog` standard library logger. It uses a `logrus` proxy so this project can still use the same API to do logging. I would like to recommend start using `slog` for all the logging and `strc` for possible tracing. See the [README](https://github.com/osbuild/logging) for more information.

Tests will fail until https://github.com/osbuild/logging/pull/13 is merged.

Before merging, the app must be tested for all major workflows since this can easily introduce bugs. It should have no side-effects, logging should work as it used to with small changes. For example, `message` field in JSON output is now `msg`, the output on the stdout is warning level in `key=value` text form. This can be changed to JSON if needed.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/redhatinsights/edge-api
 
-go 1.22
+go 1.22.1
+
+replace "github.com/osbuild/logging" => "../logging"
```
lzap, line 5: This needs to be removed once the PR is merged.

main.go
```diff
@@ -114,7 +71,12 @@ func webRoutes(cfg *config.EdgeConfig) *chi.Mux {
 		middleware.RealIP,
 		middleware.Recoverer,
 		setupDocsMiddleware,
-		logMiddleware,
+		strc.NewMiddlewareWithFilters(
+			slog.Default(),
+			strc.IgnorePathPrefix("/metrics"),
+			strc.IgnorePathPrefix("/status"),
+			strc.IgnorePathPrefix("/ready"),
+		),
```
lzap, line 79: Note for myself: make use of the panic and custom headers middleware introduced by the linked PR.

PR #2795, opened 2025-02-12
pulp integration: fix bearer auth and remaining correlation headers

loadtheaccumulator:
# Description
Fixes an issue where Authorization Bearer is overriden with Authorization Basic and Pulp requests fail as unauthorized because the user cannot be authenticated.

Also fixes some Correlation-ID header syntax.

FIXES: HMS-5490

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/clients/pulp/artifacts.go
+++ b/pkg/clients/pulp/artifacts.go
@@ -117,7 +117,7 @@ func (ps *PulpService) ArtifactsCreatePipe(ctx context.Context, filename string)
 
 	if resp.JSON201 == nil {
 		return nil, fmt.Errorf("unexpected response: %d, correlation id: %s, bytes written/size: %d/%d, body: %s",
-			resp.StatusCode(), resp.HTTPResponse.Header.Get("Correlation-Id"), written, stat.Size(), string(resp.Body))
+			resp.StatusCode(), resp.HTTPResponse.Header.Get("Correlation-ID"), written, stat.Size(), string(resp.Body))
 	}
 
 	return resp.JSON201, nil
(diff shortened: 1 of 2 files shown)


pkg/clients/pulp/artifacts.go
```diff
@@ -117,7 +117,7 @@ func (ps *PulpService) ArtifactsCreatePipe(ctx context.Context, filename string)
 
 	if resp.JSON201 == nil {
 		return nil, fmt.Errorf("unexpected response: %d, correlation id: %s, bytes written/size: %d/%d, body: %s",
-			resp.StatusCode(), resp.HTTPResponse.Header.Get("Correlation-Id"), written, stat.Size(), string(resp.Body))
+			resp.StatusCode(), resp.HTTPResponse.Header.Get("Correlation-ID"), written, stat.Size(), string(resp.Body))
```
lzap, line 120: This hunk has no effect, headers are insensitive and Go standard library will automaticaly compensate for that. Additionally, it will format them for the wire as `Header-Name-Capital-Letter-Each-Word` in case of HTTP/1.1 and lowercase for HTTP/2 which is required.

However, if Pulp requires correlation ID to be written this way, it is a bug in Pulp since HTTP headers MUST be parsed ignoring case up to the HTTP/1.1 and all lower case for HTTP/2 onwards.

pkg/clients/pulp/identity.go
```diff
@@ -21,10 +21,10 @@ func addAuthenticationHeader(ctx context.Context, req *http.Request) error {
...
 		}
+	} else {
+		// add service account mock header if we are
+		clients.AddBasicCredentialsHeader(ctx, req)
 	}
 
-	// add service account mock header if we are
-	clients.AddBasicCredentialsHeader(ctx, req)
-
 	return nil
```
lzap, line 29: Ah good catch, my bad. I thought it would stack but that is indeed not the case.

lzap (approved):
Nice.

pkg/clients/pulp/artifacts.go
loadtheaccumulator: Yeah. I just matched it up with Content Sources and can deep dive if it's still not carrying over.

PR #2796, opened 2025-02-13
handle pulp commit update with aws parent

loadtheaccumulator:
# Description
Handle a commit update stored in Pulp when the parent is stored in AWS

FIXES: HMS-2804

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [ ] Refactor

### DIFF
--- a/pkg/models/commits.go
+++ b/pkg/models/commits.go
@@ -62,11 +62,12 @@ type Commit struct {
 // Repo is the delivery mechanism of a Commit over HTTP
 type Repo struct {
 	Model
-	URL        string `json:"RepoURL"`          // AWS repo URL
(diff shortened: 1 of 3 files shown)


lzap (approved):
As I said, I do not like the idea of creating images with mixed backends, I would rather enforce customers to create a new image (set).

ezr-ondrej (approved):
Not a huge fan either, but if it fixes the issue :shrug:

loadtheaccumulator:
Needs to consider an imageset with multiple commits and possible ref changes (e.g., rhel/8/x886_64/edge & rhel/9/x86_64/edge). Will push that under separate PR.

PR #2812, opened 2025-03-04
remove pulp test vars from deployment

loadtheaccumulator:
# Description
Remove Pulp integration test vars

FIXES: 

## Type of change

What is it?

- [ ] Documentation update
- [ ] Tests update
- [x] Refactor


ezr-ondrej (approved):
This will break pulp in stage. We don't have the secrets in prod, so lets do it. I'd rather see us fixing it, but fixing prod is more important for me.

